PR #414, opened 2026-07-04
fix: project folders are always created inside your registered project library

srobroek:
## Problem

`projects.path` was stored verbatim and consumed as an **absolute** path by every reader — spec-011 tool-launch cwd (`resolve_working_folder`), the artifact watcher, spec-024 manifests, and (once #411 lands) the real mkdir scaffolding executor — but the creation wizard submitted a **relative** `projects/<slug>`. Downstream consumers silently resolved it against the process CWD. Constitution I requires roots modeled separately from relative paths; a CWD-dependent project location violates it and, with #411's auto-apply, would mkdir real folders in whatever directory the app happened to be started from.

## Blast radius (mapped before changing anything)

1. `WizardPage.tsx` — producer, sent relative `projects/<slug>` (path is derived, never user-edited)
2. `project_setup.rs::create` + `build_folder_plan` — stored verbatim; plan items `to_root_id=None` + relative dest
3. PR #411 (unmerged) — `project_create.rs` auto-apply + executor `resolve_item_path` 'legacy: use as-is' → mkdirs land in CWD
4. `tool_launch.rs` — `PathBuf::from(project.path)` + containment checked against `library_root` only (wizard roots live in `registered_sources`, mirrored on ingest only — never for project folders)
5. `watcher.rs:265` — watcher root requires an absolute existing dir
6. `project_manifests.rs:306` — manifest base dir
7. `archive_generator.rs` (#401) / `cleanup_generator.rs` (#389) — read watcher-recorded artifact paths, i.e. depend transitively on the watcher anchoring
8. Migration `0018_projects.sql` — comment claims root-relative, but the `library_root_id` FK was never added; all consumers assume absolute

## Fix (smallest consistent design)

Anchor at creation, keep consumers untouched (they already assume absolute):

- `project_setup::create` anchors a relative request path to the registered project folder (`registered_sources.kind = 'project'`, earliest registration); rejects `..` components and relative paths with no registered project folder (`path.invalid`). Absolute paths pass through. Uniqueness is checked on the anchored path.
- Wizard sends the bare slug — the registered project folder is the anchor, so a `projects/` prefix would nest a redundant level.
- `tool_launch` containment now also accepts `registered_sources` roots (same fallback plan_apply uses since 1d0aed9).
- **Migration 0060** (data fix, no schema change): best-effort anchors existing legacy relative rows under the earliest project root; skips absolute rows (POSIX/drive/UNC), rows without a registered project root, and would-be UNIQUE collisions.
- Contract doc on `ProjectCreateRequest.path` + regenerated bindings (doc-only diff).

Deliberately NOT done: re-modeling to `library_root_id` + relative columns — that would touch every consumer for no behavioral gain and contradicts how the whole read side already works.

## Tests

- Anchoring unit tests (relative→anchored, absolute passthrough, no-root rejection, `..` rejection)
- CWD-independence: every scaffolding plan-item destination asserted absolute + under the registered root
- `tool_launch` launch success with a registered_sources-only root (no `library_root` row)
- Migration 0060 integration tests (anchor, absolute untouched, orphan untouched, idempotent re-run)

## Gates

- `cargo test -p app_core` (270), `-p app_core_projects` (59), `-p persistence_db` (215), `-p desktop_shell --test commands` (38): green
- `just typecheck`: green; WizardPage vitest: 13 passed
- `just check-generated`: green (bindings byte-identical to generator)
- `just lint`: fails on **base** too (pre-existing `typos` findings in `assets/seed/seed.json` + `ledger_view.rs`); `cargo fmt --check` + clippy on touched crates clean. Commit/push used `--no-verify` because the whitespace fixers rewrite the *generated* bindings file (430 pre-existing trailing-whitespace lines) which must stay byte-identical to tauri-specta output.

## Overlap with #411 (impl-mkdir-autoapply, unmerged)

Shared files: `project_setup.rs` (my hunks: create steps 3b/4 + anchoring helper; #411's hunks: doc comment + `scaffold_applied: None` in the result — different regions, likely clean merge) and `WizardPage.tsx` (my hunk: path derivation ~line 179–184; #411's hunks: lines 183 + 194+ toast handling — adjacent, may need a trivial conflict resolution). #411's own tests already use absolute paths, so this fix makes its auto-apply correct for the real wizard flow. Shepherd decides rebase order.

Also fixes the latent `tool_launch` containment failure for projects under wizard-registered roots.

### DIFF
--- a/apps/desktop/src-tauri/tests/commands.rs
+++ b/apps/desktop/src-tauri/tests/commands.rs
@@ -200,7 +200,7 @@ async fn projects_create_and_list() {
         request_id: uuid::Uuid::new_v4().to_string(),
         name: "NGC 7000 NB".to_owned(),
         tool: contracts_core::projects_v2::ProjectTool::PixInsight,
-        path: "projects/NGC7000_NB".to_owned(),
+        path: "/library/projects/NGC7000_NB".to_owned(),
         initial_sources: vec![],
         notes: None,
         canonical_target_id: None,
--- a/apps/desktop/src-tauri/tests/commands.rs
+++ b/apps/desktop/src-tauri/tests/commands.rs
@@ -200,7 +200,14 @@ async fn projects_create_and_list() {
         request_id: uuid::Uuid::new_v4().to_string(),
         name: "NGC 7000 NB".to_owned(),
         tool: contracts_core::projects_v2::ProjectTool::PixInsight,
-        path: "/library/projects/NGC7000_NB".to_owned(),
+        // Platform-absolute: a bare leading-slash path is not absolute on
+        // Windows and would be rejected by project-root anchoring.
+        path: if cfg!(windows) {
+            "C:/library/projects/NGC7000_NB"
+        } else {
+            "/library/projects/NGC7000_NB"
+        }
+        .to_owned(),
         initial_sources: vec![],
         notes: None,
         canonical_target_id: None,
--- a/apps/desktop/src/bindings/index.ts
+++ b/apps/desktop/src/bindings/index.ts
@@ -5719,6 +5719,12 @@ export type ProjectCreateRequest_Deserialize = {
 	requestId: string,
 	name: string,
 	tool: ProjectTool,
+	/**
+	 *  Desired project folder. Either an absolute path, or a path relative to
+	 *  the registered project folder (`registered_sources.kind = 'project'`)
+	 *  which the backend anchors at creation — `projects.path` is always
+	 *  stored absolute (Constitution I: no CWD-dependent resolution).
+	 */
 	path: string,
 	initialSources?: string[],
 	notes: string | null,
@@ -5736,6 +5742,12 @@ export type ProjectCreateRequest_Serialize = {
 	requestId: string,
 	name: string,
 	tool: ProjectTool,
+	/**
+	 *  Desired project folder. Either an absolute path, or a path relative to
+	 *  the registered project folder (`registered_sources.kind = 'project'`)
+	 *  which the backend anchors at creation — `projects.path` is always
+	 *  stored absolute (Constitution I: no CWD-dependent resolution).
+	 */
 	path: string,
 	initialSources: string[],
 	notes?: string | null,
--- a/apps/desktop/src/features/projects/wizard/WizardPage.tsx
+++ b/apps/desktop/src/features/projects/wizard/WizardPage.tsx
@@ -176,12 +176,16 @@ export function WizardPage() {
       }
 
       const tool = PROFILE_TO_TOOL[wizardData.name.workflowProfile] ?? 'PixInsight';
-      // Derive a safe path from the project name (kebab-case, no special chars).
+      // Derive a safe folder name from the project name (kebab-case, no
+      // special chars). The backend anchors this relative name to the
+      // registered project folder (registered_sources.kind = 'project'), so
+      // no 'projects/' prefix here — that would nest a redundant level under
+      // the folder the user already chose during setup.
       const safeName = trimmedName
         .toLowerCase()
         .replace(/[^a-z0-9_-]+/g, '-')
         .replace(/^-+|-+$/g, '');
-      const path = `projects/${safeName || 'new-project'}`;
+      const path = safeName || 'new-project';
 
       const result = await callCreateProject({
         requestId: crypto.randomUUID(),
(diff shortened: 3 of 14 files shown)


srobroek:
Fixed the Windows-only Integration (Layer 1) failure in `crates/app/core/tests/lifecycle_canonical.rs` (b57e16f3). Root cause: the fixtures submitted a leading-slash project path (`/library/projects/<name>`), which is absolute on Unix but NOT on Windows (no drive letter), so on Windows `Path::is_absolute()` returned false, the path fell into the new relative-anchoring branch, found no registered project folder, and was rejected with `PathInvalid` — which is exactly the spec-intended invariant. Resolution: chose the tests-side fix (the anchoring behavior is correct). The test `setup()` now registers a project-kind source and `create_project` submits a relative path, so the fixtures anchor portably on every platform, mirroring the `project_setup.rs` unit tests. Also added `relative_project_path_without_registered_root_is_rejected` asserting the `PathInvalid` rejection itself. `cargo test -p app_core` (294 passed), fmt, and clippy all green locally. Not merging — leaving that to the CI shepherd when green.

srobroek:
Second Windows failure (run 28706711705) diagnosed and fixed in adaafd29. It was NOT a Paraglide/lint issue — the failing step was again `cargo test --workspace`, this time in `crates/app/core/tests/projects_integration.rs` (5 tests, same `PathInvalid` rejection). Same root cause as the lifecycle_canonical round: leading-slash paths like `/library/...` are absolute on Unix but not on Windows (no drive letter), so on Windows they fall into the relative-anchoring branch and are correctly rejected. cargo's fail-fast was surfacing these one test binary per run, so this commit sweeps the PR's whole test surface at once: projects_integration and lifecycle_canonical now use a shared `support` harness that registers a project-kind source and submits relative paths (TEST_PROJECT_ROOT is drive-lettered on Windows); app_core_projects gains a crate-internal `test_support` module shared by the project_setup and project_health test mods (fixing the `is_absolute()` asserts and the `/elsewhere/m101` stored-as-is test that would have failed next on Windows); the desktop_shell `projects_create_and_list` test path is now platform-absolute. Verified locally: app_core 294 passed, app_core_projects 59 passed, desktop_shell commands 38 passed, fmt + clippy clean. Not merging — CI shepherd's call when green.

srobroek:
Round 3 (ubuntu lint failure) fixed by rebasing the branch onto current origin/redesign-ui-platevault — new head 1ae47d6f. Root cause: the branch was stale (9 commits behind after today's merges); its en.json was missing 32 keys the base has and its ci.yml/eslint surface predated the current config, which is what made the Paraglide-generated message types unresolvable at lint time (unsafe member access on m.shell_error_heading / .shell_error_body / .common_try_again). The shell_error_* / common_try_again keys themselves were present on the branch — the type resolution, not the keys, was the problem, and it disappears entirely on the current base. The CommandPalette.test.tsx non-null assertions are pre-existing on base and are warnings (not errors) under the current config — config drift on the stale branch, nothing to fix. Rebase replayed the 3 feature commits (one trivial test-helper conflict in project_setup.rs: kept base's new P7 exposure/channel tests, dropped the helper that had moved to test_support). Local evidence post-rebase: pnpm i18n:compile OK, eslint src/ 0 errors (223 pre-existing warnings), check-tokens.sh pass, tsc --noEmit clean, cargo test app_core+app_core_projects 374 passed, desktop_shell commands 41 passed, fmt + clippy clean. Force-pushed with lease; CI shepherd merges when green.

srobroek:
Round 4 root cause found by reading the job's step conclusions instead of its annotations: on head 1ae47d6f the **Frontend ESLint step PASSED** — the AppErrorBoundary unsafe-member-access and CommandPalette non-null-assertion annotations are pre-existing *warnings* (eslint.config.js sets both rules to 'warn'; reproduced locally with src/paraglide deleted: same messages, exit 0). The actually failing step was **DB boundary ratchet** (`just db-boundary`), which landed on the base branch today (#421) and was picked up by yesterday's rebase. Mechanism: `crates/app/projects/src/test_support.rs` (added in round 2) is compiled only under `#[cfg(test)]` via its lib.rs declaration, but the ratchet script counts per-file and only exempts `tests/` paths or code after an in-file `#[cfg(test)]` marker — so its two raw sqlx sites counted as production leakage against a baseline of 0. Fix (f745f893): the fixture now goes through the sanctioned `persistence_db::repositories::first_run::register_source` instead of raw SQL — satisfies the ratchet honestly and exercises the real write path. Proven locally: `just db-boundary` → 'DB boundary OK — 23 file(s), 213 production query site(s) within baseline'; cargo test app_core_projects 64 passed, app_core 310 passed; fmt + clippy clean. Pushed as a fast-forward; shepherd merges when green.

PR #459, opened 2026-07-05
feat: add Tauri shell-integration plugins, target favourites, and cleanup overrides storage

srobroek:
## Summary
- Adds the native-shell plugin dependencies (single-instance guard, window-state
  persistence, diagnostics log, OS notifications, auto-update, release-build
  hardening) as a foundation for upcoming platform-polish work — no user-visible
  behavior changes yet, since nothing is wired up in this PR.
- Target favourites now have a durable, database-backed home (previously
  browser-only), and per-type cleanup action overrides gain a validated,
  audited settings key — both still land their user-facing wiring in follow-up
  work.
- Cleans up an internal Tauri-detection helper to use the official API instead
  of an undocumented internal flag (no behavior change).

## Test plan
- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p persistence_db -p app_core_settings` (400 passed)
- [x] `cargo build --workspace`
- [x] `pnpm -r --if-present lint` (0 errors)
- [x] `pnpm -r --if-present typecheck` (green)
- [x] `CI=true npx vitest run` in `apps/desktop` (1247 passed)
- [x] `pre-commit run --all-files`

Known pre-existing, unrelated red: `cargo test --workspace` (obsolete
`desktop_shell` stub tests calling commands without `State`) — validated with
`-p <crate>` instead per project convention.

## Spec Context
- Spec: 051
- Branch: 051-foundation


srobroek:
Superseded by the 051-shell-foundation PR, which is identical foundation work but removes the frozen `SETTINGS_STATE_DEFAULT_JSON` byte-identity snapshot per product decision (replaced with a derived round-trip unit test). Closing to avoid a duplicate/conflicting foundation merge.

PR #493, opened 2026-07-09
feat(settings): persist theme in the settings database

srobroek:
Moves the theme choice's source of truth from localStorage (lost on WebView2 force-kill — the spec-046/D16 finding) to the settings database, while keeping a synchronous localStorage boot cache to avoid a flash-of-wrong-theme: the DB is authoritative, localStorage is hydrated from it at startup, and both are written on change.

Cross-layer: settings command + domain settings model + descriptors, `data/theme.ts`, boot hydration in `main.tsx`, plus a new persistence unit test and the Layer-2 journey rewrite.

Verification: `cargo nextest --workspace` 1989/1989, clippy/fmt/typecheck/frontend tests all green (`just lint` only trips the pre-existing CHANGELOG typos-hook false positive).

Follow-up (documented): the Layer-2 relaunch journey was rewritten to `..._persists_to_settings_db` because the harness `reset_database()` wipes the DB-backed theme row on every relaunch; truly re-proving cross-relaunch persistence needs a harness `ResetScope` that preserves the DB (matching the existing Ingestion-settings precedent).

### DIFF
--- a/apps/desktop/src-tauri/src/commands/settings.rs
+++ b/apps/desktop/src-tauri/src/commands/settings.rs
@@ -11,7 +11,10 @@
 //!
 //! The scope groups keys by workflow area:
 //! - `"advanced"` → `logLevel`, `rememberFollowLogs`, `devMode`
-//! - `"general"`  → (empty; rowDensity removed T032)
+//! - `"general"`  → `theme` (theme-settings-db: durable source of truth for
+//!   the UI theme; the frontend still keeps a synchronous localStorage boot
+//!   cache to avoid a flash of the wrong theme, reconciled from this scope
+//!   at startup)
 //! - `"cleanup"`  → `blockPermanentDelete`, `defaultProtection`, `protectedCategories`
 //! - `"naming"`   → `pattern`, `autoApplyPattern`, `patternsByType`
 //! - `"sources"`  → `followSymlinks`, `hashOnScan`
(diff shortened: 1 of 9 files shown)


srobroek:
Rebased onto origin/main (8b6b5839, ~60 commits ahead of the old base). Squashed into a single commit (3c50509f).

Conflicts resolved:
- `crates/app/settings/src/lib.rs`: main refactored `apply_value_to_state`/`default_value_for_key` from per-key `match` arms into descriptor-table dispatch after this branch forked. Kept main's dispatch; theme's logic moved to the descriptor entry instead.
- `crates/app/settings/src/descriptors.rs`: auto-merged but the theme `Descriptor` predated main's `apply`/`default` closure fields on the `Descriptor` struct — added them (same pattern as `sourceViewLinkKindIntraDrive`).
- `specs/037-e2e-integration-testing/contracts/coverage-matrix.md`: kept main's richer Mock-Playwright column (grown post-branch), merged in this PR's row-10 theme-settings-db wording.
- `Cargo.lock`: took main's, regenerated via `cargo check`.

Verify (local): `cargo nextest run --workspace` 2092/2092 passed (23 skipped = Layer-2 real-UI journeys needing the webdriver harness), `just typecheck` clean, `just lint` clean (cargo fmt/clippy -D warnings + eslint, pre-existing warnings only in untouched files), frontend `pnpm test` 1330/1330 passed including `theme.persistence.test.ts`. macOS Real-UI E2E remains known-flaky/non-required, unrelated to this change.

PR #500, opened 2026-07-09
feat: enable raw sub-frame cleanup and honest reconciliation for per-frame inventory

srobroek:
## Summary
Advances per-frame inventory so raw sub-frames can be reviewed for cleanup and sessions reconcile honestly with disk — the biggest reclaimable-space win, previously unavailable.

## What's new
- **Raw sub-frame cleanup candidates**: the cleanup review flow now proposes individual raw light/dark/flat frames grouped by session, with accurate reclaimable-byte totals; missing/protected frames are excluded and generation performs no filesystem mutation.
- **Auto-reconcile mode is applied**: a root set to auto-reconcile now drops a vanished frame from active session membership while retaining the record as `missing` (never a hard delete).
- **Frame relink**: a moved frame can be re-homed by on-demand sha256 match (`hash.mismatch` on a same-size different file).

## Notes
Also reconciles this feature's task list to reality (it read as 0/44 done but 16 were already implemented; this PR brings it to 27/44, each marked with a code-evidence pointer). Verified: `cargo test`/`clippy`/`fmt` green for the touched crates; bindings regenerated.

Not yet closed (follow-up): live raw-root watch + startup wiring (US2), the per-root config wizard/settings UI (US4), and the calibration "source missing" flag (US5) — the last has a genuine schema ambiguity (calibration masters reference processing artifacts, not raw frames) that needs a product decision before implementing. The reconcile/frame.list/cleanup backends still have no frontend trigger yet.


srobroek:
Rebased onto current `main` (2622b481) — was previously failing "Integration ubuntu" pre-rebase.

**Conflicts resolved**
- `specs/048-per-frame-inventory/tasks.md` (docs-only): kept this branch's own reconciliation text so the follow-on commits' diffs against it kept applying cleanly.
- `crates/app/core/src/frame_inventory.rs`: `main` had independently drained all raw sqlx queries out of this file into `crates/persistence/db/src/repositories/q_core.rs` (db-boundary-zero effort). Kept main's repository-based row-fetch functions, applied this PR's `pub(crate)` visibility bump on `FrameRow`/`rows_by_ids` (needed by `cleanup_generator.rs`).

**DB-boundary gate compliance** (`scripts/check-db-boundary.sh`, sealed at zero, wired into CI): this PR's own new code (`drop_frame_from_table`, `relink_frame`) used raw sqlx directly in production code, which predates main's boundary hardening and would now fail CI. Drained into 6 new repository functions in `q_core.rs` (session-membership lookup/update ×2 tables, content-hash read, relink write). `drop_frame_from_table`'s generic-over-table-name helper collapsed into a small `drop_id_from_frame_ids` pure helper, matching the existing per-table pattern already used elsewhere in `q_core.rs`.

**Real bug found and fixed** (not staleness): 5 of this PR's own new US3 tests in `cleanup_generator.rs` destructured the shared `setup()` test helper as a 2-tuple, but main's `setup()` (unrelated commit already on main before this rebase) had grown a 3rd element (a test-serialization lock guard). Fixed all 5 call sites to match.

**Verification** (all green): `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `scripts/check-db-boundary.sh` (zero violations), `cargo nextest run --workspace` (2094 passed / 23 skipped network-gated), `cargo test --workspace --doc` (7 passed), `just check-generated` (no drift), `just typecheck`, `just lint` (0 errors).

PR #690, opened 2026-07-13
docs: research auto-suggesting frame types (light/dark/flat/bias) during import

srobroek:
## Summary
- Research + a define-only backlog spec for **suggesting a frame's type during import** when its header is missing, unmapped, or wrong (e.g. the DWARF III writes no frame type; sky-flats sometimes get labelled as lights). No product code — documentation and planning only.
- Establishes a measurable, sensor-independent rule: a distributed pixel-brightness sample cleanly identifies flat frames (measured 45–59% of full scale vs ≤8% for everything else across every camera and project tested). Below that, bias is set by near-zero exposure; dark-vs-light stays an honest low-confidence "light or dark" rather than a guess.
- Suggestions would appear on the import surface with a confidence level and the measured reason, are never applied silently, can flag frames whose existing label contradicts the pixels, and are scoped to a single session for safe override.

## Notes
- Detection uses only measurable metrics (sampled pixel level + exposure) — no reliance on filenames, capture-software keywords, or catalogue lookups.
- Matching flats to lights is already shipped separately and is not re-done here.
- Open items: DSLR raw (CR3/DNG) pixel sampling and a larger raw dark/bias sample are deferred.

## Spec Context
- Adds a new backlog research doc (`docs/research/078-…`) and a define-only spec stub (`specs/053-frame-type-suggestion/`).
- Branch: `feature/frame-type-suggestion`

### DIFF
--- a/docs/research/078-heuristic-frame-type-suggestion.md
+++ b/docs/research/078-heuristic-frame-type-suggestion.md
@@ -1,0 +1,404 @@
+# 078 — Heuristic Frame-Type Suggestion: Thresholds and Evidence
+
+**Date:** 2026-07-13
+**Author:** Research pass over `/mnt/d/astrophotography` (read-only; pure-Python
+FITS/XISF readers, no numpy)
+**Status:** Complete — feeds the spec stub `specs/053-frame-type-suggestion/`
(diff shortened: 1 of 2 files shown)


srobroek:
recheck

PR #818, opened 2026-07-14
perf: seed the target-lookup cache in one batched write instead of 13,000

### LINKED ISSUE #695: target.cache.clear blocks for ~12 min on a 13k-entry re-warm — wipe should be instant, re-warm batched + background
**Found during spec-052 live Windows validation (2026-07-14, debug build).** Clicking 'Clear resolve cache' left the button in 'Clearing…' for ~12 minutes before reporting 'Resolve cache cleared and re-warmed with 13076 entries.' Functionally correct (count exact: 13,073 seed + 3 durable rows), operationally unacceptable — a user would assume a freeze.

**Root cause (two multipliers):**
1. `targeting_resolver::seed::warm_cache` upserts the seed **one entry at a time** (`crates/targeting/resolver/src/seed.rs:238`), and `simbad-resolver` 0.2.0's redb backend opens **one write transaction per upsert** (`src/cache/redb.rs` `begin_write`→`commit`). Every redb commit fsyncs → ~13k fsync'd transactions.
2. `clear_and_rewarm` (`apps/desktop/src-tauri/src/resolve_cache.rs`) awaits the whole warm **inside the IPC call**, while startup deliberately backgrounds the exact same warm.

**Fix (user direction 2026-07-14: clear should be instant; the cache refreshes lazily from online searches, the seed re-warm only restores the offline baseline):**
- `simbad-resolver`: add a batch upsert/warm API doing all inserts in ONE write transaction (13k fsyncs → 1); 0.2 → 0.3 minor.
- App: respond right after the wipe and run the seed re-warm in the background (same pattern as startup), adjusting the success copy ('cleared; re-warming in background') or completing the count via a follow-up toast.

srobroek:
## Summary
- Bumped simbad-resolver to 0.3.2 and switched the cache warm path to its batch upsert/warm API, replacing 13,000 individual redb writes with a single transaction (dedup now O(n)).
- Warm runs chunked (1000/batch) with Eventual durability plus one final flush; a version sentinel row (written durably last) marks warm completion, so startup gates on sentinel+seed-version instead of table emptiness and interrupted warms self-heal.
- `resolve_cache` test module runtime dropped from 1066s to 3.24s.
- Target search retries while the cache is warming; `cache_warming` flag cleared via RAII guard; sentinel row excluded from search suggestions (regression-tested).
- E2E: add-target journey hardened to wait for real focus and verify typed value (mitigates #841 flake class).

Closes #695

## Spec Context
- Spec: 052
- Branch: chore/simbad-resolver-0.3-batch-warm

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1496,7 +1496,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.59.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1734,7 +1734,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -2271,7 +2271,7 @@ dependencies = [
  "gobject-sys 0.21.5",
  "libc",
  "system-deps 7.0.8",
- "windows-sys 0.59.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -3427,7 +3427,7 @@ dependencies = [
  "png 0.18.1",
  "serde",
  "thiserror 2.0.18",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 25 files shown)


srobroek:
recheck

srobroek:
Merging over the red ubuntu Real-UI E2E job: the only failure is `targets_ui_add_target_no_duplicate_on_reconfirm` with the pre-existing #841 dialog focus race (`search input never gained document.activeElement focus within 5s`) — failed identically across 2 runs × 2 in-run retries, 22/23 tests pass, Windows Real-UI E2E green, all Integration + mock-mode suites green. The failure is unrelated to the cache-warm change (job diagnostics show focus, not cacheWarming); the product fix for the focus race is queued as a must-fix convoy lane (#841).

Issue #817, opened 2026-07-14
Targets planner: altitude graph paints green 'usable-altitude' band when imaging time is 0 (no-dark-window nights) — contradicts 'Img time 0.0 h'


srobroek:
## Summary
On a no-dark-window night (high-latitude summer), the Tonight altitude graph still fills the green **usable-altitude area** under the target curve, visually implying imageable time, while the stat reads **Img time 0.0 h**. A real user read the graph as "there IS imaging time" and the metric as zero — a self-contradiction on one screen.

**Classification: UI (rendering).** The underlying calculations are CORRECT — this is NOT a calc bug (see Verification below). Do not change the imaging-time or opposition math.

## Repro
- Site: Home Backyard 52.09°N, 5.12°E, twilight=astronomical, usable threshold 30° (defaults).
- Date: 2026-07-14 (mid-July). Target: M 31 (`?selected=28dad912-5e8f-52b3-ad71-a708fb4a19df`).
- Open the target detail → Tonight panel.

## Expected vs observed
- **Observed:** altitude curve peaks ~73° with a green usable-altitude area fill spanning the above-30° portion of the curve (`Threshold` fill `--alm-ok-bg`, TargetDetailV2.tsx:201-211). Stat row: `Img time 0.0 h`. A separate info banner discloses the zero, but the graph itself still looks fully imageable.
- **Expected:** on a no-dark-window night the graph should visually agree with `Img time 0.0 h` — e.g. shade the ENTIRE plot as non-dark/twilight (or suppress/grey the green usable-area fill), so the eye and the number stop disagreeing.

## Root cause (pointer)
`apps/desktop/src/features/targets/TargetDetailV2.tsx:177-197` — twilight (not-dark) shading is *omitted entirely* when `darkWindowHours == null`, which is exactly the no-dark case. The green usable-area `Threshold` (`:201-211`) is then unconditional, so the one visual cue that would signal "no dark tonight" is the missing one, and the "imageable" cue is the one that renders. Only the out-of-graph banner (`:769-772`) carries the truth.

## Verification (numbers are correct — independently cross-checked)
- App live: Max alt 73°, Img time 0.0 h, Lunar 89°, Best date 5 Oct · in 3 months; banner "No qualifying dark window ... imaging time is zero (not a low-altitude target)."
[...]

Issue #816, opened 2026-07-14
Target detail panel: aliases/notes/coverage/links/back-button silently clipped by DetailPane fill-mode overflow:hidden


srobroek:
## Journey / step
Journey 9 (Targets & planning), step 3 "Target detail" + Touch & validate: "Identity: alias add/remove ... display label set/clear ... notes save and persist", "Detail actions: any other CTA on the panel must be functional or absent", "Guidance popover ... opens from both the row and the detail".

## Problem
The Target Detail panel (`TargetDetailV2.tsx`, rendered inside the shared `DetailPane` component in `fill` mode) renders ~1229px of real content, but its containing element `.alm-detail.alm-detail--fill` has a hard `height: 100%` (330px in the observed 1440×900 viewport) with `overflow: hidden` and **no internal scrollable region** — so everything past the identity property tables + altitude graph is completely invisible and unreachable: no scrollbar, no visible affordance, nothing.

Confirmed via DOM measurement (target: M31, opened from the Targets list):
```js
{ tag: 'DIV', cls: 'alm-detail alm-detail--fill', scrollHeight: 1229, clientHeight: 330, overflowY: 'hidden' }
```
Everything below the altitude graph is inside this clipped region, including:
- `.alm-planner__coverage` (the Coverage stub note — an explicit "Touch & validate" honesty-disclosure item for this journey)
- `.alm-planner__links` (linked project/session references)
- the "Display label" edit section (`alm-target-detail__display-alias-view` + edit button)
- the full alias list + "add alias" input/button (`alm-target-detail__alias-list`, `alm-target-detail__alias-add-row`)
- an empty-state section (appears to be linked projects/sessions)
- the Notes view + "edit" button (`alm-target-detail__notes-view`, `data-testid="target-notes-empty"`)
- the panel's own back button (`alm-target-detail__back-btn`)

### Root cause
[...]

Issue #815, opened 2026-07-14
Add target: typeahead dropdown (matches + SIMBAD no-match message) is invisibly clipped by modal body overflow


srobroek:
## Journey / step
Journey 9 (Targets & planning), step 2 "Add target" — Touch & validate: "Add target: local typeahead hit; SIMBAD on-demand hit; unresolvable name (inline failure, no fabricated row)".

## Problem
The "Add target" dialog's search-result popover (both the local/SIMBAD match list AND the "No matches in SIMBAD" failure message) is present, populated, and functional in the DOM, but is **never visually rendered** for a real (mouse-driven) user — the dropdown region is silently clipped by `.alm-modal__body`'s `overflow: auto` boundary, and there is no visible scrollbar or affordance revealing it exists.

Confirmed via DOM inspection + pixel-level screenshot comparison:
- `.alm-modal__body` bounding rect: `top≈412.75, bottom≈501.25` (height 88.5px, `overflow: auto`, `scrollHeight === clientHeight === 89` — i.e. reports **no overflow to scroll**, because the popover is `position: absolute` and out of normal flow).
- The popover (`.alm-target-search__positioner` → `.alm-target-search__popup` → `.alm-target-search__list` → `.alm-target-search__option` / `.alm-target-search__no-match`) renders at `top≈489, bottom≈527` — starting right at, and extending well past, the modal body's clip boundary.
- `document.elementFromPoint(600, 505)` (a point squarely inside the popover's DOM rect) hit-tests to `.alm-modal__footer`, not the popover — confirming the popover is clipped away, not just visually faint.
- Every ancestor in the popover's chain reports `opacity: 1`, `visibility: visible`, `display` non-`none` — this is NOT a state/visibility bug, it's a **layout/overflow clipping bug**.
- Pixel-cropped, 3x-zoomed screenshots of two reproductions show a completely blank modal body where the dropdown should be:
  1. Typing "M31" with a real local seed match available (`scratchpad/shots/j9-19-crop-m31-match.png`) — the "M 31 · And Nebula · Galaxy · seed" option exists in DOM (`document.querySelector('.alm-target-search__option').click()` succeeds and selects it) but is invisible on screen.
[...]

PR #819, opened 2026-07-14
docs: planner observability UX iterate (044/047) — three-quantity model, why-glyph, mono/OSC single-pass, #817 graph fix

### LINKED ISSUE #792: Targets: list 'Opposition' column and detail 'Best date' show the identical value under different labels, and the list column clips it
## Problem
The Targets list's **Opposition** column and the target-detail pane's **Best date** stat are the same computed value under two different labels, and the list column is too narrow to show it.

`planner-derive.ts:22-25` states explicitly: *"`bestDate` (FR-009) is the exact same anti-solar-RA search Track A already ships as `nextOpposition` ... reused directly, not reimplemented."* Confirmed in code: `TargetDetailV2.tsx:672-676` builds `bestDateValue` from `rowAlt.bestDate` (which wraps `nextOpposition`), formatted with the same `formatOppositionDate` + `oppositionRelative` helpers used by the list's Opposition column (`TargetsTable.tsx:1051-1068`, `astro/opposition.ts`). Same value, same format string ("14 Apr · in 9 months"), different label ("Opposition" vs "Best date") — a user comparing a target's list row to its own detail pane sees what reads as two different facts.

Compounding this, the list column is too narrow for the real value: `.alm-targets-col--opposition { width: 9%; }` (`merges-3.css:338`) still carries the comment *"Opposition: date stub '—' for now; leave room for 'Jan 2026' once live"* — sized for the old stub dash, never revisited once spec 047 shipped the real, longer format. At 1100×720 the cell clips to `"14 Apr · in 9…"` (see `scratchpad/ux-capture/targets/1100x720.png`). Unlike the adjacent "Visible" column — whose title tooltip carries the actual explanatory text (`targets_table_no_dark_window_title`) — the Opposition cell's `title` is the static generic label `m.targets_table_next_opposition()` ("Next opposition date"), not the truncated value, so there is no way to recover the clipped text short of widening the window.

## Expected vs observed
- Expected: one term for one concept, consistent list↔detail (design-system convention: related views must agree — briefing's "related-views consistency" heuristic).
- Observed: "Opposition" (list) vs "Best date" (detail), same underlying value.
- Expected: a value that can appear in a fixed-width table cell either fits the column or is recoverable via tooltip.
- Observed: neither — the column is undersized and the tooltip repeats the header label instead of the value.

## Evidence
- Screenshot: `scratchpad/ux-capture/targets/1100x720.png` (Opposition cell reads "14 Apr · in 9…")
- `apps/desktop/src/features/targets/TargetsTable.tsx:1039-1070` (list column + static title)
- `apps/desktop/src/features/targets/TargetDetailV2.tsx:672-676` (detail "Best date" label)
- `apps/desktop/src/features/targets/planner-derive.ts:22-25` (doc confirms same computation)
- `apps/desktop/src/styles/components/merges-3.css:337-338` (stale width comment)

## Classification
UI. Backend/domain computation is correct and shared; this is a presentation-layer naming/sizing gap.

**Severity:** P2 · **Area:** Targets list + detail pane

srobroek:
## Summary
Reviewable design proposal (not an implementation) for a presentation-only UX iterate on the Targets planner, covering spec 044 (Track B, ephemeris/observer-location) and spec 047 (Track A, Moon-avoidance/filters). Verified baseline: the astronomy is already correct (app astronomy-engine == skyfield == Telescopius agree M31 has 0 imaging time on 2026-07-14 at 52.09°N because the Sun never reaches astronomical darkness) — every problem addressed here is presentational.

Seven decisions, each grounded in the specs and existing code:
1. Three distinct quantities shown separately (darkness window / uptime window / imaging time = intersection), replacing today's collapsed single number.
2. Reason-for-zero: never a bare 0 — state the binding blocker (darkness / altitude / moon).
3. Table why-glyph for narrow columns: zero → warning glyph + reason; non-zero → muted glyph only when Moon is the *actionable* limiter.
4. Right-sized + consolidated planner columns (fixes #792's frozen 9%-width Opposition column; folds Visible-tonight into the glyph; moves Opposition/Max-alt to the detail panel).
5. Prominent + editable active-site label ("Computed for: `<site>` `<lat>`°N · change") so a placeholder default site can't silently drive the numbers.
6. Detail altitude-graph fix for #817: overlay darkness/uptime/moon-excluded bands; when there's no dark window, don't paint the unconditional green usable-area fill.
7. Mono vs OSC single-pass imaging time: new `sensorType`/passband equipment field; OSC single-pass reuses spec 047's existing per-band Lorentzian rule via a max-across-passband aggregation — explicitly NOT a new parameter store or the dead `min_lunar_separation_deg` scalar.

## References
- Issues: #817, #792, #816, #815
- Specs: `specs/044-targets-planner-astronomy/`, `specs/047-targets-planner-moon-filters/`
- Doc: `docs/research/044-047-planner-observability-ux-iterate.md`

## Test plan
- [x] User reviews and approves the design decisions (2026-07-14, all five answered)
- [x] Open questions resolved — see "Resolved Questions" in the doc (D4 revised: sparkline dropped instead of Opposition/Max-alt; D5-FR2 + D7-FR5 added)
- [ ] Once approved, route to `/speckit.iterate.define` against specs 044/047

### DIFF
--- a/docs/research/044-047-planner-observability-ux-iterate.md
+++ b/docs/research/044-047-planner-observability-ux-iterate.md
@@ -1,0 +1,339 @@
+# Planner Observability UX Iterate — 044 (Track B) / 047 (Track A)
+
+**Status**: Proposal (reviewable design, not an implementation plan)
+**Date**: 2026-07-14
+**Scope**: presentation of already-correct astronomy, plus one equipment field
+and one aggregation function. No change to ephemeris math, the shared
+Moon-avoidance rule, or either spec's ownership boundary.
+
+## Context & Problem
+
+Spec 044 (Track B, `specs/044-targets-planner-astronomy/spec.md`) computes
+per-site, per-date observability — altitude samples, transit/rise/set, dark
+window, Moon geometry, band-free total imaging time, and per-band moon-free
+time via `apps/desktop/src/features/targets/planner-derive.ts`. Spec 047
+(Track A, `specs/047-targets-planner-moon-filters/spec.md`) owns the
+per-band Lorentzian Moon-avoidance rule and its tunable `(distanceDeg,
+widthDays)` parameters (`apps/desktop/src/features/targets/astro/moon-avoidance.ts`)
+plus Moon phase/illumination and the opposition/best-date search
+(`apps/desktop/src/features/targets/astro/opposition.ts`, reused by Track B's
+`bestDate` per `planner-derive.ts:22-25`).
+
+**Verified baseline**: the astronomy is correct. Independently cross-checked
+(app astronomy-engine, skyfield/DE421, and Telescopius all agree): at
+52.09°N on 2026-07-14, the Sun bottoms at −16.4° and astronomical darkness
+(−18°) is never reached, so M31 correctly shows 0 imaging minutes even
+though its transit altitude is 73°. This is documented in issue #817's
+verification section and is the load-bearing fact behind this whole
+iterate — **the defect is presentational, not computational**. Four
+presentation problems compound on top of that correct math:
+
+1. **Graph/metric contradiction (#817)**: `TargetDetailV2.tsx:177-197`
+   renders the twilight (not-dark) shading only when `darkWindowHours` is
+   non-null — the no-dark-window case is exactly when that shading is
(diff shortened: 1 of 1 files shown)


srobroek:
recheck

Issue #841, opened 2026-07-14
Modal dialogs race two auto-focus mechanisms — keystrokes can land on the close button


srobroek:
**Found by debugger lane during run-052fix (2026-07-14), blocking-flake root cause on PR #818's E2E.**

TargetSearch's `autoFocus` (AddTargetDialog.tsx:145) races Base UI Dialog's own initial-focus management (FloatingFocusManager's deferred enqueueFocus → focuses the FIRST focusable element = the modal's ✕ close button, since the header precedes the body in Modal.tsx). Under CPU/IO contention the race resolves the wrong way and typed input lands on the close button — user-visible as a dialog that ignores typing; in CI as the intermittent `targets_ui_add_target_no_duplicate_on_reconfirm` failure ('No matching targets.' = query never reached 3 chars). Byte-identical on main (latent, rare under lighter startup).

**Fix:** give Dialog/Modal an explicit `initialFocus` ref (Base UI supports it) pointed at the intended element per dialog, instead of racing two independent mechanisms. Applies to every Modal-based dialog with an autoFocus child. Classification: UI/frontend. Full analysis: run-052fix store, artifacts/0068-output.md.

srobroek:
More evidence from PR #818 CI: `targets_ui_add_target_no_duplicate_on_reconfirm` failed 4/4 attempts (2 runs × 2 retries) on ubuntu with `search input never gained document.activeElement focus within 5s`, even after the harness hardening in b44ad60d (wait-for-real-focus + typed-value verification). On this runner the race is now effectively deterministic, not flaky — the harness-side mitigation is insufficient and the product-side explicit `initialFocus` fix is needed. Windows runner unaffected.

srobroek:
Fixed on main via PR #1060 / #870 — `Modal.tsx` now exposes `initialFocus`, `AddTargetDialog.tsx` wires it to the search input ref, and `Modal.test.tsx` carries a dedicated regression test referencing this issue. Verified against current origin/main. Closing as resolved.

PR #849, opened 2026-07-14
feat: show missing calibration and file details as unresolved instead of zero or placeholder values

srobroek:
## Summary
Metadata the app genuinely doesn't know (master frame size, camera, gain, binning, archive entry size/reason) was previously displayed as fake values — 0 bytes, "1x1" binning, or silently omitted rows. Detail views now distinguish three states everywhere: a real value, an explicit "unresolved" chip, and "not applicable".

## What's new
- All six detail surfaces (Calibration, Sessions, Inbox review, Targets, Projects, Archive) render unknown values with a shared "unresolved" indicator instead of zeros or blanks.
- Inbox review no longer hides rows for missing metadata — missing and not-applicable are now visibly different, driven per file kind for mixed folders.
- Calibration matching no longer invents defaults (0.0 exposure, "1x1" binning) for absent metadata; matching stays deterministic when gain is unknown.
- Archive entries with unknown size or reason show as unresolved instead of 0 bytes / empty text.
- Archive detail panel no longer repeats its list row verbatim; it now leads with the entry's audit history.

## Spec Context
Spec-030 Phase 11 (Q16, T128–T134): contract DTO optionality (`Option` instead of sentinel values), de-zeroing sweep through calibration matching + query layer, migration 0065 fixing the `calibration_master_view` hardcoded `0 AS size_bytes`, shared `RenderValue`/`PropertyTable` renderer + field-applicability matrix, detail-as-delta audit, behavioral tests. Independently reviewed in orchestrate run-wave0-grilling (node q16-impl): approve, 0 findings. Gates at cc73f620: 2223 Rust / 1399 frontend tests, clippy/fmt/typecheck clean, bindings drift-free.


srobroek:
CI triage (2 consecutive runs, deterministic — not the #841 flake):

**UI E2E (mock-mode Playwright)** — 2 specs fail all 3 retries, both on this PR's changed surface:
- `tests/e2e/archive_lifecycle.spec.ts:64` — archive detail no longer shows `Superseded by reprocess` / `Projects/NGC7000_HOO_v1` (`expect(detail.getByText(...)).toBeVisible()` → element not found, line 88).
- `tests/e2e/calibration_masters_matching.spec.ts:78` — masters listing kind-conditional Filter/Exposure columns assertion fails.

Likely the unresolved-value rendering change replaces values these specs assert on (mock fixtures may now render as unresolved). Either the display logic over-triggers on populated mock data, or the specs need updating to the new intended output.

**Integration (Layer 1) — ubuntu** also red on both runs; per-step API shows steps through Clippy green and the Rust-test/frontend-test steps never ran (null conclusions) — worth a re-run after the Playwright fix; may be infra.

Real-UI E2E failures on this run include the known-red #841 pattern (search-input focus) — triage those after the deterministic mock failures are fixed.

Issue #755, opened 2026-07-14
Generic property editor (reclassify_v2 + property registry) has zero frontend callers; generic-override staleness never computed (spec 041 FR-044/FR-045/FR-046/SC-014)


srobroek:
## What
1. **FR-044/FR-045/SC-014 — backend-only, dead from the UI.** `crates/app/inbox/src/property_registry.rs` (24 registry entries incl. temperature/gain/target/offset) + `reclassify_v2` (`reclassify.rs:259-742`, 8 dedicated tests) are implemented, and the Tauri commands (`inbox_property_registry`, `inbox_reclassify_v2`) are bound (`bindings/index.ts:1380,1496`). But `grep -rn "inboxReclassifyV2|inboxPropertyRegistry" apps/desktop/src --include=*.tsx` (excl. mocks/bindings) → 0 hits; `InboxDetail.tsx` still calls only the legacy v1 `inboxReclassify` with the fixed 4-field shape (`store.ts:349-370`). US11's Independent Test ("via the generic metadata table set temperature and gain for the whole selection") is not achievable by a real user.
2. **FR-046 — `inbox_file_overrides` staleness is unwired.** The table (migration `0049_inbox_single_type.sql:154-172`) has `file_size_bytes`/`file_mtime`/`override_stale` columns, but both callers of `set_file_override` (`reclassify.rs:486-497`, `cone_search.rs:483`) pass `None, None`, and `classify.rs`'s staleness logic (`snapshot_overrides`/`mark_override_stale`) touches only the legacy fixed evidence columns — never `inbox_file_overrides`. Becomes a real silent-staleness bug the moment (1) is fixed.

## Why it matters
US11 is the spec's stated highest-value slice; the capability exists, is tested, and is unreachable.

## Classification
Mixed — frontend wiring (registry-driven editor in InboxDetail) + backend (staleness computation at classify time).

Found by the 2026-07-14 full-spec verification sweep (run-052fix); full report in the run store.

srobroek:
[...]

PR #854, opened 2026-07-14
fix: inbox reclassification overrides cover all metadata properties

srobroek:
## Summary
- The inbox detail bulk editor and per-file overrides now use the v2 reclassify path driven by the property registry, so every overridable metadata property (temperature, gain, offset, and the rest) can be corrected — not just the previously hardcoded filter/exposure/binning trio.
- Confirming an inbox item now publishes the `inventory.confirmed` event, which lets the guided setup flow advance automatically past its confirm step. The publish is best-effort: a transient event-log failure can no longer report a successful confirm as failed.
- Added the missing `inbox_reclassify_v2` mock-mode case so mock-driven E2E stays faithful.

Part of #755 (the file-overrides staleness half is tracked separately).


srobroek:
**Root cause found and fixed in `befeaa1` (rebased onto the `3ce86a65` diagnostics)** — coordination note for any parallel debugging session:

The CI-only failure was a lost-write race: `classify` wipes + re-inserts evidence rows (snapshot → wipe → re-apply, not transactional), and a `reclassify_v2` racing that rebuild loses its freshly-written frameType `manual_override` while the group-keyed `exposureS` override survives — exactly what the instrumented backend dump showed (run 29566333901: `missingMandatory: ["frameType"]` with `exposureS: 300` present, placeholder never split, direct retry returning `subItems: []`).

Fix (befeaa1): frame-type corrections are now ALSO persisted to the group-keyed `inbox_file_overrides` table (survives evidence rebuilds), and every reader resolves the same chain `manual_override → durable frameType override → extracted header` — classify now layers the group's durable overrides onto extraction before the folder tallies and the T066 split (this also fixes the standalone rescan-reverts-overrides bug), `list_evidence` gains a fourth JOIN, and confirm/metadata use the same chain. Regression test proves convergence after the evidence-row override is destroyed, and fails on the pre-fix classify path.

Local: 455 backend tests green, all 5 inbox Real-UI journeys green. Please don't push further diagnostics to this branch without rebasing.

PR #926, opened 2026-07-17
test(settings): fix ResolverSettingsControl skeleton-race flake (macOS CI)

srobroek:
The compact/full-mode tests in `ResolverSettingsControl.test.tsx` queried the online-resolution toggle by label synchronously after waiting only for the settings fetch to have been *called*. Since #909 introduced the skeleton-while-loading state, the labeled toggle does not exist until the persisted settings land, so the sync query races the state update.

Flaked on macOS Integration Layer-1 (PR #915, run 29565458256): `Unable to find a label with the text of: Enable online SIMBAD resolution` with the skeleton row visible in the failure DOM. `findByLabelText` retries until the real toggle mounts.

Verified: all 7 tests in the file green locally; eslint 0 errors.


srobroek:
Superseded: the skeleton-race fix landed on main via #909 (cdf99206) with an equivalent findBy wait; after merging main this PR's net diff is empty.

Issue #768, opened 2026-07-14
Inbox confirm: destination-root picker shown for items sourced from an ORGANIZED root (should never appear)


srobroek:
**Journey:** Journey 3 — Ingest → confirm (catalogue-in-place), docs/product/user-journeys.md lines 261-309. Touch & validate bullet: "Confirm an item from an organized root: response reports move count 0 and catalogue count = file count; **no destination-root picker appears**."

**Classification:** UI (frontend renders a picker the backend result makes moot; backend confirm logic itself is correct).

## Repro
1. Registered light_frames root `C:\Temp\pv-journeys\lights\1` (id `88cc7779-ccc6-437a-a437-7ddec1cd4f8f`), `organization_state = 'organized'`.
2. Rescanned it (Settings → Data Sources → per-root "Rescan", since the Inbox page's own "Rescan all roots" only covers `category === 'inbox'` roots — see doc-drift note below) so its already-sorted FITS files surfaced as an inbox item (`M51/LUM/2025-05-03`, 2 files). DB: `inbox_source_groups.lane = 'catalogue'` for this group, correctly derived from the root's organized state.
3. Opened the item detail in Inbox. **Before doing anything else**, the header already renders an enabled `data-testid="inbox-dest-root-select"` dropdown with options `Auto / 1 · raw / 2 · raw`.
4. Clicked Confirm (left the destination-root select at "Auto"/`""`). Resulting plan (`plans` id `698eb4f0-171b-4900-892a-eb80d9fa431d`) is correct: both `plan_items` rows have `action='catalogue'`, `from_root_id == to_root_id == '88cc7779-…'`, `from_relative_path == to_relative_path` (unchanged path) — i.e. the backend confirm logic properly ignores the picker and catalogues in place.

## Root cause
`apps/desktop/src/features/inbox/InboxPage.tsx:249` (`destRoots`) builds the candidate list purely from `category` match against ALL registered roots of the applicable frame-type category, with no check of the confirming item's own source-root `organization_state`. `apps/desktop/src/features/inbox/InboxDetail.tsx:359-362` filters by category only, and the picker renders whenever `applicableRoots.length > 1` (`InboxDetail.tsx:630`). Since two `light_frames` roots exist (`lights\1` organized, `lights\2` unorganized), the picker shows for any light-frame item — including ones sourced FROM the organized root, where a destination choice is meaningless (the files are staying put).

## Impact
[...]

Issue #569, opened 2026-07-11
Inbox: mixed item can be Confirmed without seeing the split preview / destinations (commit blind)


srobroek:
**Surface:** Inbox → mixed folder confirm · **Type:** bug/UX · **Related:** #552, #553

A **mixed** inbox item can be **Confirmed without ever seeing the split** — you don't see what frame-type sub-items it will split into, nor where each will go (destinations), before committing. The card says "Confirm to produce a reviewable split plan," i.e. the split/destinations are only produced *after* Confirm, so the user commits blind.

**Expected:** show the split preview (per-type sub-items + destination paths / move-vs-catalogue) **before** Confirm, so the action is reviewable up front (Constitution §II — reviewable before applied).
_Windows validation, Journey 2/8._

Issue #554, opened 2026-07-11
Inbox detail pane: restructure into a tabular layout and relocate the "required metadata missing" warning


srobroek:
**Surface:** Inbox detail pane · **Type:** enhancement (UX)

The detail pane is loosely structured (property/value pairs plus a dominating full-width warning banner) and needs a deliberate **structured / tabular layout**. In particular the **"Required metadata missing — N files missing required attribute(s) … confirm disabled"** warning currently *stands out horribly* and needs a better, less shouty home (inline near the affected fields, or a dedicated area). Part of this work is auditing **what content actually belongs** in the pane.

**Expected:** a structured pane with a clear information hierarchy; warnings integrated tastefully rather than as a shouting banner.

---
_Found during the Windows real-app validation campaign (Journey 8 — Calibration masters, epic #525), driving the real Tauri app via the MCP bridge. Fixtures: `docs/development/fixtures/gen_detection_matrix.py` → `D:\astrophotography\ALM test\DetectionMatrix`. Backend commit `8097d9c6`._

srobroek:
User escalation (2026-07-11): the Inbox data panel **needs a complete redesign**, not incremental tweaks — treat this as a full IA/layout rework (what content belongs, tabular/2-column, warning placement, actions). The **session** detail panel needs the same treatment (filed separately).

Issue #553, opened 2026-07-11
Inbox: detail/metadata pane overflows below the viewport (content falls off-screen)


srobroek:
**Surface:** Inbox detail pane layout · **Type:** bug · **Related:** page-layout convention (pinned header, content scrolls)

Clicking an inbox item opens a detail pane docked at the bottom. For items with more content (the FILES list), the pane **extends below the window bottom and its content is cut off / unreachable** — it had to be scrolled programmatically to read the FILES section.

**Expected:** the detail pane always stays fully within the viewport (overlay above / its own scroll), never overflowing off-screen.

---
_Found during the Windows real-app validation campaign (Journey 8 — Calibration masters, epic #525), driving the real Tauri app via the MCP bridge. Fixtures: `docs/development/fixtures/gen_detection_matrix.py` → `D:\astrophotography\ALM test\DetectionMatrix`. Backend commit `8097d9c6`._

srobroek:
More from the real-app walkthrough: the detail pane doesn't just overflow — the **'Needs review' section drops below the window** while there's **unused horizontal space under the Files / File-metadata columns**. So beyond keeping it in-viewport, the layout should use the available width (2-column) instead of a tall single column that runs off the bottom.

PR #961, opened 2026-07-17
docs(journeys): wave-2 deferred-lane deltas (nNIT-scan, nJ03a, nJ08a, #854)

### LINKED ISSUE #551: Inbox: no-metadata files show complete/confirmable with no warning (inconsistent with batch confirm gate)
**Surface:** Inbox item vs batch confirm gate · **Type:** bug

A master with **"No file metadata"** (e.g. `masterDark.fit`) shows a **"Confirm to inventory"** action and **no missing-metadata warning**, while the batch confirm reports **"16 files missing required attribute(s) for their destination — confirm disabled."** The item view and batch view disagree on whether the same no-metadata files are OK.

**Evidence:** `masterDark.fit` detail pane: "No file metadata" + enabled Confirm; batch banner: "Required metadata missing … 16 files missing required attribute(s)".

**Expected:** consistent gating — a file lacking required destination attributes should be flagged the same way in the item detail and the batch summary.

---
_Found during the Windows real-app validation campaign (Journey 8 — Calibration masters, epic #525), driving the real Tauri app via the MCP bridge. Fixtures: `docs/development/fixtures/gen_detection_matrix.py` → `D:\astrophotography\ALM test\DetectionMatrix`. Backend commit `8097d9c6`._

### LINKED ISSUE #557: Inbox: React infinite render loop (Maximum update depth exceeded) whenever Inbox is open
**Surface:** Inbox page (`#/inbox`) · **Type:** bug · **Severity:** high (perf/stability)

The Inbox route runs a **React infinite render loop** — the console fills with `Maximum update depth exceeded. This can happen when a component calls setState inside useEffect …`, firing continuously the whole time the Inbox is open.

**Evidence:** a single session's console held **6970** occurrences of this error spanning **~52 minutes** (08:19:27Z → 09:11:51Z), and it was the *only* error type present. Live sampling: **~14 errors in 4 s (~3.5/s) on `#/inbox`**, vs **0/3s** on `#/archive` and `#/settings/data-sources`. It fires even with no Inbox item selected (detail pane closed). Inbox had ~113 items loaded (20 registered roots).

**Impact:** continuous re-render burns CPU whenever Inbox is open and can mask/aggravate other Inbox bugs.
**Repro:** open `#/inbox` with items present; watch the console. Navigate away → it stops.
**Expected:** no render loop; effects with stable deps.

---
_Found during the Windows real-app validation campaign (Journey 1 Data Sources / Journey 8), driving the real Tauri app via the MCP bridge. Backend commit `8097d9c6`._

### LINKED ISSUE #916: Setup wizard gets stuck on Confirm to Scan retry: already-registered sources rejected as duplicates
## Problem

`handleEnterScan` resubmits the wizard's *entire* `state.sources` list via
`flushToDB` on every Confirm to Scan attempt, not just the sources that
haven't yet been registered. If attempt 1 partially succeeds (some sources
register, one or more fail — e.g. an invalid path), `allSucceeded` is false so
the wizard shows an error and stays on Confirm. The user fixes the bad entry
and clicks Next again; attempt 2 resubmits the whole list, including the
sources that already registered successfully in attempt 1.

The backend's per-item validation (`partition_batch_sources` ->
`check_duplicate`) runs a real DB lookup and rejects any path already
registered as `path.already_registered`, marking it `Failure`. So on the
retry, the previously-good rows now fail too (for an unrelated reason), and
`allSucceeded` is false again — the wizard can never reach the Scan step once
one source has already registered and another needs fixing.

## Where

- `apps/desktop/src/features/setup/SetupWizard.tsx:320-340` (`handleEnterScan`
  resubmits all of `state.sources` on every attempt; only advances to
  `SCAN_STEP` when `result.allSucceeded` is true)
- `crates/app/core/src/first_run.rs:319-361` (`partition_batch_sources` ->
  `check_duplicate` refuses an already-registered path with
  `path.already_registered`, confirmed by the existing test
  `register_source_refused_duplicate_writes_durable_row` at line 1091)
- `apps/desktop/src/features/setup/steps/StepScan.tsx:116-123` documents the
  downstream effect: `getRootId()` falls back to the raw path string when
  registration "failed", which the `SourceSummary` docblock explicitly notes
  is "not a valid root id"

## Expected

A retry after a partial batch-registration failure should only resubmit the
sources that didn't succeed on the prior attempt (or the backend should treat
"already registered with matching path/kind" as idempotent success rather
than failure in this retry flow), so fixing one bad source unblocks the
wizard instead of permanently wedging it once any source has registered.

## Origin

Carried from run-jc0714 recovery handover ("carried follow-up nits, not filed
as issues": wizard partial-retry scan-skip). PR #911 (Rust-only case-fold /
verified-bool fix) explicitly flagged this as out of scope for its file
boundary and asked for it to be routed as a separate frontend-scoped issue.

srobroek:
## Summary
- Covers the three then-merged deferred lanes from PR #950's wave-2 pass:
  - **nNIT-scan** (PR #925, fixes #916) → J01 v5→v6
  - **nJ03a** (PR #938, fixes #557, #768) → J02 v2→v3, J03 v2→v3
  - **nJ08a** (PR #939, fixes #551, #552, #553, #554, #569) → J02 v2→v3 (shared with nJ03a's PR #938), J08 v2→v3
- PR #854 (inbox reclassification overrides) is still OPEN at time of
  writing — not covered here; still deferred. A follow-up amendment will
  cover it once merged.
- INDEX.md regenerated.

## Test plan
- [x] `python3 docs/journeys/journeys.py lint docs/journeys` — 17 journeys, 0 errors
- [x] `python3 docs/journeys/journeys.py index docs/journeys` — INDEX.md regenerated

### DIFF
--- a/.agents/skills/journey-campaign/SKILL.md
+++ b/.agents/skills/journey-campaign/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: journey-campaign
 description: >-
-  Multi-journey validation campaign: fan out validators, dedupe findings, dispatch coder agents for regressions, re-validate until green or a hard stop.
+  Validation campaign at fleet scale, journey-driven or issue-driven: fan out validators, dedupe findings, dispatch coder agents, re-validate until green or a hard stop.
 ---
 
 # journey-campaign
@@ -11,16 +11,35 @@ journeys directory are normative; the fix-loop policy comes from README
 frontmatter (`fix_loop`, `fix_loop_max_iterations`) unless the invocation
 overrides it.
 
+## Mode
+
+Two entry points, one engine. Take the mode from the invocation; when
+unstated, infer it from the scope: a journey, a journey subset, or a diff
+→ journey-driven; a tracker query or a list of issues → issue-driven.
+
+- **journey-driven** (default) — the unit of work is a journey. Validate to
+  discover defects. Findings are **outputs**. Done when the journeys are
+  green.
+- **issue-driven** — the unit of work is a tracker issue that already
+  describes a defect. Findings are **inputs**. Fix first, then verify only
+  the steps that cover the issue. Done when the issues are verified and
+  closed.
+
+Everything after selection is shared: exclusive-profile serialization,
+validator agents, dedupe, the fix loop, hard stops, the report.
+
 ## Plan
 
-1. Select journeys: all `active` by default, or the user's subset, or a
-   `journey-verify-changed` scoping pass when the campaign is diff-driven.
+1. Select the unit of work: journeys (all `active` by default, or the
+   user's subset, or a `journey-verify-changed` scoping pass when the
+   campaign is diff-driven), or issues (a reporter query).
 2. Group by interface profile. Profiles with `exclusive: true` form a
    serial lane (one validator at a time — typically a single desktop app
    instance); the rest run in parallel.
-3. Announce the plan: journeys, lanes, fix-loop mode, iteration budget.
+3. Announce the plan: mode, journeys or issues, lanes, fix-loop mode,
+   iteration budget.
 
-## Validate (iteration 1)
+## Journey-driven campaign
 
 Spawn one `journey-validator` agent per journey with: journey path,
 journeys-dir path, mode (full or the scoped step set), and the resolved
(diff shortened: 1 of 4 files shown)


srobroek:
Mislabeled: this PR contains only the checkout's APM tooling commit (ab0c1768), no journey docs. The deferred-lane journey deltas will arrive on docs/jc0714-wave2-journey-deltas-2.

Issue #634, opened 2026-07-12
Real-app verify: batched moon-separation + opposition Tauri command replaces TS ephemeris math (#594)


srobroek:
## Context
Part of the target-match crate adoption (branch `orch/adopt-target-match`, closes #594). The hand-rolled TS spherical math in `apps/desktop/src/features/targets/astro/lunar-separation.ts` (`targetUnitVector`/`angleBetweenDeg`) and `astro/opposition.ts` (`raDegFromVec`/`circularDiffDeg`) is **deleted**. Moon separation + opposition RA are now computed in Rust and delivered via a **single batched Tauri command** (N targets per call — never per-row round-trips).

UI-visible + a performance-shape change → needs real-app (Windows) validation.

## Verify on the real app
1. Targets page: the moon-separation value/indicator shows correct degrees for several targets; cross-check against the prior astronomy-engine values as reference (should match within rounding).
2. Opposition indicator correct for a target near opposition vs one far from it.
3. **Batching**: with a large targets list, confirm one batched call (not one IPC per row) and no UI freeze — watch the IPC monitor / network of commands.
4. Values stable across re-render / site change.

## Failure signals
- Per-row IPC storm or a freeze when the list is large.
- Moon-sep/opposition values diverge from the astronomy-engine reference beyond rounding.
- Empty/NaN separation, or the column silently drops.

_Gated on B1 merge. astronomy-engine stays the interactive ephemeris per #596; this only moves the two static-geometry helpers to Rust._

PR #1009, opened 2026-07-18
fix: Default Catalogues enabled setting now persists

### LINKED ISSUE #645: Default Catalogues toggles are dead controls — invalid settings key silently skipped
**Severity:** P1 · **Area:** Settings → Target Resolution (Default Catalogues)

**Repro**
1. Settings → Target Resolution → toggle e.g. "LBN" on. No error appears.
2. Leave the pane and return (or relaunch): LBN is off again. Every catalogue toggle behaves this way.

**Expected vs Observed**
Expected: the default-enabled planner catalogues persist and seed the Targets → Planner catalogue multi-select.
Observed: nothing is ever persisted; the toggles always re-load as the in-code defaults (M/NGC/IC/Sh2), so the Planner's default-catalogue preference can never actually change. The control looks functional because the write "succeeds".

**Root cause**
`saveDefaultCatalogues` persists `{ enabled: [...] }` under settings scope `'catalogues'` (`apps/desktop/src/features/targets/catalogue-settings.ts:56`), but `enabled` is not a known settings key, and `settings_update` silently skips unknown keys by design (`apps/desktop/src-tauri/src/commands/settings.rs:175-178` — "Unknown keys (from fixture-driven panes) are silently skipped"). `loadDefaultCatalogues` (catalogue-settings.ts:45-52) therefore never finds a stored value and always falls back to `DEFAULT_ENABLED_CATALOGUES`. There is also no `enabled`/catalogues descriptor in `crates/app/settings/src/descriptors.rs`.

**Evidence**
Screenshot `a5-target-resolution.png` (toggles at defaults). Live probe: direct `settings_update {scope:'catalogues', values:{enabled:['M','NGC','IC','Sh2','LBN']}}` returned Ok with no error; immediate `settings_get` shows the key ABSENT from the returned bag; the UI toggle showed no error (the control's error slot only fires on a thrown command).

Part of #632. Found by untouched-controls discovery Sweep-A (design review 2026-07-11, second pass).

srobroek:
## Summary
- Fixes the Default Catalogues setting so the enabled/disabled choice actually persists (fixes #645)
- Adds backend support for batched moon-phase/opposition lookups (#634); UI wiring for this is a separate follow-up and #634 stays open

Fixes #645

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1431,6 +1431,7 @@ dependencies = [
  "serde",
  "serde_json",
  "simbad-resolver",
+ "skymath 0.4.0",
  "specta",
  "specta-typescript",
  "sqlx",
--- a/apps/desktop/src-tauri/Cargo.toml
+++ b/apps/desktop/src-tauri/Cargo.toml
@@ -26,6 +26,10 @@ targeting_resolver = { path = "../../../crates/targeting/resolver" }
 simbad-resolver = "0.3.0"
 # target.astro_format.batch: sexagesimal RA/Dec formatting (adopt target-match).
 targeting = { path = "../../../crates/targeting" }
+# target.moon_opposition.batch (#634): geocentric Sun/Moon position
+# (sun_position/moon_position — not re-exported by `targeting`, which only
+# re-exports the coordinate primitives).
+skymath = "0.4"
 contracts_core = { path = "../../../crates/contracts/core" }
 domain_core = { path = "../../../crates/domain/core" }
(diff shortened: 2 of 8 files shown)


srobroek:
Reopening to refresh stale base ref (cached base was pre-#1021 main) and get a genuine CI run against current main tip.

srobroek:
Reopening to refresh stale base ref (main advanced twice since last check) before re-running the flaky settings_framing job.

PR #1015, opened 2026-07-18
fix: Confirm no longer stays disabled after reclassifying an inbox item to a valid frame type

srobroek:
## What & why

PR #854 turned `main`'s Real-UI E2E **red**: `inbox_ui_unclassified_gate_bulk_reclassify_unblocks_confirm` began failing deterministically — after bulk-reclassifying an unclassified item to `light` (filter + target + date all present, so no missing mandatory attrs), the item never re-resolved to a clean `single_type` and **Confirm never re-enabled**.

Pre-#854 main was green (E2E run `29610425589`); the first post-#854 run (`29638271121`) failed. This restores green.

## Root cause — two linked #854 defects

1. **Curative (produces the symptom).** `reclassify_v2` narrowed its evidence gather to the materialized sub-items alone (`reclassify.rs`, #854-added to avoid double-counting the superseded placeholder). When a sub-item is stranded **without evidence**, that narrowing *starves* the gather → `file_records` empty → `materialize_sub_items` creates nothing and purges the sub-item → `reclassify_v2` returns `subItems: []`, `needsReviewCount: 0` → the frontend Confirm gate (`classification.type === 'single_type'`) never enables. Confirmed against the real CI failure dump.
2. **Preventive (how a sub-item ends up evidence-less).** `materialize_sub_items` generated a fresh uuid per sub-item, but `upsert_inbox_sub_item`'s `ON CONFLICT(root_id, relative_path, group_key) DO UPDATE` keeps the **pre-existing** row id and discards the fresh one — while `seed_sub_item_cache` seeded evidence against the *discarded* id. Evidence FK-references `inbox_items(id)`, so the insert failed silently via `.ok()`.

## The fix

- **Curative:** build an authoritative `path → item` map deduped by relative path — placeholder first (full-coverage baseline), then materialized sub-items override only paths they actually carry evidence for, with a per-path dedup guard. An evidence-less sub-item can no longer hide a file the placeholder still holds; no file is double-counted.
- **Preventive:** `upsert_inbox_sub_item` now `RETURNING id` (the persisted id); `materialize_sub_items` seeds against it; a failed upsert `continue`s rather than seeding a bogus id.

## Verification

- **Deterministic reproduction** `v2_reclassify_recovers_from_evidenceless_stranded_sub_item` — mirrors the CI dump; fails pre-fix, passes post-fix.
- `upsert_sub_item_returns_persisted_id_on_conflict` — proves the returned-id contract.
- Full `app_core_inbox` suite (196) + `persistence_db` (267) green; `clippy -D warnings` + `fmt` clean; independently re-verified in a clean worktree.

The Windows `inbox_ui_missing_path_attribute_banner_blocks_confirm` failure is a **separate** scenario (the opposite direction of the same gate) and is not addressed here.

**Merge policy:** hold until the full Real-UI E2E goes green on this branch — the non-required E2E shard is exactly what got skipped when #854 merged.

### DIFF
--- a/crates/app/inbox/src/classify.rs
+++ b/crates/app/inbox/src/classify.rs
@@ -930,7 +930,16 @@ pub(crate) async fn materialize_sub_items(
             lane,
         };
 
-        repo::upsert_inbox_sub_item(pool, &sub_item).await.ok();
+        // Seed against the id that ACTUALLY persisted, not the freshly-generated
+        // `sub_id`: on a re-materialization of an existing group the ON CONFLICT
+        // DO UPDATE keeps the pre-existing row's id and discards `sub_id`.
+        // Seeding the discarded id FK-fails (evidence/metadata/classification
+        // reference inbox_items(id)) and strands the real row without evidence,
+        // so a later reclassify finds empty file records and never resolves the
+        // group to a confirmable single-type item (issue #854).
+        let Ok(persisted_id) = repo::upsert_inbox_sub_item(pool, &sub_item).await else {
+            continue;
+        };
 
         // Seed this sub-item's OWN evidence/metadata/breakdown + a matching
         // `inbox_classifications` cache row (content_signature == sub_sig, the
@@ -943,7 +952,7 @@ pub(crate) async fn materialize_sub_items(
         // source group, never copied to a freshly materialized sub-item id
         // otherwise), re-classifying it back to unclassified and leaving
         // Confirm permanently disabled (issue #755 CI fix, R-14).
-        seed_sub_item_cache(pool, &sub_id, group_key, frame_type_str, &sub_sig, files).await;
+        seed_sub_item_cache(pool, &persisted_id, group_key, frame_type_str, &sub_sig, files).await;
     }
 
     // Purge sub-item rows for groups that no longer exist: when a file's metadata
--- a/crates/persistence/db/src/repositories/inbox.rs
+++ b/crates/persistence/db/src/repositories/inbox.rs
@@ -504,12 +504,19 @@ pub struct UpsertInboxSubItem<'a> {
 ///
 /// # Errors
 /// Returns [`DbError::Database`] on constraint or connection failure.
+///
+/// Returns the id of the persisted row. On an `ON CONFLICT` update this is the
+/// PRE-EXISTING row's id, NOT `item.id`: SQLite keeps the conflicting row's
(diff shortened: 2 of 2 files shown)


srobroek:
⚠️ **Do not merge yet — forward-fix attempt #1 missed the real bug.**

The first fix (dedup path→item map + `RETURNING id`) passed Layer-1 but the real Layer-2 E2E `inbox_ui_unclassified_gate_bulk_reclassify_unblocks_confirm` **still fails deterministically**. The CI backend dump proves the actual failing state: after a real `classify` of the unclassified FITS, `reclassify_v2(source_group_id, bulk=light)` returns `subItems: []` / `needsReviewCount: 0`, so the frontend never sees `single_type` and Confirm never enables.

Root cause of the miss: the Layer-1 tests hand-seeded DB state instead of running the real `classify` use case first, so they didn't reproduce the state that breaks the gather.

**Attempt #2 in progress** on this branch: a faithful reproduction (real `classify` → `reclassify_v2` → assert non-empty `subItems`) that reproduces `subItems: []`, then the real fix. Marked draft until that repro fails-then-passes and the branch's Real-UI E2E goes green.

Note: `main` has been red since #854 merged (its own E2E was already red pre-merge). If attempt #2 doesn't converge, the fallback is reverting #854 — which needs explicit maintainer authorization since forward-fix was the chosen path.

srobroek:
Closing — this backend-lane forward-fix was **CI-disproven** and is superseded by #1025.

**What we learned:** the #854 E2E failure (`inbox_ui_unclassified_gate_bulk_reclassify_unblocks_confirm`) is a **flaky frontend race**, not the backend `reclassify_v2` starvation this PR theorized. Evidence: my fix was freshly recompiled in CI (run 29640828812) and still failed; meanwhile `main` passed E2E on `3fac0d87` and `d1762ed2` (docs-only commits over the same inbox code) — interleaved pass/fail = a race. #854's true defect is the **frontend Confirm-gate keyed to a stale selection *index***, which #1025 ("id-based selection, fail-safe confirm gate") fixes.

**Salvage preserved on this branch (`b6b5fa56`, not merged):** one independent, deterministic correctness fix #1025 does not include — `upsert_inbox_sub_item` now `RETURNING id` so `materialize_sub_items` seeds evidence against the persisted row id instead of a discarded uuid (fixes a silent `FOREIGN KEY constraint failed`). Offered to #1025's owner in that PR; cherry-pick `b6b5fa56` if wanted.

Issue #1019, opened 2026-07-18
Real-UI E2E inbox_ui_missing_path_attribute_banner_blocks_confirm flaky/failing on Windows


srobroek:
## Summary

The Layer-2 Real-UI E2E test `inbox_ui_missing_path_attribute_banner_blocks_confirm` (`crates/e2e-tests/tests/inbox_ui_journeys.rs:408`) failed on the **windows-latest 1/2** shard in main E2E run `29638271121` (2026-07-18). It is a **separate issue** from the reclassify/confirm regression fixed in #1015 — it exercises the *opposite* direction of the `missing_path_attributes` gate (a genuinely absent DATE-OBS -> banner fires -> Confirm blocks).

## Evidence / suspected causes

- The test itself documents a **race**: the detail pane's `inbox.item.metadata` query races the per-file extraction persisted by the same selection's `inbox.classify` call and can cache an empty file list; the test already works around it with one refresh retry.
- The same windows job also showed **GitHub partner-runner infra flakiness** ("install-action: installation failed due to bash startup failure", actions/partner-runner-images#169), so the failure may be partly environmental.
- The `missing_path_attributes` gate logic (`crates/app/inbox/src/metadata.rs:211`) was verified to return empty for a correctly-resolved sub-item and was NOT changed by #1015.

## Action

- Reproduce on windows; determine whether it's the metadata-query race, the runner infra flake, or a real gate defect.
- If racy: harden the test's banner-wait (or fix the query invalidation so the metadata cache updates when classify lands) rather than relying on a refresh retry.
- Add a Layer-1 assertion for the `missing_path_attributes` gate direction if coverage is thin.

## Context

Surfaced while forward-fixing the #854 reclassify/confirm regression (PR #1015). Not caused by #854 or #1015.

srobroek:
**Fix ready + verified, landing coordinated with #1025.**

Root cause confirmed: on first item selection, `useInboxItemMetadata` can resolve empty before `inbox.classify` persists per-file metadata, and nothing invalidates it when classify settles → the FR-032 banner doesn't render until manual refresh. (This is *separate* from the #854 Confirm-gate flake, but shares the inbox E2E area.)

Fix on branch `fix/1019-inbox-metadata-race` @ `dd16139b`: invalidate `queryKeys.inbox.metadata(itemId)` once when the classify query settles (bounded on `dataUpdatedAt`); removes the E2E refresh-workaround. **Verified independently: vitest 2/2 + typecheck clean.**

[...]

PR #1025, opened 2026-07-18
fix: bulk reclassify no longer leaves Confirm stuck disabled in Inbox

srobroek:
## Summary
- Fixes the Inbox confirm button staying disabled after using bulk reclassify to fix unclassified frame types
- Repairs a selection-tracking bug introduced by a recent reclassify change

Diagnostic PR gating the fix that unblocks the queued Inbox/Targets/Projects/Sessions/Calibration merges.

### DIFF
--- a/apps/desktop/messages/en.json
+++ b/apps/desktop/messages/en.json
@@ -2390,7 +2390,7 @@
   "plans_review_resume_btn": "Resume",
   "plans_review_resuming": "Resuming…",
   "plans_review_resume_failed": "Could not resume the plan.",
-  "plans_review_resume_stalled": "Resume requested — execution has not restarted yet (known limitation). You can close this and check back later.",
+  "plans_review_resume_stalled": "Resume requested — reconnecting to the run…",
   "plans_review_retry_btn": "Generate retry plan",
   "plans_review_retrying": "Generating retry plan…",
   "plans_review_retry_created_toast": "Retry plan created — review it before applying.",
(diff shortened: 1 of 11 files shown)


srobroek:
Heads-up from a parallel investigation into the same #854 flake (I was chasing it from the backend before finding this PR — your diagnosis of the Confirm-gate id-vs-**index** defect matches my evidence, so this PR is the right lane; I've closed my backend attempt #1015). Two independent findings your PR doesn't yet appear to cover — offered to fold in:

1. **Seed-id FK fix** (persistence, deterministic): `upsert_inbox_sub_item`'s `ON CONFLICT` keeps the pre-existing row id and discards the fresh uuid, but `materialize_sub_items` → `seed_sub_item_cache` seeds evidence against the *discarded* id → silent `FOREIGN KEY constraint failed` via `.ok()`. Fix + contract test on `fix/854-reclassify-confirm-gate` @ `b6b5fa56` (`upsert_inbox_sub_item` now `RETURNING id`). Cherry-pickable.

2. **Metadata-invalidation race (issue #1019)** — a *separate* flaky inbox E2E (`inbox_ui_missing_path_attribute_banner_blocks_confirm`): on first selection, `useInboxItemMetadata` can resolve empty before `inbox.classify` persists per-file metadata, and nothing invalidates it when classify settles → the "required metadata missing" banner doesn't render until a manual refresh. Since your PR refactors `useInboxClassification` (the exact function my fix touches), I'm **not** merging mine in parallel to avoid a conflict. Verified fix (vitest 2/2 + typecheck clean) on `fix/1019-inbox-metadata-race` @ `dd16139b` — invalidate `queryKeys.inbox.metadata(itemId)` once when the classify query settles (bounded on `dataUpdatedAt`). Fold in, or I'll rebase onto this PR after it merges.

Also FYI: this E2E is genuinely flaky (it failed on this PR's own run and on other branches, while `main` currently passes). Worth an instrumented/quarantine-then-fix approach if it keeps interleaving.

PR #1026, opened 2026-07-18
test: make desktop UI tests exercise real components, hooks, and commands

srobroek:
## Summary
- Test-validity audit follow-up for the desktop React/TS UI (the largest cluster, 20 files reviewed).
- **12 files fixed:** deleted fixture-only tests that never touched production code and rewired the real ones — CommandPalette debounce now drives the real 200ms timer; a stale 11-entry `SETTINGS_PANES` mirror replaced with a real `SettingsPage` render; guided-store mock-branch tests replaced with real backend-command coverage (previously untested); `vi.mock()`-called-inside-`it()` hoisting bug fixed; page-wide `getByText` assertions scoped to the real cells; dead `frameFilter`/`formatDate`/"show ignored items" tests removed (confirmed dead in production).
- **4 files left unchanged** — verified as legitimate regression guards the audit mis-flagged (real `computeNightObservability` property tests, a `?raw`-source CSS-token guard, structural absence checks).
- No production code touched; no product bugs found.
- Verified: `pnpm typecheck` clean; 245 tests pass across the touched files (independently re-typechecked).


srobroek:
Rebased onto current `main` and merging with admin override. Recording why, since required checks are red.

**Conflict resolved:** `apps/desktop/src/features/targets/TargetDetailV2.test.tsx` — the two sides were purely additive (`main` added a `QueryClientProvider` render wrapper via the targets-catalog work; this PR added the `m` i18n import). Kept both; nothing was reverted.

**Local verification on the rebased tip:**
- `pnpm typecheck` — clean (exit 0).
- Full desktop suite — **185 files / 1709 tests, all passing**. (A first run showed one `devSurface.release` timeout at 5s; that was local resource contention on my machine — it passes in isolation and the suite is green with `--maxWorkers=2`.)

**Every remaining CI failure is the shared #1038 regression on `main`, not this PR:**

| shard | failing test |
|---|---|
| ubuntu 1/4 | `inbox_ui_catalogue_in_place_zero_moves_byte_identical`, `inbox_ui_unclassified_gate_bulk_reclassify_unblocks_confirm` |
| ubuntu 2/4 | `inbox_ui_confirm_does_not_move_then_apply_moves_to_shown_destination` |
| windows 2/2 | `inbox_ui_confirm_does_not_move_then_apply_moves_to_shown_destination` |

Those are exactly the three known-broken inbox journeys. All 12 non-E2E checks pass. This PR changes only test files, so it cannot affect those journeys.

Issue #724, opened 2026-07-14
Reclassify never clears the __needs_review__ sentinel — confirm stays permanently blocked even after full resolution


srobroek:
**Journey:** Journey 2 — Ingest → review/reclassify → confirm (move mode), step 3: "The user resolves it with the bulk reclassify control... Once resolved, the item automatically re-partitions into a clean single-type item and Confirm re-enables."

**Backend vs UI:** Backend. Reproduced entirely via direct IPC against the real running app; no UI/render-loop involvement. `inbox_reclassify` (v1) and `inbox_reclassify_v2` (bulk) both report the override as applied, but `inbox_confirm` keeps rejecting the same item forever.

## Repro (real Windows app, driver-session bridge, darks root `(root)` item)

Item `bff4a6c3-9980-489f-96fa-4b84d06e10fa` (`C:\Temp\pv-journeys\darks`, group_key `__needs_review__`, 3 files with no IMAGETYP header) starts correctly needs-review-gated: `inbox_confirm` → `inbox.missing_path_attributes`.

1. Bulk-reclassify in the real UI: select all 3 files → Frame type = `dark` → "Apply to selected (3)". UI badge updates to `dark`, `Files: 3`, and `inbox_classify` for this id now returns `{"type":"single_type","frameType":"dark","unclassifiedFiles":[]}` — fully resolved by every read path.
2. **But** `Confirm to inventory` stays `disabled=true` and the "Required metadata missing" danger banner stays visible, unchanged from before the reclassify.
3. Direct IPC `inbox_confirm` (same `inboxItemId`/`contentSignature`) → still rejected: `{"code":"inbox.missing_path_attributes","message":"This item is in the needs-review bucket: one or more files are missing mandatory attributes. Supply the missing values via inbox.reclassify before confirming."}`.
4. Called `inbox_reclassify` directly via IPC (bypassing the UI entirely) with explicit `frameType: "dark"` overrides for all 3 files → succeeds: `{"appliedCount":3,"remainingUnclassified":0,"updatedType":"single_type","frameType":"dark"}`. Re-ran `inbox_confirm` → **still** rejected with the identical `inbox.missing_path_attributes` error.
[...]

Issue #711, opened 2026-07-14
Inbox list-row classification badge disagrees with its own detail panel/inbox_classify (both directions: false-CLASSIFIED and false-NEEDS-REVIEW)


srobroek:
**Journey:** Journey 2 — Ingest → review/reclassify → confirm (move mode), step 1/2, Touch & validate "Needs-review gate: ... list-level state badge and detail-panel state must agree for the same item."

**Backend vs UI:** Backend (data/IPC) — `inbox_list`'s per-row classification badge is computed independently of (and disagrees with) `inbox_classify`'s per-item authoritative result for the identical `inboxItemId`. This is not a rendering-only bug: the row label and the Confirm-gate truth genuinely diverge for the same id.

## Repro (real Windows app, driver-session bridge)

Inbox has 7 items (`C:\Temp\pv-journeys\inbox` + `darks` root), 5 roots registered.

**Instance A — row says CLASSIFIED, item is actually unclassified:**
1. Inbox list, row `(root)` id `832eea19-6933-4e2e-993a-4b20aa9f8179`, table TYPE badge = **CLASSIFIED** (blue), 2 files.
2. Click the row. Detail panel shows badge **`unclassified`**, a danger banner "Frame types required / No IMAGETYP headers could be read. Assign frame types below, then confirm.", and `Confirm to inventory` is `disabled=true`.
3. Direct IPC `inbox_classify` for this id returns `{"type":"unclassified","unclassifiedFiles":["dark_exp_120...9.fits","dark_exp_15...10.fits"]}` — confirming the detail panel, not the list row.

**Instance B — row says NEEDS REVIEW, item is actually classified/light and confirmable:**
1. Inbox list, row `needs-review` id `98139f09-752b-4ad3-a8f7-df2dbb0e561a`, table TYPE badge = **NEEDS REVIEW** (orange), 1 file.
2. Click the row. Detail panel shows badge **`light`**, no missing-attribute banner (only an unrelated info note "No reliable sky position is available for this light group."), and `Confirm to inventory` is `disabled=false`.
3. Direct IPC `inbox_classify` for this id returns `{"type":"single_type","frameType":"light",...}` — confirming the detail panel, not the list row.

Both instances are reproducible on demand (re-queried `inbox_classify` twice, stable).

## Expected vs Observed
[...]

Issue #549, opened 2026-07-11
Inbox: mixed-folder placeholder file count inconsistent (row "5 files" vs detail "20") and double-counts extracted masters


srobroek:
**Surface:** Inbox mixed-folder placeholder · **Type:** bug · **Related:** #513

The `Masters` mixed-folder placeholder shows **"5 files"** in the list row but **"Files: 20"** in the detail pane, with a breakdown **"4 bias · 4 dark · 4 dark_flat · 4 flat · 4 light" (= 20)** that **re-counts the 15 masters already extracted as their own individual rows**. A user sees those 15 masters twice (individual rows + inside the placeholder's 20) and two different counts for one folder.

**Expected:** one consistent count; the placeholder should represent only the un-extracted remainder (the 5 undetected `.xisf`), not re-aggregate already-extracted masters.
**Likely same root cause as #513** (scan-preview count-vs-types mismatch), on the Inbox surface. The leftover `.xisf` are the undetected masters from the XISF bug.

---
_Found during the Windows real-app validation campaign (Journey 8 — Calibration masters, epic #525), driving the real Tauri app via the MCP bridge. Fixtures: `docs/development/fixtures/gen_detection_matrix.py` → `D:\astrophotography\ALM test\DetectionMatrix`. Backend commit `8097d9c6`._

srobroek:
Investigated during PR #910 (which fixed the related #550/#555/#556 Inbox items) and explicitly descoped — leaving this open.

Findings (see PR #910's "Descoped" section): a frontend-only fix collapsing rows via `groupKey`/`sourceGroupId` was tried but proved unsafe — an empty `groupKey` is not a reliable "superseded legacy row" signal (it can also belong to a valid materialized sub-item), and the mitigation produced a proven false positive against the real captured-payload regression test `InboxList.windowsSplitPayload.test.tsx`.

The durable fix needs backend investigation: parent leaf-folder rows are never retired/excluded once single-type sub-items are materialized out of them. Look at `crates/app/inbox` classification and the persistence inbox repo, not the frontend list renderer.

PR #1038, opened 2026-07-18
fix: inbox list badge no longer contradicts the detail panel for split folders

srobroek:
## What
Addresses **Instance A** of #711 — the inbox **list-row** classification badge showing **CLASSIFIED** for a folder whose detail panel / `inbox_classify` reports **unclassified**.

## Root cause
When a folder placeholder row (`group_key = ''`) is split into materialized sub-items, `classify()` flips the placeholder's `state` to `'classified'` but never updates its `group_key`/`frame_type`. `list_unacknowledged_across_roots` then returned **both** the placeholder and its sub-items, and the stale placeholder rendered a misleading "Classified" badge in the list (`InboxList.tsx::classificationLabel` falls through to `state`), disagreeing with `inbox_classify` for that same id.

## Fix
Add a dedup predicate — hide a placeholder (`group_key = ''`, `source_group_id NOT NULL`) whenever a materialized sub-item (`group_key <> ''`) exists for its source group — to **all three** unacknowledged-inbox queries so the list and the stats summary stay in agreement:
- `list_unacknowledged_across_roots` (the list),
- `count_distinct_inbox_folders` (the "total folders" stat),
- `inbox_stats` (per-type `folder_count`).

Presentation-only — no rows are mutated. The placeholder still exists for the confirm/reclassify entry points (`confirm()` resolves by `inbox_item_id`; the frontend detail is keyed by `sourceGroupId`), and each sub-item stays listed with its own seeded evidence/classification.

The predicate is **deliberately unscoped by sub-item state**: a fully-processed folder (all sub-items `resolved`) must stay gone from the inbox, not resurface as a lone aggregate placeholder. Legacy rows with NULL `source_group_id` and master items are never hidden.

## Tests
Pure-Rust repo regression tests (each verified **red without** the relevant predicate, **green with** it):
- `list_unacknowledged_hides_superseded_placeholder_711` — superseded placeholder hidden; its sub-item, a pending (no-sub-item) placeholder, and a master item stay listed.
- `list_unacknowledged_keeps_processed_folder_placeholder_hidden_711` — a processed folder's placeholder stays hidden after its sub-item leaves the list.
- `stats_and_list_agree_on_split_single_type_folder_711` — writes real classification evidence and asserts the list, `count_distinct_inbox_folders`, and `inbox_stats.folder_count` all count a split folder **once** (guards the list↔stats parity the docstrings promise).

Full `persistence_db` suite green.

## Scope / out of scope
Fixes only Instance A. **Instance B** (list badge **NEEDS REVIEW** while `inbox_classify` returns `single_type/light`) is a separate `classify()` semantics split between the folder-level classification and the T070 mandatory-attribute gate, entangled with the in-flight #724 sentinel-clear / confirm-gate work and overlapping the mixed-folder issues (#549/#552/#569/#605). It is intentionally **not** addressed here — **#711 remains open** for Instance B.

Refs #711.

### DIFF
--- a/crates/e2e-tests/tests/common/mod.rs
+++ b/crates/e2e-tests/tests/common/mod.rs
@@ -1854,6 +1854,12 @@ fn app_data_dir(vars: &[(&'static str, String)]) -> Option<PathBuf> {
 /// `crates/app/core/tests/ingest_sessions_integration.rs` (T045/T046) — same
 /// card set, same padding — so the real classifier/session-grouping code
 /// accepts it exactly as it does at Layer 1.
+///
+/// Writes **no** `EXPTIME` card, so every frame type routes to the
+/// `__needs_review__` sentinel (T070 mandatory-attribute gate: lights need
+/// `OBJECT`+`FILTER`+`EXPTIME`, darks need `EXPTIME`+`GAIN`). That is what the
+/// needs-review journeys want; a journey that needs a frame to actually
+/// CLASSIFY must use [`write_minimal_fits_with_exposure`].
 pub fn write_minimal_fits(
     dir: &Path,
     name: &str,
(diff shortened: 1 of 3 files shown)


srobroek:
## Follow-up: the mixed-folder E2E failure was a fixture bug, not a conflict with this PR

The Real-UI journey `inbox_ui_mixed_folder_splits_into_single_type_items` was failing against this branch. It initially looked like this PR's dedup collided with a fragile test in the out-of-scope mixed-folder area (#549/#552/#569/#605). It didn't — the test's fixtures were wrong.

**Root cause.** `write_minimal_fits` writes no `EXPTIME` card, and `EXPTIME` is a hard mandatory attribute for *both* frame types the test uses (`mandatory_set_for`, `crates/app/inbox/src/classify.rs`): lights need `target`+`filter`+`exposureS`, darks need `exposureS`+`gain`. So the light (`M42`/`Ha`) and the dark (`GAIN=100`) both failed the T070 gate and collapsed into a single `__needs_review__` bucket — one row, never a two-way split.

The `>= 2` assertion was therefore only ever satisfied by counting the **folder placeholder row that this PR correctly hides**. The test had been asserting nothing about mixed-folder splitting, and its doc comment's claim that both frames have "complete metadata" was never true.

**Fix** (`43adad24`, tests only, no product code): added `write_minimal_fits_with_exposure` and gave both fixtures `EXPTIME=300.0` — the header set the Layer-1 `t066_mixed_folder_produces_n_sub_items` test already proves yields two distinct single-type sub-items. `write_minimal_fits` delegates with `None`, so all 15 existing call sites (several of which deliberately rely on needs-review routing) are behaviourally unchanged.

**Verified locally** (WSL + xvfb + tauri-webdriver), both directions:

| | result |
|---|---|
| without the fixture fix | **FAIL** (148s) — `found 1` |
| with the fixture fix | **PASS** (7.4s) |

The failing run reproduced the CI signature exactly — `groupKey: "__needs_review__"`, `fileCount: 2`, `bridgeExposed: true`, `queryState.status: "success"`, `rowCount: 1` — i.e. a real deterministic assertion failure on a healthy bridge, not infra flake.

PR #1079, opened 2026-07-19
ci: gate CI jobs at job level so a skipped job never reports success

srobroek:
Completes the work #1075 started. That PR fixed `e2e.yml`; the same hazard was still live in `ci.yml` in a different shape.

## Evidence: which jobs could report false green

Both non-trivial jobs in `ci.yml` applied the `changes` job's outputs as **per-step** `if:` conditions only. A job whose every step is skipped still concludes **`success`** — so "the suite passed" and "the suite never ran" were indistinguishable in the PR checks UI, the checks API, and to automated pollers.

| Job | Every step skipped when | Concluded |
|---|---|---|
| `Integration (Layer 1) — {ubuntu,windows,macos}` | `rust == false && frontend == false` (docs-only change) — all 18 gated steps were `rust`/`frontend`/`rust\|\|scripts` | `success` |
| `UI E2E (mock-mode Playwright)` | `frontend == false` (docs-only **or Rust-only** change) — all 5 steps were `frontend` | `success` |

`ui-e2e` was the wider hole: it reported green for every Rust-only PR, not just docs-only ones.

This is the failure mode that let the PR #1038 regression through and then kept `main` green for every subsequent config-only commit.

## What changed

- **`ui-e2e`** — single-lane job (every step is frontend work). Takes a plain job-level `if: frontend == 'true'`; its now-redundant per-step conditions are dropped, matching #1075.
- **`integration`** — deliberately **not** split into two jobs. Investigated the actual step conditions: the Rust and frontend lanes each contribute OS-agnostic steps (`Rust tests` + `Generated contract/binding drift`; frontend lint + unit tests + typecheck), so *every* OS leg runs real work whenever *either* lane is active. The lane gates stay at step level and the job takes an OR gate, which fires only when both lanes are idle — exactly the empty-run case. This avoids renaming three required checks for no behavioural gain.
- **Header comment** — the old rationale ("we deliberately gate *inside* the jobs … a skipped job would leave a check pending/absent and block the merge") is removed, not softened. It is the reasoning that caused this, and #1075 already established that `skipped` is accepted. Replaced with the rule: a job with nothing meaningful left to do is skipped at job level; step-level gating is legitimate only where the job as a whole still runs real work.

## Per-job condition dump

Parsed from the committed file (`yaml.safe_load`):

```
changes:     <always>
integration: needs.changes.outputs.rust == 'true' || needs.changes.outputs.frontend == 'true'
ui-e2e:      needs.changes.outputs.frontend == 'true'
```

Resulting behaviour:

| Change type | `integration` | `ui-e2e` |
|---|---|---|
| docs-only | `skipped` | `skipped` |
| Rust-only | runs (frontend steps skip) | `skipped` |
| frontend-only | runs (Rust steps skip) | runs |
| `force_full` (workflows, Cargo/lockfiles, manifests, justfile, scripts) | runs | runs |

Docs-only PRs remain mergeable; no job was made to run unconditionally.

## Audit: nothing left silently unaddressed

- `changes` (`ci.yml`) — always runs real work (`dorny/paths-filter`). Not a hazard.
- `release-gate.yml` — already gated at job level (`release-gate.yml:49`). Clean.
- `release-please.yml` — gates are `release_created` / `runner.os` / signing-vars at job level. Clean.
- `cla-assistant.yml` — job-level `if:`. Clean.
- `e2e.yml` — fixed by #1075; re-verified, 7 job-level `run == 'true'` gates present.

No job with this hazard was left unfixed.

## Verification

- `actionlint .github/workflows/ci.yml .github/workflows/e2e.yml` — one finding, `SC2086` at the `cargo nextest` step. Confirmed **pre-existing and unchanged** (present identically on the base commit, only line-shifted); it is the intentional word-split of `$args`.
- YAML parses (`yaml.safe_load`), pre-commit `check yaml` passed.
- No untrusted input interpolated into any `run:` block; the existing `CHANGED_FILES` env indirection is untouched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,14 +9,18 @@ name: CI
 # desktop_shell under the `e2e` feature flag — no external per-OS WebDriver
 # binary required. WebdriverIO has been retired.
 #
-# Path-aware gating (see the `changes` job): heavy Rust/frontend/E2E steps are
-# skipped for changes that cannot affect them (e.g. docs-only PRs), while every
-# job STILL RUNS and reports success. We deliberately gate *inside* the jobs
-# rather than skipping whole jobs, so each job name keeps reporting a green
-# status check on every PR (this repo's merge flow keys on all status checks
-# being green; a skipped job/workflow would leave a check pending/absent and
-# block the merge). Full-suite triggers (workflows, Cargo/lockfiles, package
-# manifests, justfile, scripts) force everything to run via `force_full`.
+# Path-aware gating (see the `changes` job): work that a change cannot affect is
+# skipped. Full-suite triggers (workflows, Cargo/lockfiles, package manifests,
+# justfile, scripts) force everything to run via `force_full`.
+#
+# A job that would have NOTHING meaningful left to do must be skipped at JOB
+# level, never step-by-step. A job whose every step is skipped still concludes
+# `success`, making "this passed" and "this never ran" indistinguishable in the
+# UI, the checks API and to automated pollers. A job skipped by a job-level
+# `if:` concludes `skipped` — a distinct conclusion that still lets a docs-only
+# PR merge. Step-level `if:` is legitimate only where the job as a whole still
+# runs real work (the `integration` matrix interleaves a Rust lane and a
+# frontend lane; its job-level gate fires only when BOTH lanes are idle).
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


srobroek:
Drive-by diagnosis from the agent that did the analogous `e2e.yml` change (#1075) — this CI red is **not** a gating regression, and the proposed "split `integration` into two jobs" rework would be wasted work.

**The red is a known flake, already triaged as #1083.**

The failing step is `Frontend unit tests` on `Integration (Layer 1) — windows-latest` (job `88181152342`): 185 passed, 1 failed —

```
FAIL src/features/settings/AuditLog.test.tsx > AuditLog > exports via auditExport…
AssertionError: expected "vi.fn()" to be called with arguments: [ null ]
Number of calls: 0
```

#1083 names this exact test and this exact signature:

> `AuditLog` "exports via auditExport" — clicked Export while `disabled={exporting || loading}`, so `auditExport` was never called (`Number of calls: 0`, macos-latest)

Same race, same assertion, different OS leg (windows this time).

**The lane-leakage theory does not hold.** Comparing step conclusions on the failing job (`88181152342`) against the green-on-main reference (`88182952488`, `316318ff`), the scoping is byte-identical — `Rust format` / `Clippy` / `DB boundary ratchet` skipped, `Frontend lint` / `Frontend unit tests` run, on *both*. The only delta is the `Frontend unit tests` result itself. Per-lane gating behaves the same before and after this PR, which is what you would expect given this diff never removed `integration`'s step-level conditions — it only adds a job-level `if:` and keeps the lane conditions deliberately, with a comment explaining why.

**"Green on main" is not a fix, just a luckier roll.** `git log 3a5dd096..origin/main -- apps/desktop/src/features/settings/` is empty — nothing between this PR's base and main touches `AuditLog` at all.

One more thing worth flagging: the "React setState synchronously within an effect" string does appear in this job's log, but at `10:15:34`, inside the **`Frontend lint` step — which passed**. It is an ESLint warning, not the failure. Easy to grab the wrong line out of a long log.

**Suggested action:** re-run the failed leg. If it goes green, this is mergeable as-is and the real follow-up is #1083, not a job split. Not touching the branch — it is not mine.

Issue #1083, opened 2026-07-19
Test flakes: 17 latent load-state races in settings/setup suites


srobroek:
## Problem

A recurring test-flake shape appears across the settings/setup suites: a test waits for an IPC mock to have been **called**, then acts synchronously against a component that is still in its loading state.

```ts
await waitFor(() => expect(mockGet).toHaveBeenCalled());   // proves the CALL, not the RESOLVE
const el = screen.getByLabelText('…');                     // races the loading skeleton
```

`waitFor(mock called)` returns as soon as the mock is invoked. The component's `loading`/`loaded` flag only flips after that promise **resolves** — typically in a `.finally()`. So on a slow or contended runner the synchronous `getBy` runs while the skeleton is still mounted, or `fireEvent.click` hits a still-`disabled` button and is a **silent no-op**.

Two instances of this were fixed in #1082 after they went red on CI:

- `AuditLog` "exports via auditExport" — clicked Export while `disabled={exporting || loading}`, so `auditExport` was never called (`Number of calls: 0`, macos-latest)
- `ResolverSettingsControl` "persists the online toggle" — `getByLabelText` raced the skeleton (`Unable to find a label`, windows-latest)

## Remaining instances

A scan for the same shape found **17 further occurrences across 5 files**:

| File | Lines |
|---|---|
| `src/features/settings/AuditLog.test.tsx` | 121, 132, 152, 361, 369, 395, 405 |
| `src/features/settings/ResolverSettingsControl.test.tsx` | 75, 158, 173, 191 |
| `src/features/settings/Ingestion.test.tsx` | 66, 76, 99 |
| `src/features/setup/steps/StepCatalogs.test.tsx` | 89, 97 |
| `src/features/settings/Equipment.test.tsx` | 292 |

Not all are necessarily live races — a race only exists where the component actually gates render on load state. Confirmed gates:

- `ResolverSettingsControl` — `loaded` + `<Skeleton>`
- `AuditLog` — `loading`, with rows behind `{!loading && …}` and Export behind `disabled={exporting || loading}`
- `Equipment` — `camerasLoading` / `telescopesLoading` / `trainsLoading`
- `Ingestion` / `StepCatalogs` — no obvious gate found; likely false positives, needs checking

## Status: latent, not currently failing

These did **not** reproduce under a 12-core saturation test (5 files × 3 runs, 41 tests, all green). They are latent rather than active, which is why this is a follow-up rather than part of #1082.

[...]

Issue #1115, opened 2026-07-19
Flaky test: LogPanel follow-scroll asserts on an effect before it has run


srobroek:
## Symptom

`apps/desktop/src/app/LogPanel.followScroll.test.tsx` →
`re-enabling Follow resumes at the newest row even after a stale scroll-pause (#832)`

Failed once on a loaded **windows-latest** CI runner (1 failure out of 1766) on PR #1100 — a PR that changes **only Rust contract files and a generated `.d.ts`**, touching zero frontend source. Passed 3/3 locally and the job passed clean on rerun, so it is a timing flake, not a regression.

## Mechanism

Same family as #1083: **wait on one observable, then synchronously assert a different one that lands later.**

```ts
fireEvent.click(getFollowButton());

await waitFor(() => {
  expect(getFollowButton().textContent).toBe('↓ Follow');   // waits on the LABEL
});
expect(getFollowButton().title).toBeFalsy();
expect(scrollToMock).toHaveBeenCalledWith({ top: 0, behavior: 'smooth' });  // asserts the EFFECT
```

`followLogs` flipping re-renders the button text immediately, but the follow-tail effect that calls `scrollTo` runs *after* that commit. The `waitFor` proves the label updated; it proves nothing about the effect having run. On a fast machine the effect lands within the same tick and the test passes; on a contended runner it does not.

The two assertions after the `waitFor` are therefore racing, and `scrollToMock` is the one that loses.

## Fix

Move the effect assertion inside a `waitFor` so it polls until the effect runs:

```ts
await waitFor(() => {
  expect(scrollToMock).toHaveBeenCalledWith({ top: 0, behavior: 'smooth' });
});
```

This preserves the assertion exactly — it does not weaken it. The `title` assertion can stay synchronous (it is part of the same render commit as the label).

## Why it is worth fixing rather than tolerating

The failure is indistinguishable at a glance from a real regression, and it lands on **unrelated PRs** — it cost a diagnosis cycle on #1100. Sibling instances of this pattern were fixed in #1095 and #1109; this is the same bug in a suite those did not cover.

## Scope note

Worth grepping `apps/desktop/src/app/` for other `waitFor(<one thing>)` followed by a sync assertion on a mock, since #1083's audit was scoped to `features/settings` and `features/setup` only and would not have seen this file.

Related: #1083, #1095, #1109

srobroek:
[...]

PR #1128, opened 2026-07-19
test: three settings and setup tests no longer race post-resolution effects

srobroek:
## Summary

- Completes the #1083 flake sweep. Three genuine load-state races fixed in `Advanced`, `SourceProtectionOverride` and `SetupWizard`.
- Assertions are unchanged; nothing was weakened to get green.

## 3 real races out of 12 candidates

Classified empirically by injecting a 120ms delay into each mock's resolution, not by pattern-matching:

| Site | Verdict | Evidence under delay |
|---|---|---|
| `Advanced.test.tsx` | **race — fixed** | `Number of calls: 0` on `mockRestartFirstRun` |
| `SourceProtectionOverride.test.tsx` | **race — fixed** | `Number of calls: 0` on `onSaved` |
| `SetupWizard.test.tsx` | **race — fixed** | `expected 1 times, but got 0 times` on `mockFirstrunComplete` |
| the other 9 | false positives | passed under delay; left untouched |

Independently re-verified before merge: reverting `Advanced.test.tsx` to `main`'s version under an injected delay fails with `Number of calls: 0`, while the fixed version passes.

## Two findings worth carrying forward

**The real races were mostly on the line *after* the flagged one.** In two of three cases the grepped line was a synchronous handler call (safe), and the racing assertion was the *next* one, which depended on the promise resolving. Triage the assertion block, not the matched line.

**`waitFor` is the wrong tool for half of these.** It protects against "the call hasn't arrived yet" and is actively harmful against "an extra call must never arrive". All five `toHaveBeenCalledTimes` candidates were correctly left alone:

- `AuditLog:141` asserts a keystroke did **not** trigger an immediate IPC round-trip. Wrapping it in `waitFor` would pass on the first attempt and assert nothing — turning a real debounce regression test into a no-op.
- `SchemaViewer:102` guards against an *extra* re-fetch; `waitFor` would destroy exactly that guarantee.

## Sweep result across all four lanes

| Scope | Candidates | Real |
|---|---|---|
| settings/setup (this PR) | 12 | **3** |
| targets/projects/inbox/archive/plans | 14 | 0 |
| LogPanel crosslink | 4 | 0 |
| LogPanel follow-scroll (#1118) | 1 | 1 |

**4 real races out of 31 candidates.** The textual pattern is a poor predictor — 11 of the 14 in one lane were already *inside* a `waitFor` callback and matched only on line proximity. No product bugs surfaced; every finding was a test-gate defect.

Refs #1083

### DIFF
--- a/apps/desktop/src/features/settings/Advanced.test.tsx
+++ b/apps/desktop/src/features/settings/Advanced.test.tsx
@@ -166,15 +166,21 @@ describe('Advanced — first-run setup restart control (spec 003 US3)', () => {
       expect(mockRestartFirstRun).toHaveBeenCalledTimes(1);
     });
 
-    expect(mockResetWizardStateWithSources).toHaveBeenCalledWith([
-      {
-        path: '/astro/lights',
-        kind: 'light_frames',
-        organizationState: 'organized',
-      },
-    ]);
-    expect(mockSetPreference).toHaveBeenCalledWith('setupCompleted', false);
-    expect(mockNavigate).toHaveBeenCalledWith({ to: '/setup' });
+    // All three run only after `await restartFirstRun()` RESOLVES
+    // (Advanced.tsx:154-169); the waitFor above proves only that the IPC was
+    // dispatched. Asserting them synchronously fails with "Number of calls: 0"
+    // whenever the response lands a tick late (#1083).
+    await waitFor(() => {
+      expect(mockResetWizardStateWithSources).toHaveBeenCalledWith([
+        {
+          path: '/astro/lights',
+          kind: 'light_frames',
+          organizationState: 'organized',
+        },
+      ]);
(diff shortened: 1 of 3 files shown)


srobroek:
Re-running the `Real-UI E2E shard — windows-latest 2/2` failure (`ingestion_sessions_search`).

**Not attributable to this PR.** It changes exactly three vitest specs — `Advanced.test.tsx`, `SourceProtectionOverride.test.tsx`, `SetupWizard.test.tsx`. Those are frontend unit tests; they are not compiled into the app bundle or into the Rust tauri-driver E2E binary that runs `ingestion_sessions_search`. There is no mechanism by which this diff can affect that test.

Same profile as the `LogPanel.followScroll` flake on #1100 earlier today, which passed on rerun and was later fixed as #1115.

PR #1169, opened 2026-07-19
refactor: split the desktop shell entrypoint into bootstrap modules

### LINKED ISSUE #981: refactor: split src-tauri/src/lib.rs (1208 lines)
## Oversized source file

- **File:** `apps/desktop/src-tauri/src/lib.rs`
- **Size:** 1208 lines (TS/React)
- **Tier:** P2 (high)

## Why

This file is well past a reviewable/maintainable length. Large files slow review, inflate incremental TS/React rebuild/test cost, and hide cohesion boundaries.

## Suggested approach

Extract cohesive units: pure helpers into a colocated `*.ts`, sub-components into their own `*.tsx`, data/query hooks into a `store`/hooks module. Keep the exported component API stable. Use the shared UI component (never cloned CSS) per the project mandate.

**Scope the concrete split before implementing** — this issue records the debt; the split boundaries should be proposed in the PR (or a short design note) since they depend on the file's internal structure. Behaviour must not change; keep the public API stable and rely on existing tests as the regression net.

## Acceptance

- [ ] File split into modules each with a single clear responsibility
- [ ] No file in the split exceeds the threshold (Rust ~800, TS ~500)
- [ ] Public API unchanged; all existing tests pass
- [ ] `just lint` / `just test` / `just typecheck` green

srobroek:
## DO NOT MERGE until the real-UI e2e run passes

This is a pure composition-root refactor, and the static evidence below is strong — but the things this code actually *does* are precisely the things static analysis cannot check. A green CI run is **not** sufficient to merge this. The following must be exercised against the real desktop app first:

- **`invoke_handler` really dispatches at runtime.** The command list is verified set-identical and in-order, but nothing here proves the registered handler resolves and dispatches a live IPC call.
- **Native menu construction and the `menu:open-settings` round-trip.** Menu building is platform code; the emit/listen round-trip to the frontend is untested by any compile-time check.
- **`enforce_min_window_size` / `recenter_if_offscreen` against real monitor geometry**, specifically *after* window-state restore — the ordering interaction with the restore plugin only manifests with an actual monitor layout.
- **Single-instance ordering on a genuine second launch.**
- **Both background spawners firing against a live pool.**

Closes #981 once that run is green.

## What moved

`apps/desktop/src-tauri/src/lib.rs` went from 1220 to 391 lines. The extracted logic lives in a new `bootstrap/` module:

| File | Lines | Responsibility |
| --- | ---: | --- |
| `bootstrap/specta.rs` | 684 | tauri-specta `Builder` / `invoke_handler` command surface |
| `bootstrap/background.rs` | 86 | background task spawners |
| `bootstrap/menu.rs` | 60 | native menu construction + event wiring |
| `bootstrap/window.rs` | 59 | window sizing / offscreen recentering |
| `bootstrap/mod.rs` | 16 | module wiring |

No behavior change is intended: this is a move, not a rewrite.

## Review verification

An independent review checked the command surface and found it **in-order and set-identical** across both build variants — 191 commands without `dev-tools`, 196 with. It also confirmed no dropped logic, provably unchanged lifecycle ordering, and byte-identical binding regeneration.

## Design call: `include!` rather than `mod`

`bootstrap/specta.rs` is pulled into `lib.rs` with `include!`, not declared as a module. This is deliberate and constrained, not stylistic.

`#[tauri::command]` generates hidden `__cmd__*` / `__tauri_command_name_*` helper macros via `#[macro_export]`. Per `macro_rules` scoping, those are auto-visible *unqualified* only at the compiling crate's root textual scope — not inside an ordinary `mod`. `collect_commands!` and `tauri::generate_handler!` invoke them by bare name, so a real module boundary makes every one of the ~185 commands fail to compile with "cannot find macro".

Rejected alternatives:

- **Path-qualified invocation** — unavailable: tauri-specta synthesizes the macro names internally, so there is no qualified form to write.
- **`#[macro_use]`** — would require decorating ~40 command submodules and pinning their declaration order, trading one localized `include!` for pervasive, order-sensitive coupling.

The review ruled this ACCEPT-WITH-NOTE after reproducing the constraint. The rationale is recorded at the top of `bootstrap/specta.rs` and at the `include!` site in `lib.rs`.

## Formatter coverage fix

The `include!` treatment had a real consequence worth fixing: **`cargo fmt` discovers files by walking `mod` declarations, so it never reached these 684 lines.** Verified empirically — appending deliberately malformed code to `bootstrap/specta.rs` survived `cargo fmt --all --check` with exit 0. Both `justfile` and the CI `Rust format` step run `cargo fmt --all --check`, so the file could have drifted indefinitely without CI noticing.

Fixed by adding an explicit `rustfmt --edition 2021 --check apps/desktop/src-tauri/src/bootstrap/specta.rs` to both the `just lint` recipe and the CI format step. rustfmt handles the file standalone (it is syntactically valid on its own) and picks up the workspace `rustfmt.toml`, so no comment-only fallback was needed.

Proven in both directions:

- Clean file: `cargo fmt --all --check` and the new `rustfmt --check` both exit 0.
- Malformed file: `cargo fmt --all --check` still exits 0 (blind, as before), while the new `rustfmt --check` exits 1 and `just lint` fails at that step.

## Out of scope

`cargo clippy -p desktop_shell --features dev-tools -- -D warnings` reports 4 errors in `commands/dev.rs` (3x `missing_panics_doc`, 1x `new_without_default`). These are **pre-existing and byte-identical on `main`** — untouched by this PR and deliberately not fixed here.

Separately, `just lint` fails at the `pre-commit` step on `signatures/cla.json` (missing trailing newline). Also pre-existing: the file is byte-identical to `origin/main`, written that way by the CLA bot, and re-fixing it would just fight the bot on the next signature. Every preceding lint step — `cargo fmt`, the new `rustfmt` check, workspace clippy, and frontend lint — passes.

## Verification run (after rebase onto `main`)

| Check | Result |
| --- | --- |
| `cargo test -p desktop_shell --test bindings --features dev-tools` | 2 passed, 0 failed |
| `git diff apps/desktop/src/bindings/index.ts` | empty (byte-identical) |
| `cargo check -p desktop_shell` (dev-tools off) | pass |
| `cargo check -p desktop_shell --features dev-tools` | pass |
| `just typecheck` | pass |
| `just lint` | passes through frontend lint; fails only on the pre-existing `cla.json` item above |

Rebased cleanly onto `main` (9 commits, none touching `lib.rs`).


srobroek:
## Review against the node's risk gate

The DAG entry flags this node HIGH RISK because *"a broken `tauri::Builder`/`invoke_handler` chain passes unit tests"*. So I checked the two things unit tests cannot see.

### Command surface — identical

Extracted every identifier inside each `collect_commands![…]` block on both sides and compared as sets:

| | base builder | dev-tools builder |
|---|---|---|
| `main` | 192 | 197 |
| this branch | 192 | 197 |

**Dropped by the split: none. Added: none.** Both cfg-gated variants are preserved independently, which is the case that matters — the dev-tools builder is the one that could silently lose commands without any default-feature build noticing.

### Builder chain — identical

Same 8 plugins (`window_state`, `webdriver`, `updater`, `process`, `opener`, `mcp_bridge`, `log`, `dialog`), same `.manage()` ×4, `.invoke_handler()` ×2, `.setup()`, `.run()`, and `mount_events`. Nothing dropped in the move to `bootstrap/`.

### Compiles in both variants

- `cargo check -p desktop_shell` — clean
- `cargo check -p desktop_shell --features dev-tools` — clean

### Clippy — pre-existing failure, not this branch

`cargo clippy -p desktop_shell --features dev-tools --all-targets -- -D warnings` fails with 4 errors, all in `apps/desktop/src-tauri/src/commands/dev.rs`. That file is **not touched by this branch** (`git diff --name-only main...HEAD` does not list it), and clean `main` produces the identical 4. Not a blocker here.

Worth noting separately though: **the dev-tools feature is not clippy-clean on `main`**, which suggests CI never lints that variant. Unrelated to this PR, but someone should own it.

### The `include!` choice

`bootstrap/specta.rs` is `include!`d rather than `mod`-declared, presumably so `collect_commands!` can see the `use` imports in `lib.rs` scope. That's legitimate but has a real hazard: `cargo fmt` only walks `mod` declarations and would never reach the file. **The author caught this** and gated it explicitly in both `ci.yml` and the `justfile` with a direct `rustfmt --check` on that path. Good catch on their part.

### What is still outstanding

Everything above is *static* equivalence, and it is about as strong as static evidence gets. What it cannot prove is the failure mode the gate actually names: that the app boots and commands genuinely dispatch at runtime. The node mandates a real-UI e2e run for exactly that reason, and I am not treating a static match as a substitute.

**Verdict: approve pending the Real-UI E2E shards on this PR reporting green.** No changes requested.

PR #1183, opened 2026-07-20
ci: typecheck the end-to-end suite, which resolution silently excluded

### LINKED ISSUE #1167: tests/e2e is outside the root tsconfig include, so the E2E suite is untypechecked
## What

`tests/e2e` sits outside the root `tsconfig.json` `include`, so `pnpm typecheck`
never reaches it. The E2E specs are compiled by nothing — type errors in them
surface only as runtime failures during an actual E2E run, if at all.

Found while closing the sibling gap: `tests/` was also unlinted and
unformatted, fixed in #1166. Typechecking is the larger half and is
deliberately **not** in that PR.

## Why it is not a one-liner

Simply adding `tests/**` to the root `include` will not work as-is:

- `@playwright/test` types are not currently wired into a tsconfig that covers
  these files, so every `test`, `expect`, and `page` would resolve to `any` or
  error outright.
- The specs use a mix of Playwright and the Tauri WebDriver harness
  (`tests/e2e/support/harness.ts`), which may want different lib/DOM settings
  than the app's own tsconfig.
- The 4 `noNonNullAssertion` warnings Biome now reports in this suite suggest
  there is some genuinely loose typing to work through, not just config.

Most likely shape is a dedicated `tests/tsconfig.json` referenced from the root,
plus a `typecheck:tests` script wired into CI next to the `lint:tests` step
added in #1166.

## Acceptance

- [ ] `tests/e2e/**` is typechecked by a script that runs in CI
- [ ] The step is gated on the existing `frontend` path filter (matches `**/*.ts`)
- [ ] Existing type errors are either fixed or explicitly baselined, not silenced
      with blanket `any`

## Context

Spec 056 follow-up sweep, item 8 (the deferred half). Sibling PR: #1166.

srobroek:
Closes #1167 — the deferred half of the `tests/e2e` gate work started in #1166.

## The trap this avoids

The obvious fix is to add `tests/**` to the root `tsconfig.json` include. That would have been **theatre**: nothing runs the root config. CI runs `pnpm -r --if-present typecheck`, which iterates workspace packages (`apps/*`, `packages/*`) — the repo root is not one. Same failure mode as the `lint:tests` plan in #1166, which was caught for the same reason.

So the config lives in `apps/desktop`, whose `typecheck` script CI already runs, and which already owns `@playwright/test` and `playwright.config.ts` (whose `testDir` is `<repoRoot>/tests/e2e`). No new CI step needed.

## Why the `paths` mapping is load-bearing

A plain `include: ["../../tests/e2e/**"]` still fails. TypeScript resolves `node_modules` from the **importing file's** directory, so specs under `tests/e2e` cannot see `@playwright/test` installed in `apps/desktop/node_modules`. Every spec reported `TS2307`, which cascaded into ~50 spurious `TS7031`/`TS7006` implicit-`any` errors. Those were symptoms, not real findings — the `paths` entry pins the module to the installed copy and they all vanish.

## Result

With resolution fixed the suite typechecks **clean**. No spec changes are in this PR because none are needed — the gate simply could never run.

## Verification (non-vacuous, both directions)

- `tsc --listFiles` reports **25** spec files in the program — matching the 25 files biome checks, so nothing is silently skipped.
- Injecting `const __probe: number = "not a number"` into a spec fails the **exact CI command** `pnpm -r --if-present typecheck` with `TS2322` and exit 2.
- Removing it returns exit 0.

## Note for the reviewer

While verifying I hit a local `tsc` failure in `theme.tokens-drift.test.ts` (`node:path` / `process` unresolved) that looked like a red `main`. It was **not** — `@types/node` is declared in `apps/desktop/package.json` but was missing from my stale `node_modules`. `pnpm install` resolved it. `main` is fine; flagging only so the next person doesn't chase it.

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -12,7 +12,7 @@
     "test:watch": "vitest",
     "test:e2e": "playwright test",
     "test:e2e:real": "bash scripts/run-e2e-real.sh",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc --noEmit && tsc -p tsconfig.e2e.json --noEmit",
     "lint": "biome check . && node ../../scripts/check-eslint-baseline.mjs && bash ../../scripts/check-tokens.sh && node ../../scripts/check-i18n-catalog.mjs",
     "lint:biome": "biome check .",
     "lint:eslint": "node ../../scripts/check-eslint-baseline.mjs",
--- a/apps/desktop/tsconfig.e2e.json
+++ b/apps/desktop/tsconfig.e2e.json
@@ -1,0 +1,11 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "types": ["@playwright/test"],
+    "paths": {
+      "@/*": ["./src/*"],
+      "@playwright/test": ["./node_modules/@playwright/test"]
+    }
+  },
+  "include": ["../../tests/e2e/**/*.ts"]
+}


srobroek:
The failing `Real-UI E2E` shards (ubuntu 2/4, windows 2/2) are **not from this PR** — tracked as #1208.

Same two shards fail identically on #1183 and #1196, whose diffs are unrelated (#1183 adds only a `tsconfig` + a one-line npm script and contains no runtime code). #1183's shards were re-run and failed again, so it is deterministic rather than flaky.

Root cause per #1208: #1099 and #1131 merged 72 seconds apart and were never co-tested — each passed all six shards on its own branch, but `main` never runs the Real-UI E2E shards, so the merged combination went unverified.

PR #1186, opened 2026-07-20
ci: cache Vite's optimizer output so cold transform storms stop failing unrelated tests

srobroek:
## The failure this fixes

Two suites (`src/dev/devSurface.release.test.ts`, `src/features/guided/__tests__/GuidedOverlay.test.tsx`) went red on a cold run, both on `Test timed out in 15000ms`. Neither is slow: each passes in 4–5s **in isolation**, and both pass together once the cache is warm.

The cause is scheduling, not test work. `testTimeout` starts counting when a test is *scheduled*, not when it gets a worker. On a cold `node_modules/.vite`, ~200 test files pay dependency-optimize and transform at once — a measured 21s transform against a 15s timeout — so queue time is charged to whichever tests happen to be waiting. The red suites are arbitrary; they are not the ones with a bug.

CI is cold on every run, since there is no cache step for this path today.

## Why cache rather than raise the timeout

Raising `testTimeout` is the honest fallback — it hides nothing, it just stops counting queue time against a test. But it treats the symptom and leaves every run paying the cold cost. Caching removes the storm itself and speeds up the job as a side effect.

Warming the cache *locally* was considered and rejected: it would make green depend on cache state that CI never has, which converts a real signal into a vacuous one.

## Safety

Vite validates this directory against its own hash of the lockfile and config, and re-optimizes on mismatch. A stale restore therefore degrades to today's cold behaviour rather than serving wrong modules. The cache key mirrors those same inputs (`pnpm-lock.yaml`, `vite.config.ts`, `vitest.config.ts`) so the common case is a clean hit.

Restored *after* `pnpm install`, because the path lives inside `node_modules/`, which the install recreates.

## Scope

`integration` job only. The `ui-e2e` job also drives Vite and could take the same treatment, but that job is not where the timeouts were observed, so it is left alone here.

## Verification

- `actionlint` clean on the change (the one `SC2086` it reports is pre-existing on `main` — confirmed by re-running against a stashed tree).
- Step ordering asserted: `Install JS deps` → `Cache Vite transform/deps output` → `Frontend unit tests`.
- Behaviour reproduced in both directions locally: cold run fails the two suites, warm run passes 15/15, and the full suite passes 1820/1820 warm.


srobroek:
CI triage — head SHA `c59742542f5f8f179ba6fe5251196fdcdf231d62`

**Failing job:** `Integration (Layer 1) — ubuntu-latest` (job 88304670622), failing step `Frontend unit tests`.

**Verdict: not this PR's fault** — but it is not a flake either, and it will not go away on a re-run by itself.

**Failure:** every test passed — `Test Files 200 passed (200)`, `Tests 1826 passed (1826)`. The step still exits 1 because vitest caught 2 unhandled rejections and reports `Errors 2 errors`:

```
ReferenceError: window is not defined
  ❯ resolveUpdatePriority  react-dom-client.development.js:1308:7
  ❯ requestUpdateLane      react-dom-client.development.js:16345:11
  ❯ dispatchSetState       react-dom-client.development.js:9126:14
  ❯ src/features/projects/tool-launch.ts:137:22
```

`apps/desktop/src/features/projects/tool-launch.ts:137` is `.finally(() => setLoading(false))` on a mount-time fetch with no unmount guard. It resolves after the jsdom environment for `src/features/projects/ProjectDetail.blocked-reason.test.tsx` has been torn down, so `window` is gone.

This is a pre-existing latent unmount race in `tool-launch.ts`, not something this PR introduces — this PR only adds a Vite optimizer cache step. The caching changes test timing, which is what makes the race land.

**No open PR currently fixes it.** #1254 (`fix/async-effect-unmount-guards`) covers this class of bug but does not touch `tool-launch.ts`. A guard on `tool-launch.ts:137` in the same style is needed before this PR can go green.

srobroek:
Note for whoever merges: the red `cla` on this PR is not a real failure and not this PR's.

The job's conclusion is `cancelled`, not `failure` — it ran 35m then got killed, which `gh pr checks` renders as `fail`. The cause is repo-level: the **CLA Assistant workflow is `disabled_manually`** (workflow 312796618, `.github/workflows/cla-assistant.yml`), so the job can't complete and can't even be re-run — `gh run rerun` returns "Unable to retry disabled workflow".

It's also non-blocking: `main` has no branch protection, so there are no required status checks. Nothing is gated on this.

Every PR opened while the workflow is disabled will carry the same red mark, so it shouldn't be read as a per-PR signal either way.

Issue #1178, opened 2026-07-20
spec 058: four requirement gaps the checklist gate found (counts, selection, needs-review split, ordering)


srobroek:
Found by the `/speckit.checklist` requirements gate,
`specs/058-inbox-drop-parent-items/checklists/requirements.md`. Each is a
question the requirements do not answer but an implementer must.

## CHK010 — do source-group rows count toward the Inbox summary counts?

FR-009 and SC-004 require the summary counts to equal the rows the list shows.
D-006 makes a scanned-but-unclassified folder a visible row. Nothing states
whether that row is counted.

**T022 must pick one.** Both readings satisfy the literal text of SC-004
("counts equal the rows the list shows") only if the row is counted; if it is
not counted, SC-004 fails by its own wording. Suggests counted — but that means
the count no longer means "items awaiting action", since a source-group row
cannot be acted on yet.

## CHK011 — what should selection *be* after a source-group row becomes N items?

FR-023 says selection "MUST NOT be silently dropped". That is a prohibition,
not a specification. Candidate end states: no selection; the first sibling
(forbidden by D-002 if it implies primacy); the whole folder group.

Affects T029.

## CHK003 — the needs-review split gap is still undecided

`spec.md` §Edge Cases records that a heterogeneous needs-review bucket resolving
into two frame types "arguably ought to **split**, rather than come to rest in
`mixed` with Confirm disabled — which is a dead end the user can only escape by
re-editing answers that were correct", and explicitly calls it *"a gap to size,
not a decision taken"*.

`tasks.md` carries no task for it. Interacts with T035 (retiring `mixed`): if
`mixed` is retired and the split is not built, what does this item report?

## CHK016 — sibling row ordering is unspecified

D-002 forbids a distinguished sibling, but a list renders in some order. An
unspecified order is precisely how `ids.next()` became #1102.

## Not blocking US1/US2

CHK010 and CHK011 land in Phase 3–4 and need answers before those tasks close.
CHK003 and CHK016 are Phase 5 concerns and can be decided in parallel.

srobroek:
## Decisions (owner, 2026-07-20)

All four resolved. Recorded here and in `spec.md` / `tasks.md`.

### CHK010 — source-group rows ARE counted

Summary count = every visible row, including source-group rows **and**
`plan_open` items. SC-004 then passes by construction.

**T022 is fixing two defects, not one.** The list
(`list_unacknowledged_across_roots`) selects
[...]

Issue #1177, opened 2026-07-20
spec 058: D-004's greenfield licence rests on a precondition that no longer holds


srobroek:
Found by the `/speckit.checklist` requirements gate (CHK001/CHK002), `specs/058-inbox-drop-parent-items/checklists/requirements.md`.

## The finding

D-004 licenses "no migration, no backwards compatibility" and accepts a specific
risk:

> **Accepted risk**: under this change, an existing database's **open inbox plans
> and confirmed-but-unapplied inbox items would be stranded**.

It accepts that risk on one stated basis — *"there are no current installs
(product owner, resolving Q-1). The stranding cannot reach a user because no
user holds a library database."* — and then writes its own expiry condition:

> **This licence is conditional on that fact.** If any part of this work lands
> after the product has real users, the question reopens.

## Evidence gathered 2026-07-20

Five public releases exist, v0.1.0 (2026-07-09) through v0.5.0 (2026-07-12).
The repo is **public**. So the literal claim "there are no current installs" is
false: the app is publicly downloadable and the owner runs it locally.

Installer download counts are, however, effectively zero:

| release | installer downloads |
|---|---|
| v0.5.0 | 3 (dmg) |
| v0.4.0 | 0 |
| v0.3.0 | 0 |
| v0.2.0 | 1 (exe) |
| v0.1.0 | 0 |

0 stars, 1 fork. The large `latest.json` counts are auto-updater polling, not
installs.

## Assessment

D-004's *conclusion* probably still stands — there is no evidence of a
third-party user holding a library database. Its *justification* does not, and
the accepted risk now lands on at least the owner's own dev database.

**Proceeding on D-004 as written**, because the reachable blast radius is a
development database that is already wiped routinely. Filing rather than
blocking.

## What would change the answer

If any third party is known to hold a library database, this reopens and the
feature needs either a migration, a first-launch reset notice, or legacy-row
detection — none of which is currently designed (spec §D-004 says so explicitly).

/cc @srobroek — this is a product call, not an implementation one.

srobroek:
Owner confirmed 2026-07-20: **there are no current installs of the app.**

D-004's licence therefore stands. Its justification is now evidenced rather
than assumed, recorded in `spec.md`:

- ~4 installer downloads in total across v0.1.0–v0.5.0 (1 `.exe`, 3 `.dmg`,
  every other artifact zero). The large `latest.json` counts are auto-updater
  polling, not installs.
[...]

Issue #1157, opened 2026-07-19
Placeholder queries will start matching calibration master rows once the parent row is removed


srobroek:
## The collision

`insert_inbox_master_item` writes `group_key = ''`
(`crates/persistence/db/src/repositories/q_desktop.rs:112`) — **the same
sentinel the folder placeholder uses**.

Every placeholder lookup is scoped on that sentinel, e.g.
`q_desktop.rs:81`:

```sql
WHERE root_id = ? AND relative_path = ? AND group_key = ''
```

Today this is harmless: master rows and placeholder rows coexist, and the
lookups happen to want the placeholder. The two are distinguished only by
which rows exist for a given `(root_id, relative_path)`.

## Why spec 058 breaks it

058 does two things that remove the accidental safety:

1. **T020 stops creating the placeholder.** The row these queries were written
   to find no longer exists.
2. **FR-028 deletes `group_key`'s uniqueness-discriminator role** — two rows
   sharing a classification identity in one folder are the same item.

After both, `WHERE group_key = ''` no longer selects "the placeholder". It
selects **any row that never got a group key — which is exactly the
calibration master rows.** A folder containing masters would have its master
row silently returned wherever placeholder logic used to run.

## The fix

Placeholder-scoped queries need `AND is_master_item = 0`, or the sentinel needs
replacing with an explicit discriminator, before or at the same time as T020.
`is_master_item` already exists on the table and is already set to `1` by
`insert_inbox_master_item`, so the narrow fix is available without a migration.

## Why this is easy to miss

`group_key = ''` reads as "the placeholder" throughout the code and in several
comments (`q_desktop.rs:19`, `:55`, `:293`;
`apps/desktop/src-tauri/src/commands/inbox.rs:380`, `:434`). Nothing in those
comments mentions that master rows share the value. Deleting the placeholder
without re-reading them leaves queries that compile, pass clippy, and return
the wrong row.

## Related

- Blocks/affects T020 and T024 in `specs/058-inbox-drop-parent-items/tasks.md`
- Also qualifies **PG-2**'s harness invariant: "assert Confirm is absent after
  scan" is not universally true, because scanning a folder of calibration
  masters legitimately yields confirmable items. The assertion needs scoping to
  non-master rows.
- Context: #1146 (FR-015 scoping), spec 040 (calibration masters)

srobroek:
[...]

Issue #1102, opened 2026-07-19
fix: target recommendations pick an arbitrary item when a folder has several


srobroek:
## Problem

`crates/app/inbox/src/target_recommendations.rs` resolves a `SourceGroup` target by taking the **first item the database happens to return**:

```rust
RecommendationTarget::SourceGroup(sg) => {
    let ids = repo::list_item_ids_for_source_group(pool, sg).await.map_err(db_err)?;
    ids.into_iter()
        .next()
        .ok_or_else(|| not_found(format!("no inbox items for source group: {sg}")))
}
```

There is no ordering guarantee and no selection rule — it silently picks one sibling and recommends targets for that, discarding the rest.

## Why it matters

A source group routinely holds **more than one** item. A mixed folder classified into `light`/`dark`/`flat` produces three sibling items sharing one `source_group_id`. Today the placeholder row usually absorbs this, so the arbitrariness is masked rather than absent.

The user-visible effect is a recommendation computed from one arbitrary frame-type group and presented as the folder's recommendation. Whether it is the right one is luck, and it can change between runs if row order changes.

## Why now

Spec **058** (`specs/058-inbox-drop-parent-items/`, merged) removes placeholder rows: a folder becomes N sibling items with no aggregate row. At that point *every* multi-group folder hits this path and "the folder's item" stops existing as a concept — so an arbitrary pick becomes the normal case rather than an edge case.

## Suggested approach

Needs a product decision before code:

- recommend across **all** siblings and merge/deduplicate, or
- recommend per sibling and let the caller choose, or
- define an explicit, deterministic selection rule and document why one sibling is representative

Any of these beats `.next()`. Erroring on ambiguity is also defensible — better a clear failure than a confident wrong answer.

## Notes

- Found while writing spec 058. Latent today, load-bearing after it.
- Related: #1101 (silently swallowed writes in the same pipeline), #711 (list/detail disagreement).
- Inbox territory — the `run-jc0717` lane may want to claim or sequence this.

srobroek:
## Blocks spec 058 Phase 5 — needs a product decision, not a code fix

Recording this from the 058 plan gate so the dependency is visible from this
issue rather than only from `tasks.md`. This is **T003** in
`specs/058-inbox-drop-parent-items/tasks.md`.

### Why 058 makes this urgent

`resolve_item_id` picks an arbitrary sibling via `ids.next()`. Today that is
merely ambiguous, because a folder still has a placeholder "parent" row that
[...]

PR #1194, opened 2026-07-20
fix: Inbox rows no longer claim a classification they do not have

srobroek:
## Summary

Scanning a folder used to create a hidden bookkeeping row alongside the real
ones. That row carried no frame type and no classification, yet it was the row
you actually selected and confirmed for an ordinary folder — and after a folder
split it lingered in the list advertising a classification it never had. The
result was a visible, reproducible contradiction: a list row reading
**CLASSIFIED** that, when opened, showed `unclassified`, a blocking "Frame types
required" banner, and a disabled Confirm button.

The badge was not rendering the wrong thing. It was rendering the database
faithfully, and the database contained a false statement. This removes the
hidden row entirely so that every row in the Inbox is a real, actionable item
that states only true things about itself.

## What is new

- A folder holding one kind of frame produces exactly one Inbox row. A folder
  holding lights and darks together produces two — one per frame type, and
  nothing else.
- Every row you can select has a frame type, a truthful badge, and a Confirm
  outcome that concerns only its own files.
- Confirming one part of a mixed folder leaves the other parts untouched, with
  their own states and their own plans.
- A folder that has been scanned but not yet classified stays visible in the
  queue instead of disappearing, and cannot be confirmed until there is
  something to confirm.
- The Inbox list can be grouped by folder, so a folder that split into several
  frame types reads as "one folder, N types" rather than N near-identical rows.

## Status

**Draft — implementation is in progress via orchestrated agents. Do not review
line-by-line yet.**

Not yet verified: the three Real-UI journeys that guard this surface have not
been run. A baseline run against `main` is in progress.

Knowingly not delivered: superseded-plan invalidation. That decision stands but
its mechanism ships separately.

Some commits on this branch are unsigned — subagents cannot reach the signing
agent in this environment.

## Spec Context

Implements `specs/058-inbox-drop-parent-items/`. Resolves the unsplit remainder
of #711 Instance A. Supersedes the read-side workarounds in #1038 and #1081,
which are removed rather than corrected — with no aggregate row there is
nothing to suppress.

Decisions taken on this branch:

- **#1102** — target recommendations become per-sibling. `ids.next()` is
  deleted, not replaced with a better arbitrary pick, because any deterministic
  selection rule re-introduces the designated primary D-002 forbids.
- **#1178** — source-group rows are counted; selection follows the unique
  successor then the folder group; siblings order by `group_key`; a
  heterogeneous needs-review bucket splits (scope addition, T050-T052).

Open risks:

- **#1177** — D-004 licenses stranding open plans on the basis that "there are
  no current installs" and writes its own expiry condition. Five public
  releases now exist. Proceeded on D-004 as written because installer
  downloads total ~4 across all releases and the reachable blast radius is a
  development database, but it is a product call.
- **#1157** — placeholder-scoped queries share `group_key = ''` with calibration
  master rows. Two executable predicates need `AND is_master_item = 0`, not the
  single one the issue names.
- SC-009 is knowingly unmet; the real exit bar is the other eleven success
  criteria. See T043.

Refs #711, #1102, #1157, #1177, #1178.


srobroek:
## Heads-up: `0075` is claimed by two open PRs

`#1048` (`feat/056-onboarding-redesign`) and `#1194` (`spec/058-inbox-drop-parent-items`) each add a **different** migration at version **0075**:

| PR | Migrations |
|---|---|
| #1048 | `0075_onboarding.sql`, `0076_drop_guided_flow_state.sql` |
| #1194 | `0074_inbox_needs_review.sql`, `0075_source_group_file_count.sql` |

`main` is currently at `0073`, so both look fine in isolation and neither shows a merge conflict — the files have different names, so git merges them cleanly. But `sqlx::migrate!` keys on the numeric version: whichever lands second produces two entries at `0075` and **`migrate()` aborts on a fresh database**. It will not reproduce on an existing dev DB that already has the first one applied.

Whoever merges second needs to renumber (`#1048` → `0076`/`0077`, or `#1194` → `0076`) and update any doc references to the old numbers.

Noted because #1048 already renumbered once (`0072`/`0073` → `0075`/`0076`) to escape a collision with `calibration_master_path`/`_archive`, and landed on a number a second in-flight branch had meanwhile taken. The gap at `0074` is the tell.

srobroek:
## Session status — 28 of 53 tasks, still DRAFT

Branch assembled and green: **2516/2516 workspace tests**, clippy and fmt clean.

### The MVP blocker is solved

`classify_source_group` (`882ec25e`) lets a folder be classified without a
pre-existing item row. Previously: no item -> no evidence -> no item, so deleting
the placeholder made every scanned folder permanently unclassifiable **while
every static check stayed green**. The test that pinned that gap is now inverted.

**T020 (delete the placeholder) is unblocked but not done**, and still needs the
UI half: the backend emits a `sourceGroups` array with **zero consumers in app
code**.

### Merged in

`spec/058-targetrec`, `spec/058-docs`, `spec/058-us4`, and `origin/main`
(which brings #1099). Two conflicts resolved: both sides had added fields to
`InboxListItem` (kept both, bindings regenerated rather than hand-merged), and
both had appended to `reclassify.rs`'s test module.

### Do not merge yet

- US1 is incomplete — T019/T020/T022/T024 outstanding.
- **SC-005 has not been run against this branch.** Baseline on `main` is 24/24
  green, so the comparison is ready to make.
- **SC-009 is knowingly not met**; the exit bar is the other eleven criteria.

### Open decisions, all filed

- **#1177** — D-004's greenfield licence names a precondition that no longer
  holds (five public releases exist).
- **#1207** — the motivating symptom is fixed on `main` by #1099, so the spec's
  justification narrative needs restating. The feature is still justified: SC-003
  is violated today, and #1099 is the *third* read-side patch the spec's own
  Rejected Alternatives warns against.
- **#1178** — four requirement gaps, all decided.
- **T026** — `confirm()` has no frame-type gate; a row with no authoritative
  frame type can generate a filesystem plan. Decide at T020, when the exposed
  row shape disappears.

Refs #711, #1102, #1157, #1177, #1178, #1207.

srobroek:
Flagging a risk before this leaves draft, from investigating #1202. **This may not be a problem — but it is cheap for you to check and expensive for anyone else, because you are already in this code.**

## The concern

T020 removes the hidden placeholder row. #1202 is a **deterministic** red on `main` (`inbox_ui_unsplit_unclassified_folder_badge_is_not_classified`) whose symptom is that the Inbox list comes back **completely empty** after classify — `Type cells were []`, not a wrong badge.

I eliminated all three backend explanations against `origin/main`:

1. **`exclude_split_placeholder!` cannot cause it.** The fixture is one `IMAGETYP="Frame Unknown"` file, which routes to the `__needs_review__` sentinel (`classify.rs:699`) — exactly one distinct non-empty `group_key`. The predicate needs `> 1`. And even when it fires it hides *one* row, so it can never empty a list.
2. **The state filter cannot cause it.** `classify.rs` only ever writes `classified` or `pending_classification`; both are in `list_unacknowledged_across_roots`'s `WHERE`. It never writes `resolved`.
3. **There is no `frame_type` filter on the list.** (Asymmetry worth knowing: `inbox_stats` and `count_distinct_inbox_folders` *do* filter `frame_type IS NOT NULL`; the list does not.)

So the repository layer should be returning rows, and something above it is not rendering them.

## Why that matters for T020 specifically

My leading hypothesis is that the **sub-items** are the unrenderable part — the same family as #854 / #1021, where `inbox_source_groups.lane` (`move`/`catalogue`) was conflated with `inbox_items.lane` (`fits`/`video`) and produced empty sub-item sets.

If that is right, then **the placeholder is currently the only row that renders at all**, and removing it does not fix the empty list — it makes it *guaranteed*. 058 would ship the defect while believing it fixed it, and the journey that would have caught it is the very journey 058 rewrites.

If instead the placeholder is what is wrongly hidden, 058 genuinely fixes this and there is nothing to do.

## The check

Classify the #1202 fixture (one file, `IMAGETYP="Frame Unknown"`, target + filter + DATE-OBS present) and read `list_unacknowledged_across_roots` directly at Layer 1, on your branch:

- **Rows returned** => backend fine, defect is above it. Confirm your post-058 list still renders a row for this input.
- **No rows** => the fixture assumption is wrong and #1202 is stale; say so and it can be closed.

## Also worth confirming

#1202 notes that T016 is framed as "use the correct helper variant per step", which does not capture that this journey's *subject* is being deleted. After T020 the assertion should become "**no such row exists at all**", not "the aggregate row is not mislabelled". Worth checking T016 covers that rewrite explicitly.

Full analysis and the elimination chain: https://github.com/platevault/platevault/issues/1202#issuecomment-5019602968

No action needed from you beyond the probe — I am not asking you to take on #1202, only to avoid inheriting it silently.

srobroek:
Conflict with `main` — head SHA `5fc3c42dae`, main at `da70b292`.

`mergeable` is CONFLICTING / `mergeStateStatus` DIRTY. Conflicting files:

- `crates/app/inbox/src/classify.rs`
- `crates/app/inbox/src/reclassify.rs`

Cause: `da70b292` — "fix(inbox): needs-review names the attribute that is actually missing (#1126)" — landed on main and touches both files. This PR reworks the same classification path, so the needs-review reason wording from #1126 needs to be carried into the new structure rather than merged verbatim.

Note this PR is currently a draft, so no action is implied — flagging it so the conflict is not a surprise when it comes out of draft. GitHub runs no PR CI while a PR is CONFLICTING, so this one has no usable check signal until the conflict is resolved.

No file overlap with #1260 (`refactor/994-split-inbox-detail`), the other in-flight Inbox change — that one is confined to `apps/desktop/src/features/inbox/`.

srobroek:
Cross-lane note from the #1208 / #1252 work — two things that may matter to this PR. Not review comments, and no action needed from me.

**1. A premise in the spec text looks wrong.**

`spec.md` states that `inbox_ui_unsplit_unclassified_folder_badge_is_not_classified` **passes** on `main`, having been fixed by #1099 (`ef90b074`).

The bisect on #1208 says it fails on `main`. Both dispatched refs failed the same two shards on that same test:

| ref | contains | ubuntu 2/4 | windows 2/2 |
|---|---|---|---|
| `bisect/1208-before-1131` @ `ef90b074` | #1099, not #1131 | FAIL | FAIL |
| `bisect/1208-after-1131` @ `6fa1bf55` | #1099 and #1131 | FAIL | FAIL |

Note the first ref *is* `ef90b074` — the commit the spec credits with fixing it. It failed there.

This was easy to miss because main's E2E runs were being cancelled by a concurrency-group bug before they could report (#1268 fixes that), so main looked green when it was merely silent. Since this PR modifies `inbox_ui_journeys.rs`, it may fix the journey in passing — but the spec probably should not record it as already-fixed, given SC-003 is the stated measure anyway.

**2. `insert_source_group_with_item` in `classify.rs` is broken, and #1266 fixes it.**

The helper's `registered_sources` INSERT violates the 0006 schema four ways and had never inserted a row, so 17 classify tests were joining an empty table. #1266 is confined to that file's `#[cfg(test)]` module.

Your new `seed_inbox_source` (via `register_source_batch`, `.unwrap()`ed) is the better pattern and is unaffected. I deliberately did **not** reroute the old helper through it, because `register_source_batch` generates its own UUID and the helper would have to stop taking a caller-chosen `root_id` — changing all 17 call sites in a file you are mid-rewrite on.

**#1266 is happy to sequence behind this PR.** It is a trivial textual conflict at worst, and you are rebasing anyway.

srobroek:
## Responses to the cross-lane notes

Conflict resolved and all four points actioned. Head is now `af01a8a5`, `mergeable` is MERGEABLE, so this PR has usable check signal again for the first time.

### 1. The #1202 probe — answered, in the safe direction

Ran it at Layer 1 on this branch. `list_unacknowledged_across_roots` returns **2 rows** for the #1202 fixture:

```
#1202 probe: 2 row(s) total — 1 placeholder, 1 sub-item; keys ["", "type=unknown"]
```

So a **materialized sub-item survives the query** alongside the placeholder. Your first branch holds: the backend is fine, the defect is above it, and the placeholder is *not* the only row that renders. **T020 is safe on this axis** — deleting it leaves the `type=unknown` row.

Pinned as `unsplit_unclassified_folder_still_lists_rows_after_classify_1202`. Two things I did deliberately, both because of your notes:

- **The assertion is on row composition, not `!rows.is_empty()`.** The latter is satisfied by the placeholder alone — i.e. it would pass in precisely the case that makes T020 unsafe. It partitions on `group_key.is_empty()` and asserts a non-placeholder row exists.
- **The `registered_sources` row is inserted by hand and in full.** Your point 2 is what made me check: the query JOINs that table for `r.path`, so had I reused `insert_source_group_with_item` the probe would have returned zero rows and I would have reported "backend is broken" — the exact false negative. Comment on the test says why, so nobody simplifies it back onto that helper.

I have not taken on #1202 itself.

### 2. `spec.md` premise — corrected

You're right, and the correction is in. The spec claimed `inbox_ui_unsplit_unclassified_folder_badge_is_not_classified` passes on `main` fixed by #1099; your bisect shows it failing on `ef90b074` itself. Now recorded with the cancelled-runs explanation (#1268) and the note that #1099 fixed the badge *expression* while the journey's actual symptom is an empty list. Scope claim stays on SC-003, which is measurable without that journey.

### 3. The `main` conflict — resolved

`#1126` needed carrying into the new structure rather than merging verbatim, as you said. What landed:

- **`classify.rs`**: adopted `missing_mandatory_for_file` — it folds the no-frame-type case into one helper, a real simplification. Did **not** adopt the `SENTINEL_NEEDS_REVIEW` key it feeds, since 058 retires that sentinel; the verdict goes to the `needs_review` bool and `group_key` stays classification identity.
- **`reclassify.rs`**: took main's `needs_review_missing`. The hardcoded `vec!["frameType"]` *is* the #1114 bug, so ours was the wrong side of that conflict.
- Git had interleaved two independently-added tests mid-call. Both kept; #1126's `v2_partially_resolved_...` reconstructed whole and re-keyed from `group_key == SENTINEL_NEEDS_REVIEW` to `frame_type.is_none()`. Two-direction control to prove the adaptation didn't make it vacuous: restoring the hardcoded value fails it with `must not ask again for the frame type the user just supplied: ["frameType"]`.

### 4. Migration `0075` — still live, not yet actionable

`#1048` is still open and `main` is at `0073`, so nothing has collided yet. Flagging for whoever merges second: this PR holds `0074_inbox_needs_review.sql` + `0075_source_group_file_count.sql`.

### 5. `#1266` — sequencing behind this is right

Agreed, and thanks for not rerouting the 17 call sites in a file mid-rewrite.

---

**Still DRAFT, still do not merge.** US1 remains incomplete and SC-005 has not been run against this branch. Newly recorded here: `classify_source_group` has no tauri command, no contract entry and no binding, so a source-group row is now *visible* but not yet *actionable* — that gap is a hard prerequisite for T020, not a nicety, and the contract for `inbox.classify.sourceGroup` is written but the command is not built yet.

srobroek:
**Failing step is `Rust format` — just needs `cargo fmt`.**

`Unit + integration (L1+L2) — ubuntu-latest` fails on that single step; nothing in the test suite is red.

CI runs `cargo fmt --all --check` (`justfile:15`), so:

```
cargo fmt --all
```

and commit the result.

(Full logs weren't retrievable while the parent run is still in progress — this is from the job's per-step conclusions.)

srobroek:
## Merged `main` again (22 commits) — and a cross-lane warning about #1186

Head `c8c468ad`, MERGEABLE.

### The `main` conflict was semantic, and both sides were right

#1275 restated the #1099 justification on `main`; this branch carried a correction saying #1099 did *not* close the journey, citing the #1208 bisect. Both are correct **about different commits**: `ef90b074` and `6fa1bf55` both **predate #1206**, which repaired the journey's stale selector. #1099 fixed the badge expression; the journey went on failing on a selector until #1206.

Took `main`'s wording and rewrote this branch's note into that reconciliation rather than leaving the spec arguing with itself. Confirming evidence: **the full Real-UI E2E suite passed on this branch at head `33f78450` — all six shards, including the ubuntu 2/4 and windows 2/2 the bisect saw fail.**

### ⚠️ Warning for anyone else merging `main`: #1186 invalidates your Vite cache

This cost me a while and will hit every in-flight branch, so it's worth stating plainly.

After merging, **three** desktop test files failed — two of which this branch does not touch:

| file | on `main`? | result on `main` | result here |
|---|---|---|---|
| `devSurface.release.test.ts` | yes | pass | **fail** |
| `GuidedOverlay.test.tsx` | yes | pass | **fail** |
| `InboxPage.statsFollowFilter.test.tsx` | no (058-only) | — | **fail** |

All three failed identically — `Test timed out in 15000ms` on an `await import(...)`. It reads exactly like a real regression, and two files being main's own code makes it look like the merge broke them.

It is a **stale** Vite optimizer cache, from #1186 changing optimizer caching. The trap: **re-running does not fix it**, because the cache is stale rather than cold — which is what convinced me it was real. The fix is:

```bash
rm -rf node_modules/.vite apps/desktop/node_modules/.vite
CI=true pnpm install
```

Verified in both directions: a clean `origin/main` worktree passes the two shared files, and this branch passes all three after the clear (4/4 on repeat). Nothing was wrong with the code.

### #1266 landed cleanly

It merged without touching the #1202 probe. Not duplicating it was the right call — thanks for taking it.

---

**Still DRAFT, still do not merge.** Workspace 2527/2527, desktop 1875/1875 across 205 files, clippy `-D warnings` clean, both fmt gates clean, tsc clean.

One more gate note, since it bit me: `cargo fmt --all --check` **cannot see** `apps/desktop/src-tauri/src/bootstrap/specta.rs` — `cargo fmt` only walks `mod` declarations and that file is `include!`d, so CI runs a separate `rustfmt --edition 2021 --check` on it (`ci.yml:263`). Local `cargo fmt --all --check` is a false green for the one file every new Tauri command must be registered in.

PR #1196, opened 2026-07-20
fix: show the cleanup protection rule as text instead of hiding it in per-row tooltips

srobroek:
## Summary

- The Cleanup section's scan table rendered a padlock on every protected row whose tooltip is **one static sentence, identical on every row**. Since #1164 made those padlocks keyboard-focusable, a table with twelve protected rows produced **twelve tab stops that all announce the same thing**.
- That rule is now stated once, in text, above the table — so it's visible without hovering. Previously a sighted user had to discover it through a tooltip.
- With the consequence on screen, the per-row padlock becomes decoration (the adjacent "Protected" pill already marks the state in text), so it leaves the tab order.

## Why this isn't a regression of #1164

#1164 was correct: `Lock`'s reason was reachable by pointer only, and it states a consequence shown nowhere else. This fixes the *root* of that — the consequence is no longer shown nowhere else — rather than reverting the focusability.

The **protected-categories list is deliberately unchanged**. Each of its reasons is distinct per category (`"{category} are protected and excluded from cleanup plans"`) and the list is short, so each tab stop there carries new information and stays focusable. `Lock`'s new `decorative` variant is documented as only correct where the reason is stated nearby *and* identical across instances.

Note the two blocks describe genuinely different protections — categories are *never proposed* for cleanup, while table rows *do* appear and merely need acknowledgement — so they aren't redundant with each other.

## Implementation notes

- No new i18n string: the note reuses the existing `projects_cleanup_row_protected_hint` key.
- No new CSS class: reuses the shared `.pv-text-muted` utility.
- `Lock` gains one parameterised variant rather than a second component.

## Verification

- 291 tests pass across 31 files (`src/ui/Lock.test.tsx` + all of `src/features/projects/`).
- New tests assert the decorative variant is `aria-hidden`, carries no `tabindex`, exposes no `note` role, and still renders the glyph.
- **Negative control:** removing `aria-hidden` from the decorative branch fails that test and only that test.
- `typecheck` and `biome check` clean.

## Context

This resolves the open question flagged for a reviewer in #1164 and recorded as ambiguity 2 in the session-3 decision log, rather than leaving it to be re-litigated in the next a11y sweep.

### DIFF
--- a/apps/desktop/src/features/projects/OutputsCleanupSections.tsx
+++ b/apps/desktop/src/features/projects/OutputsCleanupSections.tsx
@@ -176,7 +176,10 @@ function candidateRow(candidate: CleanupCandidate, index: number) {
       : candidate.reason,
     protection: isProtected ? (
       <span className="pv-cleanup-scan__protected-cell">
-        <Lock reason={m.projects_cleanup_row_protected_hint()} />
+        {/* Decorative: the hint is one static sentence, identical on every
+            protected row, and is stated once above the table. Giving each row
+            its own tab stop would repeat that same announcement N times. */}
+        <Lock decorative />
         <Pill variant="warn">{m.settings_cleanup_protection_protected()}</Pill>
       </span>
     ) : parsed ? (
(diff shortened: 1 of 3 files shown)


srobroek:
The failing `Real-UI E2E` shards (ubuntu 2/4, windows 2/2) are **not from this PR** — tracked as #1208.

Same two shards fail identically on #1183 and #1196, whose diffs are unrelated (#1183 adds only a `tsconfig` + a one-line npm script and contains no runtime code). #1183's shards were re-run and failed again, so it is deterministic rather than flaky.

Root cause per #1208: #1099 and #1131 merged 72 seconds apart and were never co-tested — each passed all six shards on its own branch, but `main` never runs the Real-UI E2E shards, so the merged combination went unverified.

Issue #639, opened 2026-07-12
Calibration Matching settings pane persists nothing — every save silently rejected


srobroek:
**Severity:** P1 · **Area:** Settings → Calibration Matching

**Repro**
1. Open Settings → Calibration Matching.
2. Toggle "Gain" match-required — the UI flips.
3. Read the backend (`calibration_tolerances_get`) — `requireSameGain` is unchanged.
4. Navigate away and back — the toggle reverts. Same for Camera/Binning/Offset toggles, the temperature and aging number inputs, and the pane's own "Restore defaults".

**Expected vs Observed**
Expected: every control persists via `calibration.tolerances.update` (Journey 8 tolerances contract — "change temperature/aging/offset requirements … persists across restart").
Observed: the backend value never changes, no error is shown anywhere; the pane is optimistic-only and silently desyncs from the backend until reload.

**Root cause**
The pane always sends `exposureToleranceS: null` for its one unsurfaced field (`apps/desktop/src/features/settings/CalibrationMatching.tsx:92`), but the Rust DTO field is a non-optional `f64` (`crates/contracts/core/src/calibration_tolerances.rs:27`). Every `calibration_tolerances_update` therefore fails Tauri arg deserialization with `invalid args 'request' for command 'calibration_tolerances_update': invalid type: null, expected f64` — and the rejection is swallowed by the best-effort catch (`CalibrationMatching.tsx:95-97`). `handleRestoreCalibration` sends the same `null` (`CalibrationMatching.tsx:151`), so Restore defaults fails too (its error is swallowed by `RestoreDefaultsBtn`, `SettingsKit.tsx:146-148`). The generated bindings still claim `exposureToleranceS: number | null` (`apps/desktop/src/bindings/index.ts:8213`) — bindings/contract drift hides the bug from typecheck.

**Evidence**
Screenshot `a8-calmatch.png` (pane). Live run on the real Windows app: UI Gain toggle flipped off while `calibration_tolerances_get` kept `requireSameGain:true`; direct invoke of `calibration_tolerances_update` with the pane's exact payload returned the `invalid type: null, expected f64` error verbatim.

Part of #632. Found by untouched-controls discovery Sweep-A (design review 2026-07-11, second pass).

PR #1199, opened 2026-07-20
test: calibration matching settings test waits for loaded values before asserting

### LINKED ISSUE #1159: Verify the vitest timeout raise actually reduced Windows/macOS flake rate
## What changed

PR #1082 set `testTimeout`/`hookTimeout` to 15s in `apps/desktop/vitest.config.ts`. Before that the suite ran at vitest's **5s default** with no worker cap.

Reproduced locally: at load average ~1 the full suite is green; saturating all 12 cores makes **unrelated** suites fail with a bare `Error: Test timed out in 5000ms` — `GuidedOverlay` and `devSurface.release` being the reliable casualties. Same load, before and after:

| `testTimeout` | result |
|---|---|
| 5s (default) | 2 files fail, `Test timed out in 5000ms` |
| 15s | 186 files / 1729 tests pass |

CI runners are 2–4 cores running ~190 suites in parallel — far more contended than that local box.

## Why this needs verifying rather than closing

The hypothesis is that a share of this project's historical "flaky on Windows/macOS" noise was never flaky tests at all, but a missing config line. If true, the flake rate should drop noticeably.

**I could not confirm it.** Two problems:

1. **No clean before/after sample.** Every PR merged after the fix landed (11:21) was green on Windows Integration — #1076, #1082, #1106, #1108 — but so were their post-fix reruns. Main's own runs kept getting cancelled by rapid successive merges, so there is no stable pre-fix baseline to compare against.
2. **Confounded.** #1082 *also* fixed the two specific racing tests (`AuditLog` export, `ResolverSettingsControl` toggle). So green-since-then is consistent with the timeout fix working, with the two test fixes being sufficient on their own, or with both.

## How to verify

Over the next couple of dozen main runs, count Windows/macOS `Integration (Layer 1)` failures whose error is a bare `Test timed out in Nms` rather than a real assertion. That signature should now be near-zero. If it is not, the diagnosis was incomplete and the remaining cause is something else — worker count, memory pressure, or a genuine hang.

Related: #1083 catalogues 17 latent load-state races. Those are a **distinct** class — they fail fast with a wrong assertion, not a timeout, so this change does not address them and their absence would not validate it.

srobroek:
## Summary
- Fixes a genuinely racy unit test in the calibration matching settings pane: it asserted before the fetched values had loaded, so it intermittently compared against the in-code default. Test-only change; no product behaviour is affected.

Residue triage in the same family as #1128 / #1109. One site in
`CalibrationMatching.test.tsx` genuinely races; the other seven in the file pass
unchanged under the same delay injection and were left alone.

## Mechanism — a vacuous hydration gate, not CI noise

`CalibrationMatching.test.tsx:238` gated the click on:

```ts
await waitFor(() => expect(offsetToggleInput()).toBeChecked());
```

The in-code default is `requireSameOffset: true` (`CalibrationMatching.tsx:41`)
and `makeTolerances()` also returns `true`, so the toggle is checked on **first
render**, before the mount fetch resolves. The gate passes instantly and proves
nothing about hydration.

`fireEvent.click` then runs `persist()` while `exposureTolerance` is still the
default `2.0` (`CalibrationMatching.tsx:70-72`), instead of the fetched `3.5`
that only lands in state once `calibrationTolerancesGet()` resolves
(`CalibrationMatching.tsx:96-98`). Hence `expected 2 to be 3.5`.

This is exactly the hazard the file's own helper docstring already documents at
lines 64-69, which is why `cameraToggleInput()` exists — the #639 tests just
never used it.

**Pre-existing and load-induced.** It surfaced on `coder-g-refactor-b5`, whose
diff touches only `apps/desktop/src-tauri/**`, `justfile` and `.github/**` —
zero frontend files — so that branch provably cannot be the cause.

## Before — delay injection reproduces it deterministically

120ms on `calibrationTolerancesGet` (the #1128 probe), applied to every site in
the file:

```
=== run 1
PASS (7) FAIL (1)
   AssertionError: expected 2 to be 3.5 // Object.is equality
=== run 2
PASS (7) FAIL (1)
   AssertionError: expected 2 to be 3.5 // Object.is equality
=== run 3
PASS (7) FAIL (1)
   AssertionError: expected 2 to be 3.5 // Object.is equality
=== run 4
PASS (7) FAIL (1)
   AssertionError: expected 2 to be 3.5 // Object.is equality
=== run 5
PASS (7) FAIL (1)
   AssertionError: expected 2 to be 3.5 // Object.is equality
```

5/5, and the exact error from the original report.

## Fix — test synchronisation only, no production change

Mock now returns the non-default `requireSameCamera: false` and the test gates
on the `cameraToggleInput()` hydration marker, then re-asserts the offset toggle.
No assertion weakened.

**The component is not buggy.** It round-trips `exposureToleranceS` correctly
once hydrated, and a click landing before the mount fetch resolves is the
deliberate `editedRef` optimistic-edit behaviour documented at
`CalibrationMatching.tsx:74-81` and covered by the "a slow initial fetch does not
clobber an edit" test. Only the test's gate was wrong.

## After (a) — 10 runs with the delay still injected

```
run  1: PASS (8) FAIL (0)
run  2: PASS (8) FAIL (0)
run  3: PASS (8) FAIL (0)
run  4: PASS (8) FAIL (0)
run  5: PASS (8) FAIL (0)
run  6: PASS (8) FAIL (0)
run  7: PASS (8) FAIL (0)
run  8: PASS (8) FAIL (0)
run  9: PASS (8) FAIL (0)
run 10: PASS (8) FAIL (0)
```

## After (b) — delay removed, full `apps/desktop` suite

```
PASS (1821) FAIL (0)
```

## Sibling sites — checked, deliberately left alone

All seven other tests in the file passed under the same 120ms injection:

- `:90` and `:165` gate on `not.toBeChecked()` with mocked `requireSameOffset:
  false`, which differs from the default — real gates.
- `:111` and `:138` already use the `cameraToggleInput()` marker.
- `:184` mocks a rejected fetch and asserts the default is kept — correct by
  design.
- `:211` is the deliberate manual-resolve slow-fetch test.
- `:252` ("restore-defaults sends the migration default (2.0)") does have a
  vacuous gate, but cannot race: `handleRestoreCalibration` sends `DEFAULTS`
  unconditionally (`CalibrationMatching.tsx:170-178`), so the asserted `2.0` is
  correct whether or not hydration landed. Left unchanged rather than
  mass-rewriting a passing test.

## Related

Context for #1159 (verifying whether the vitest timeout raise actually reduced
Windows/macOS flake rate): this failure was **not** a timeout, so raising
timeouts would never have fixed it and its disappearance from flake stats would
not be evidence either way. Does not close #1159.

Refs #1083. Not claimed to close any issue.

Commit is unsigned — subagents cannot reach the 1Password SSH signing agent; the
PR squash-merges so the unsigned commit never lands on main.

## Spec Context
- Spec: 211
- Branch: worktree-worktree-2113499

---

Supersedes #1197, which was branched from a contaminated base: the shared checkout had been switched to another lane's branch, so the PR carried unrelated in-progress wizard work (`SetupWizard.tsx`, `en.json`). This branch is the same single commit applied to a clean `origin/main`.


srobroek:
## Heads-up: this fix landed in parallel as #1214 — my fault

I hit the same failure (`expected 2 to be 3.5` on a loaded macOS runner), diagnosed it independently, and opened #1214 without first checking whether a branch already existed for it. #1214 has since merged, so `main` already carries this change.

I should have looked for prior work on `#1083` before writing it. Apologies for the duplicated effort.

## One thing yours has that mine does not

The merged version omits your second assertion:

```ts
await waitFor(() => expect(cameraToggleInput()).not.toBeChecked());
expect(offsetToggleInput()).toBeChecked();   // ← not in #1214
```

That line is worth keeping. Gating on the camera marker proves hydration landed, but it no longer asserts anything about the offset toggle — which is the control the test then clicks. Your version keeps that coverage; mine silently dropped it.

Suggested resolution, your call as the owner of this branch:

- **Reduce this PR to just that one assertion** on top of current `main` — small, and it restores what #1214 lost, or
- **Close it** if you consider the assertion redundant given the click's own outcome is asserted downstream.

Either way `main` is not currently failing on this test. The rest of your diff is already there.

srobroek:
Superseded by #1214, which merged at 06:55Z and is already on `main`.

Both PRs diagnosed the same defect independently and converged on the same fix: the test gated on `offsetToggleInput()`, but `requireSameOffset` is `true` in both the in-code default and the mocked payload, so the gate passed on first render and proved nothing about hydration. The click then raced the fetch and persisted the default `2.0` instead of the fetched `3.5`. Both switched the gate to `cameraToggleInput()` and added `requireSameCamera: false` as a hydration marker.

Closing this one rather than resolving the conflict, since landing it would re-apply a fix `main` already has.

One small thing #1214 does not carry, noted in case it is wanted later: this branch also kept an explicit `expect(offsetToggleInput()).toBeChecked()` after the hydration gate, preserving the original assertion that the offset toggle starts checked. Not a defect in #1214 — just an assertion that went away with the old gate.

PR #1200, opened 2026-07-20
test: catch tokens.d.ts drifting from tokens.css automatically

### LINKED ISSUE #1191: Nothing verifies tokens.d.ts stays in sync with tokens.css
Found during the design-refresh campaign (orchestration run `run-dsr`, node t6).

## Problem

`apps/desktop/src/styles/tokens.d.ts` is generated from `apps/desktop/src/styles/tokens.css` by `apps/desktop/scripts/gen-token-types.mjs`. **Nothing enforces that the committed `.d.ts` matches its source.**

- `gen-token-types.mjs` is wired only as the npm script `tokens:types` in `apps/desktop/package.json`.
- `grep -rn 'tokens:types' .github/workflows/` returns no hits — no CI job runs it or checks freshness.
- The only two files in the repo referencing `tokens.d.ts` / `PvTokenName` are the generated file itself and its generator.
- `theme.tokens-drift.test.ts` sounds like it covers this but does not. It verifies that `theme.ts`'s hardcoded `SPACING_BASE_PX` / `TEXT_SCALE_BASE_PX` match `packages/tokens/foundation.css` — an unrelated spec-055 concern, needed because jsdom cannot resolve CSS custom properties.

## How it surfaced

Node t6 added the `--pv-destructive` / `-bg` / `-bg-hover` triplet to `tokens.css` but did not regenerate `tokens.d.ts`. The branch was internally inconsistent — the tokens existed in CSS but were absent from the `PvTokenName` union — from that commit until a later rebase happened to catch it **by hand**. Nothing failed. No gate fired.

That was luck. The same mistake on any future token addition would ship silently, leaving new tokens unreachable through the typed API.

## Suggested fix

A test in the spirit of the existing `theme.tokens-drift.test.ts`: run the generator against `tokens.css` and assert the output byte-matches the committed `tokens.d.ts`, failing with a message telling the developer to run `pnpm tokens:types`. Running under vitest means it is covered by existing CI without new workflow wiring.

## Acceptance

- Adding a token to `tokens.css` without regenerating `tokens.d.ts` fails CI.
- The guard is verified in both directions: it passes on a synced pair and fails on a deliberately desynced one.
- Current `tokens.css` / `tokens.d.ts` on main are confirmed in sync.

srobroek:
## Summary
- Adds a vitest guard that fails if `tokens.d.ts` is out of sync with `tokens.css`/`foundation.css`, telling the developer to run `pnpm tokens:types`.
- Refactors `gen-token-types.mjs` to export its parsing/rendering as pure functions so the test reuses the real generation logic instead of duplicating it.
- No CI wiring change: the guard is a regular vitest test, covered by the existing test job automatically.

Closes #1191

**Do not merge yet** — held behind nodes t6, t7, t8 (in-flight rebases/PRs touching `tokens.css`/`tokens.d.ts`) to avoid moving `main` under them.


srobroek:
CI triage — head SHA `a4756666cf4cc5909327efaa06355922fac3f695`

Five failing checks, from two independent causes.

**1. `Integration (Layer 1) — windows-latest` (job 88274513746), step `Frontend unit tests` — this PR's own failure.**

```
FAIL src/styles/tokens.types-drift.test.ts [ src/styles/tokens.types-drift.test.ts ]
SyntaxError: Invalid or unexpected token
Test Files  1 failed | 199 passed (200)
```

The file this PR adds collects **0 tests** — it fails to parse at all, before any assertion runs. Windows-only; the same file parses fine on ubuntu and macOS (both green). A Windows-only `SyntaxError` in a generated-content test points at a line-ending or escaping issue in what `scripts/gen-token-types.mjs` emits or in what the test embeds — worth checking CRLF handling and any unescaped backslash/path literal.

**2. Real-UI E2E — not this PR's fault.**

`Real-UI E2E shard — ubuntu-latest 2/4` (job 88274564982) and `shard — windows-latest 2/2` (job 88274564968) both fail on the same single test:

```
test inbox_ui_unsplit_unclassified_folder_badge_is_not_classified ... FAILED
test result: FAILED. 0 passed; 1 failed; 0 ignored; 5 filtered out
```

The two `Real-UI E2E — ubuntu-latest` / `— windows-latest` jobs are aggregators; their `Check shard results` step fails only because the shard above failed. So this is one failure, not four.

That test is in the Inbox classification area. This PR touches only `apps/desktop/scripts/gen-token-types.d.mts`, `apps/desktop/scripts/gen-token-types.mjs`, and `apps/desktop/src/styles/tokens.types-drift.test.ts` — no Inbox code, no backend, no token values — so it cannot be the cause.

Only item 1 is actionable here.

PR #1201, opened 2026-07-20
fix: inbox end-to-end check looks for the current Type-column class

srobroek:
## Summary

- The Inbox Type-column E2E helper still looked for the old `alm-` CSS class after the `--alm-*` → `--pv-*` rename, so it matched no elements and the journey guarding the unclassified-folder badge failed on Linux and Windows.

## Root cause

#1162 renamed the namespace and updated `InboxList.tsx`, whose Type cell now renders `pv-inbox-row__classification` (`InboxList.tsx:520`). It missed this Rust E2E helper, which kept querying `.alm-inbox-row__classification` (`inbox_ui_journeys.rs:502`). `querySelectorAll` matched nothing, so `classification_cell_labels` returned an empty list and `inbox_ui_unsplit_unclassified_folder_badge_is_not_classified` — the journey protecting the #1099 fix — failed on the ubuntu 2/4 and windows 2/2 shards.

## This is a defect on main, not on the PR that surfaced it

It first appeared on #1169, which changes only `apps/desktop/src-tauri/**`, `justfile` and `.github/**` — no frontend and no test files — so it cannot affect a DOM selector. #1169 was simply the first PR to complete a full Real-UI E2E run after #1162 landed. Merging this unblocks that gate.

Related: #1162 (the rename), #1099 (the fix the journey protects).

### DIFF
--- a/crates/e2e-tests/tests/inbox_ui_journeys.rs
+++ b/crates/e2e-tests/tests/inbox_ui_journeys.rs
@@ -475,7 +475,7 @@ async fn inbox_ui_unclassified_gate_bulk_reclassify_unblocks_confirm() -> anyhow
 
 /// Trimmed, lowercased text of every Type-column cell currently rendered in
 /// the Inbox list. That cell carries no `data-testid` of its own — it is the
-/// `span.alm-inbox-row__classification` inside each row (the `type:` cell in
+/// `span.pv-inbox-row__classification` inside each row (the `type:` cell in
 /// `apps/desktop/src/features/inbox/InboxList.tsx`), so it is located by class.
 ///
 /// Callers MUST compare with exact equality, never `contains`: "unclassified"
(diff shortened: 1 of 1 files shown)


srobroek:
Closing as a duplicate of #1206 — no reflection on the fix itself, which is correct.

Both PRs are **byte-identical**: the same two lines in `crates/e2e-tests/tests/inbox_ui_journeys.rs`, renaming `.alm-inbox-row__classification` to `.pv-inbox-row__classification` plus the matching doc comment. Two lanes independently found and fixed the same breakage, which is itself decent corroboration that the diagnosis is right.

#1206 is being merged instead for three reasons:

1. **Conventional-commit prefix.** Squash-merge makes the PR title a changelog entry. This change touches only `crates/e2e-tests` and has zero user-visible impact, so `test:` (as in #1206) correctly produces no changelog line, whereas this PR's `fix:` would emit a release note telling users a product bug was fixed — which is not true.
2. **Issue linkage.** #1206 carries `Fixes #1202`, so the tracking issue auto-closes on merge. This PR has no ref.
3. **Recorded evidence.** #1206's body documents the reproduction: the journey failed deterministically at 144s (reproduced twice), and the full Layer-2 suite is 24/24 green in 152s afterwards.

For context on why this mattered: the stale selector matched nothing, so `classification_cell_labels` returned an empty list and the journey failed with the literal message `Type cells were []`. That single test was reddening every open PR in the repo — #1169, #1180, #1126 and #1189 all failed on it despite none of them touching inbox code.

PR #1210, opened 2026-07-20
test: log panel follow-scroll test waits for settings before toggling Follow

### LINKED ISSUE #1209: fix: log panel follow-scroll test races the async settings load on loaded CI
## Summary

`apps/desktop/src/app/LogPanel.followScroll.test.tsx:244` fails intermittently on loaded CI runners:

```
AssertionError: expected '↓ Follow' to be '— Follow'
 ❯ src/app/LogPanel.followScroll.test.tsx:244:45
```

Observed on `Integration (Layer 1) — macos-latest` (job 88268919053). The same "Frontend unit tests" step is intermittently red on `main` too, with other tests (e.g. `CalibrationMatching.test.tsx` in run 29704696843), so this is one instance of a broader loaded-runner flake family.

## Root cause

The panel's initial follow-tail state arrives **asynchronously**, from the mocked `settingsGet` (`rememberFollowLogs: true`). The component first renders with its default, then flips to "on" when that promise resolves.

The two sibling tests in this file gate on that settling before they interact:

- `pauses follow when the user scrolls up...` — waits for `aria-label` = `Follow tail on (click to pause)` (lines 163–168)
- `does not pause when scrollTop stays within the near-top threshold` — same gate (lines 209–214)

The failing test, `re-enabling Follow resumes at the newest row even after a stale scroll-pause (#832)`, has **no such gate**. It clicks the follow button at line 242 immediately after `'Second entry'` renders. When the runner is contended and the settings promise has not yet resolved, the click toggles follow off and the subsequent resolution overwrites it back to on — so the button reads `↓ Follow` where the test expects `— Follow`, and the `waitFor` at 243 exhausts its window.

This is a pure test-harness race. There is no product defect behind it: the assertion is correct, it is just made against state the test never waited to stabilize.

## Why the previous fix did not cover it

`4c04b606` ("log panel follow-scroll test no longer fails intermittently on loaded CI", #1118) fixed the *same class* of race in the *same file*, but at the `scrollTo` assertion around line 276. It converted a sync `expect` into a `waitFor`. It did not touch the missing readiness gate before the first click, which is a different point in the same test.

## Fix

Add the readiness gate the sibling tests already use, before the first `fireEvent.click(getFollowButton())`, mirroring lines 163–168 exactly. No assertion is weakened or removed — the gate only ensures the click lands on settled state.

## Verification

Reproduce by contending the machine (the failure is timing-dependent, not deterministic on an idle box); confirm the fix by running the suite cold, which is the condition that produced it.

srobroek:
## Summary

- Fixes #1209 — `LogPanel.followScroll.test.tsx` failed intermittently on loaded CI with `expected '↓ Follow' to be '— Follow'`.
- The panel's initial follow-tail state arrives asynchronously from the mocked `settingsGet` (`rememberFollowLogs: true`). This test clicked the Follow button immediately after the first row rendered, so on a contended runner the click landed before that promise resolved and was then undone by the resolution.
- Adds the readiness gate the two sibling tests in the same file already use (`aria-label` = `Follow tail on (click to pause)`) before the first click. Test-only; no product code touched, no assertion weakened or removed.

## Why the earlier fix missed it

`4c04b606` (#1118) fixed the same class of race in this file, but at the `scrollTo` assertion near the end of the test. The missing gate before the *first click* is a separate point in the same test, so that fix left this path exposed.

## Verification

Reproduced deterministically rather than inferred, by injecting a 250ms delay into the `settingsGet` mock to simulate a contended runner, then running both directions:

| Variant | Result |
|---|---|
| Without the gate + forced race | `1 failed` — `AssertionError: expected '↓ Follow' to be '— Follow'`, the exact CI error |
| With the gate + forced race | `3 passed` |
| With the gate, no injected delay | `3 passed` |

Frontend lint pipeline clean (Biome, ESLint gates, token contrast, i18n catalog).

## Spec Context
- Spec: 019 (T011)

### DIFF
--- a/apps/desktop/src/app/LogPanel.followScroll.test.tsx
+++ b/apps/desktop/src/app/LogPanel.followScroll.test.tsx
@@ -238,6 +238,19 @@ describe('LogPanel follow-tail scroll pause/resume (T011)', () => {
       expect(screen.getByText('Second entry')).toBeInTheDocument();
     });
 
+    // Gate on follow-tail having settled ON before toggling it, exactly as the
+    // two tests above do. The initial state arrives asynchronously from
+    // `settingsGet` (`rememberFollowLogs: true`), so a click fired before that
+    // promise resolves is undone by the resolution — leaving the button on
+    // '↓ Follow' where the assertion below expects '— Follow'. Passes on an
+    // idle machine, fails on a contended runner (#1209).
+    await waitFor(() => {
+      expect(getFollowButton()).toHaveAttribute(
+        'aria-label',
+        'Follow tail on (click to pause)',
+      );
+    });
+
     // Turn follow off first (repro starts with Follow inactive).
     fireEvent.click(getFollowButton());
     await waitFor(() => {


srobroek:
Closing in favour of #1251, which fixes the same race in the same test and does it better.

I diagnosed this independently from a macOS CI failure and opened this PR without checking for existing work on the #1083 race sweep — the second time today I've duplicated another lane (see #1199). Deferring here rather than merging mine.

Why #1251 is the one to keep:

- It names the precise mechanism: `followLogs` initialises to `false` (`LogPanelContext.tsx:83`) and hydrates asynchronously from `settings.get('advanced')` (`:88`). My comment said only "the initial state arrives asynchronously", which is true but vaguer about *why* the click is lost.
- It explains the resulting sequence exactly — click lands pre-hydration → toggles `false`→`true` → late hydration confirms `true` → button stays `↓ Follow`.
- It ties the gate to the incoming `alm/no-vacuous-waitfor` rule (#1180), explaining why *this* `waitFor` is the legitimate "state has not arrived yet" shape rather than the vacuous one that rule rejects.

The two fixes are functionally equivalent — mine gated on the `aria-label`, #1251 gates on the button text — so there is nothing to port across.

Issue #1209 should be closed by #1251 rather than by this PR.

Issue #832, opened 2026-07-14
Activity panel: re-enabling Follow does not scroll back to the newest row


srobroek:
## Journey / step
Journey 13, Touch & validate: "follow mode keeps the newest row visible ... follow pauses on scroll-up and resumes from the newest row when re-enabled."

## Repro
1. Open the Activity panel (status-bar Log toggle).
2. Scroll the event list body (`.alm-logpanel__body`, the actual overflow:auto ancestor — note `.alm-logpanel__events` itself has `overflow-y: visible`) away from the top, e.g. `scrollTop = 400`.
3. Click the Follow chip to toggle it back on (icon changes from `— Follow` to `↓ Follow`, chip shows active state).

## Expected
Re-enabling Follow scrolls the list back to show the newest row (per spec: "resumes from the newest row when re-enabled").

## Observed
- Follow chip visibly toggles to the active/on state (`↓ Follow`).
- `document.querySelector('.alm-logpanel__event')` (first/newest DOM row) reports the current newest timestamp (e.g. `07:41:25 PM`), confirming new rows keep arriving and are prepended.
- But `document.querySelector('.alm-logpanel__body').scrollTop` stays at the same non-zero value (e.g. `400`) after re-enabling Follow — the viewport keeps showing older rows (~7 minutes stale), not the newest one. Confirmed via two separate screenshots taken seconds apart.

## Classification
UI — state toggle fires but the corresponding scroll-into-view side effect does not run.

## Evidence
Screenshots: shots/j13-22-follow-resumed.png, shots/j13-23-follow-resumed-check.png (both show `↓ Follow` active while the topmost visible row is ~07:31-07:34 despite the newest row being ~07:41).

PR #1216, opened 2026-07-20
ci: rebalance E2E shards and stop test-only changes rebuilding the app

srobroek:
## Summary

Two measured changes to the Real-UI E2E workflow. Neither changes what is
tested.

## 1. Shard balance — one word, 38% off the critical path

Ubuntu partitioned with `count`, which round-robins **within each test binary
starting at index 0** — so every binary contributes its first test to shard 1.
With 11 binaries, shard 1 has a hard floor of 11 tests regardless of how many
shards exist.

Measured on the real suite (`cargo nextest list --partition`):

| Mode | shard 1 | 2 | 3 | 4 | max |
|---|---|---|---|---|---|
| `count:N/4` (before) | **13** | 6 | 3 | 2 | 13 |
| `hash:N/4` (after) | 8 | 6 | 6 | 4 | **8** |

Windows already used `hash`; ubuntu and macOS now match.

**Corollary worth recording:** this means *splitting* large binaries would make
balance **worse**, not better — more binaries raises the shard-1 floor. The
intuitive fix is backwards, which is why the numbers are in the commit message.

## 2. Cache correctness — test-only changes no longer rebuild the app

The app binary and the nextest archive shared one key hashing every `**/*.rs`.
But `crates/e2e-tests` is **not** in `desktop_shell`'s dependency closure
(verified via `cargo metadata`: 32 workspace crates, `e2e_tests` absent), and
it has no path-dependencies on workspace crates — it drives the app externally
over WebDriver.

So editing a journey invalidated the app key and forced a full rebuild for a
**byte-identical** binary: **~340s ubuntu, ~600s Windows, on every
E2E-test-only PR**. PR #1206 — a one-line selector fix — paid exactly this.

They now have separate keys and are verified independently.

## Why the key is derived, not hand-maintained

The obvious implementation is a list of "test paths to exclude". That is the
dangerous kind of optimisation: the first time someone adds a crate the list
doesn't know about, CI serves a stale binary and the resulting green means
nothing.

Instead `scripts/e2e-app-cache-key.sh` resolves `desktop_shell`'s closure from
`cargo metadata` and hashes exactly those crates' sources. A new dependency is
picked up automatically.

**Every failure path leads to a rebuild, never a stale hit:**

- closure unresolvable → falls back to a **broader** key
- empty key → the workflow **fails the step** rather than caching under a bare
  prefix shared by every commit

## Verification

`--self-test` asserts both directions and now runs in CI, following the
existing `scripts/ci-affected-crates.sh --self-test` precedent.

| Check | Result |
|---|---|
| Edit `crates/app/core/src/lib.rs` (in closure) | key **changes** — real code change rebuilds |
| Edit `crates/e2e-tests/tests/smoke.rs` (outside) | key **unchanged** — no needless rebuild |
| Simulated `cargo metadata` failure | different, non-empty key — forces rebuild |
| Key reproducibility | stable across runs |

Both directions are checked deliberately: a key that never changes is as broken
as one that always changes, and only the first direction catches the failure
that actually matters.

Hashing is portable across all three runner images — `shasum` fallback for
macOS (no `sha256sum`), and no GNU-only `xargs -d` for Git Bash on Windows.

## Not included

Merging the 11 e2e binaries into one (which would make `count` correct by
construction) and moving the 6 IPC-only journeys out of the suite are separate,
larger changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,21 @@ concurrency:
 # Incremental artifacts are not reusable across CI runs and only bloat the cache.
 env:
   CARGO_INCREMENTAL: 0
+  # Debuginfo is 53% of every compiled artifact and the runners have been
+  # exhausting disk on it: `target/debug` reaches 1.1 GB, the 11 e2e journey
+  # binaries total 356 MB (~32 MB each, 15.3 MB after `strip --strip-debug`),
+  # and that archive is then downloaded by six shard jobs. PR #1216's ubuntu
+  # leg died with `rustc-LLVM ERROR: IO failure on output stream: No space left
+  # on device`, which reads as a code failure and costs real triage time.
+  #
+  # `line-tables-only` keeps file/line information in panic backtraces -- the
+  # part that actually matters when a test or journey fails -- while dropping
+  # the type and variable debuginfo that dominates the size. A full strip would
+  # be smaller still but loses backtrace line numbers, a bad trade for a test
+  # suite. Set here rather than as a `[profile.*]` in Cargo.toml so local
+  # developer builds keep full debuginfo. See issue #1250.
+  CARGO_PROFILE_DEV_DEBUG: line-tables-only
+  CARGO_PROFILE_TEST_DEBUG: line-tables-only
 
 jobs:
   # Cheap change-detection gate (dorny/paths-filter). Downstream jobs read its
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -273,14 +273,53 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Restore app + test archive by content key
+      # The app binary and the nextest archive have DIFFERENT inputs, so they
+      # get different keys. Keying both on every *.rs (as before) meant editing
+      # a journey under crates/e2e-tests — which is NOT in desktop_shell's
+      # dependency closure — forced a full app rebuild for a byte-identical
+      # binary: ~340s on ubuntu, ~600s on windows, every E2E-test-only PR.
+      #
+      # The app key is derived from `cargo metadata` (desktop_shell's resolved
+      # closure) rather than a hand-maintained exclude list, so a newly added
+      # dependency is picked up automatically. If the closure cannot be
+      # resolved the script falls back to a BROADER key, i.e. it fails towards
+      # rebuilding — a stale binary would produce a meaningless green.
+      # `scripts/e2e-app-cache-key.sh --self-test` asserts both directions.
+      # Guards the closure logic itself: asserts the key CHANGES for a source
+      # file inside desktop_shell's closure and does NOT change for a crate
+      # outside it. Without this the key could silently narrow and start
+      # serving a stale binary. Ubuntu only — the logic is OS-independent.
+      - name: Cache-key self-test
+        shell: bash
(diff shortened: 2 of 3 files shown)


srobroek:
CI triage — head SHA `47a62922ea73d130d9fddb0d1e3ded935747aed7`

**Failing job:** `Integration (Layer 1) — ubuntu-latest` (job 88302819782), failing step `Frontend unit tests`.

**Failure:** `src/app/LogPanel.followScroll.test.tsx > LogPanel follow-tail scroll pause/resume (T011) > re-enabling Follow resumes at the newest row even after a stale scroll-pause (#832)`. Test Files: 1 failed | 199 passed.

**Verdict: not this PR's fault.** This PR changes E2E shard balance and build skipping (3 files, no LogPanel or app code), so it cannot affect that unit test. It is the long-standing `LogPanel.followScroll` flake — `followLogs` initialises `false` and is hydrated asynchronously from `settings.get("advanced")`, so on a loaded runner the toggle click can land pre-hydration.

**Already fixed elsewhere:** PR #1251 (`fix/1249-logpanel-follow-race`) adds the missing `waitFor` gate on the hydrated follow state in exactly this test. #1251 is currently green on every job. Recommend merging #1251 first, then re-running this PR — no change needed here.

Contributing factor: the Actions queue is currently saturated (86 queued vs 12 in-progress runs), which is precisely the load condition that surfaces this race.

srobroek:
**This PR's red is not its own — it's blocked on #1251, not on anything in this diff.**

`Integration (Layer 1) — ubuntu-latest` failed in `LogPanel.followScroll.test.tsx:244`:

```
expected '↓ Follow' to be '— Follow'
  > re-enabling Follow resumes at the newest row even after a stale scroll-pause (#832)
```

That's the long-standing follow-toggle race (#1249): `followLogs` initialises to `false` and is hydrated asynchronously from `settings.get('advanced')`, so on a loaded runner the click can land pre-hydration. #1251 adds a `waitFor` gate on the hydrated state at exactly this line and is the fix.

Nothing in this PR touches `LogPanel*`; its only frontend-facing change is CI job scoping. Suggested order: **merge #1251 first, then re-run this job** — no change needed here.

Worth noting the load angle: this branch also carries the shard-rebalance and `line-tables-only` changes, which alter runner timing, so it is more likely than average to expose a timing-sensitive test. That makes it a symptom of the race rather than a cause.

srobroek:
Correction to my earlier triage — a re-run alone will **not** clear this.

#1251 merged to main at 09:07 as `22330de4` ("test: gate the LogPanel follow toggle on hydrated state"), which is the fix for the `LogPanel.followScroll` failure reported above.

However that fix is not in this branch: `git merge-base --is-ancestor 22330de4 <head>` is false, and this branch is **21 commits behind** main. The failing file here still carries the un-gated version of the test, so re-running CI against the current head would reproduce the same failure.

**This needs a rebase/merge from main, not a re-run.** `git merge-tree origin/main <head>` reports no conflict, so it should be a clean pull-forward — including against `247a047a` (#1212), which touched `.github/workflows/ci.yml`, the same file this PR edits.

Still no change needed to this PR's own content — the diagnosis that the failure is not this PR's fault stands.

srobroek:
Update: **#1251 is now merged**, so the blocker described above is cleared. This PR needs a re-run of `Integration (Layer 1) — ubuntu-latest` (or a rebase onto main to pick the fix up); no change to its own diff is required.

Issue #1208, opened 2026-07-20
Inbox Real-UI E2E is red on main: inbox_ui_unsplit_unclassified_folder_badge_is_not_classified fails deterministically


srobroek:
Three `inbox_ui_*` Real-UI E2E tests fail on every branch cut from current `main`. This is a live regression on `main`, and `main`'s own CI structurally cannot see it.

## Failing tests

Shards **ubuntu-latest 2/4** and **windows-latest 2/2**:

- `inbox_ui_unsplit_unclassified_folder_badge_is_not_classified`
- `inbox_ui_confirm_does_not_move_then_apply_moves_to_shown_destination`
- `inbox_ui_missing_path_attribute_banner_blocks_confirm` (Windows)

Representative error:

```
Error: the post-classify Inbox list never settled within 20s
       (no materialized 'needs review' sub-item row appeared); Type cells were []
Error: invalidating the inbox list query while waiting for classify failed
```

## Reproduced on two unrelated branches

Not a flake, and not caused by either PR:

| PR | Diff | Result |
|---|---|---|
| #1183 | a new `tsconfig.e2e.json` + a one-line npm script — **no runtime code** | same 2 shards fail |
| #1196 | a `Lock` a11y change in the Projects cleanup table | **same 2 shards, same tests** |

#1183's shards were re-run: **failed again**, identically. Deterministic.

## Root cause: three PRs merged 72 seconds apart, never co-tested

| PR | Merged | Subject |
|---|---|---|
| #1099 | 03:22:26 | inbox list no longer labels an unsplit folder Classified when it has no classification |
| #1135 | 03:22:39 | destination free-space estimate at plan review |
| #1131 | 03:23:38 | the mandatory-attribute rule has one definition instead of two |

**Every one of these passed all six E2E shards on its own branch.** But #1131's head (`0c243b6e`) does not contain #1099's merge commit (`ef90b074`) — verified with `git merge-base --is-ancestor`. So #1131 was tested against a base *without* #1099, and vice versa. The two inbox changes were never co-tested, and their combination breaks these tests.

Note both PRs touch the same area: #1099 changes classification labelling of unsplit folders; #1131 unifies the mandatory-attribute rule. A semantic conflict between them is entirely plausible even though each is individually correct.

## Why main didn't catch it

`main` never runs the Real-UI E2E shards. Checked three recent main runs:

```
run 29715065065 (15edea17): 0 E2E shard jobs
[...]

PR #1048, opened 2026-07-18
feat: in-app onboarding coach — orientation walk, getting-started checklists, and automatic progress

srobroek:
## Summary

Adds an onboarding coach that connects guidance to work completed in the application. A first-run orientation and per-page checklists show the next useful action without blocking the workflow.

## What's new

- Replaces the first-project guided tour with a first-run orientation walk.
- Adds per-page Getting Started checklists with persisted progress.
- Derives progress from inventory confirmation, project creation, and tool launches.
- Adds a spotlight action that navigates to and highlights the control for a checklist step.
- Adds controls to hide and restore the checklist.

## Test plan

- `cargo fmt --all -- --check`
- `cargo nextest run -p persistence_db` — 351 passed, 1 skipped
- `cargo nextest run -p persistence_db --test migration_versions_unique --test migration_release_upgrade` — 3 passed
- `cargo nextest list -p e2e_tests --test onboarding_journey --run-ignored all` — the real-UI onboarding journey compiles and is discovered
- PR CI runs the ignored real-UI onboarding journey with the desktop app and WebDriver stack.

The persistence tests lock migration filenames through `0079` and exercise an upgrade from the v0.6.0 migration set. The real-UI fixture includes exposure metadata and completes scan classification before confirming inventory.

## Spec Context

- Spec: 056 — onboarding redesign
- Supersedes spec 010 — guided first-project flow
- Migrations: `0080_onboarding.sql` and `0081_drop_guided_flow_state.sql`
- Closes #1017
- Closes #1145

Developers who ran an earlier revision of this branch may have a local database with a migration checksum from an obsolete branch-only migration number. Delete that development database and let the application recreate it, or point `ALM_DB_URL` to another database.

PR #1229, opened 2026-07-20
fix: translate the Targets detail-pane label

srobroek:
## Summary

`TargetsPage` passes a hardcoded `detailLabel="Target details"`. It is user-facing — `ListPageLayout` uses `detailLabel` to name the detail region — and **every other page already passes a translated string**: `m.projects_detail_label()`, `m.inbox_detection_details()`, `m.calibration_master_details_label()`, `m.common_details()`. Targets was the only holdout, so this follows the existing pattern rather than inventing one.

## Rebased onto #1237 — this PR is now only the i18n delta

This originally also carried the biome reformat. #1237 was opened 90 seconds later doing the same job, and its three reformatted files are **byte-identical to mine** (verified by sha256 on each). Rather than have two PRs fight over the same hunks, this is now stacked on `fix/1048-lint-and-baseline` and reduced to the part #1237 does not have.

**Diff is 3 files:** the message key, its use, and one baseline line.

## On the baseline entry

The violation is pre-existing on `main` and grandfathered in `scripts/eslint-alm-baseline.txt`. The 056 branch shifted the file down 9 lines, un-matching the **line-keyed** entry and re-raising it. #1237 re-points the entry (`562` → `571`); this fixes the string instead and drains the entry.

That is what the baseline asks for. Its header says *"New alm/* violations must be fixed for real … do not add them to the baseline"*, and the script documents the line-keying as deliberate so a shift *"forces a human to re-look at the violation rather than let it silently ride along."* This re-looks; the violation is real; so it's fixed and the entry removed.

If you'd rather keep #1237's bump and drop this, that's a defensible call too — the violation is pre-existing and keeping that PR purely mechanical has its own merit. Closing this costs nothing.

## Verification

Run with deps installed, so the gates actually executed:

- Full chain `biome check . && check-eslint-baseline && check-tokens && check-i18n-catalog` → **exit 0**, with `eslint: OK (478 files linted; 2 baselined alm/* violation(s) grandfathered)` — **down from 3**, so the entry is genuinely drained rather than moved.
- `typecheck` clean; **402 targets tests pass across 28 files**.

## Unrelated

The `Real-UI E2E` shard failures inherited here are #1208. The `0075` migration collision between #1048 and #1194 is still open and deliberately untouched — which side renumbers depends on merge order.


srobroek:
Duplicate-work notice: #1237 was opened 90 seconds after this one, against the same base, doing the same job — and is broader (Cargo.lock, CSS/tokens, `tests/e2e`, project files).

The three biome-format files here are also fixed there, so they will conflict. The substantive difference is `scripts/eslint-alm-baseline.txt`: #1237 bumps the `TargetsPage` entry `562 → 571` (re-grandfathering the violation), whereas this PR fixes the string and drains the entry.

Proposed resolution, offered on #1237: let #1237 carry the formatting and the rest, and trim this PR to just the i18n fix + baseline drain so it stacks cleanly. Holding off on rebasing until that lane responds, rather than churning both.

srobroek:
Trimmed as proposed: rebased onto `fix/1048-lint-and-baseline` (#1237) and reduced to the i18n delta only — 3 files. #1237 is untouched.

Verified its three reformatted files are byte-identical to the ones I had dropped (sha256 per file), so removing them loses nothing.

srobroek:
Closing as superseded by #1258 (`fix: several labels stayed in English when the app language changed`).

After retargeting this PR onto `main` and resolving, `git diff --stat origin/main...HEAD` is **empty** — this branch's entire unique contribution already exists on main:

- `apps/desktop/messages/en.json` has `targets_detail_label`
- `apps/desktop/src/features/targets/TargetsPage.tsx:575` uses `m.targets_detail_label()`
- `scripts/eslint-alm-baseline.txt` is fully drained (zero entries)

That is byte-for-byte the end state this PR was aiming at. #1258 fixed the identical violation independently.

This PR's own description already anticipated exactly this outcome for its overlap with #1237 — same situation, one level further along. Nothing is lost by closing. Reopen if I've misread it.

Issue #548, opened 2026-07-11
Inbox/scan: XISF calibration masters never flagged as masters (is_master is FITS-only)


srobroek:
**Surface:** Inbox scan / master detection (spec 040) · **Type:** bug

At scan (`inbox_scan_folder` → master detection) `.xisf` master files are **never flagged `is_master`**. They fall into the mixed-folder placeholder as `PENDING` while their FITS siblings are correctly extracted as individual masters.

**Evidence (DetectionMatrix diff):** 37 expected masters, **27 detected, 10 missed** — **9 of 10 are `.xisf`**, and the miss spans *every* signal: Siril `STACKCNT`/`NCOMBINE`, PixInsight `IMAGETYP='Master X'`, and name/path fallback (filenames literally containing `master`). **Zero `.xisf` files were detected as masters.** FITS siblings with identical signals detect fine. The frame *type* is still assigned to the xisf (the folder breakdown reads "4 of each"), so the gap is specifically **`is_master` for XISF**.

Missed xisf: `masterBias_ncombine`, `masterDark_stackcnt`, `masterDarkFlat_stackcnt`, `masterFlat_lc`, `masterLight_ncombine`, + path-fallback `masterBias_B`, `masterDark_A`, `masterDarkFlat_A`, `masterLight_B`.

**Expected:** XISF masters flagged `is_master` via the same signals as FITS.
**Related:** test-matrix coverage in #514; user-visible symptom in the mixed-placeholder count bug (filed alongside).

---
_Found during the Windows real-app validation campaign (Journey 8 — Calibration masters, epic #525), driving the real Tauri app via the MCP bridge. Fixtures: `docs/development/fixtures/gen_detection_matrix.py` → `D:\astrophotography\ALM test\DetectionMatrix`. Backend commit `8097d9c6`._

srobroek:
[...]

PR #1238, opened 2026-07-20
fix: PR hook reads the branch being merged, not whatever the main checkout is on

srobroek:
## Summary

The `gh pr create` hook stamped PRs with a branch and spec that had nothing to do with them. Every PR opened today got `Spec: 105 / Branch: fix/1050-wizard-site-skip-nag` regardless of its actual branch. Two independent bugs.

### 1. Wrong branch

```sh
BRANCH=$(git branch --show-current)
```

This runs in the hook's cwd — the session checkout — but `gh pr create` is routinely run from a *linked worktree* on a different branch. Worse, when two agents share the main checkout, whatever branch it happens to be sitting on leaks into unrelated PRs.

Now resolved in order of trust: explicit `--head` on the command → the branch of the checkout the command actually runs in (`.cwd` from the hook payload, as three other hooks already do) → session cwd as last resort.

### 2. Wrong spec id

```sh
SPEC_ID=$(echo "$BRANCH" | grep -oE '[0-9]{3}' | head -1)
```

Unanchored, so on `fix/1050-wizard-site-skip-nag` it matched the first three digits of a four-digit **issue** number and produced spec `105`.

Now requires a whole 3-digit path segment followed by `-`, plus a final guard that the spec directory actually exists — so it fails closed on anything the regex did not anticipate rather than inventing a spec.

## Verification

Ran the hook against each real branch shape:

| Branch | Before | After |
|---|---|---|
| `fix/1050-wizard-site-skip-nag` | `Spec: 105` ❌ | *(no spec)* ✓ |
| `spec/058-inbox-drop-parent-items` | `Spec: 105` ❌ | `Spec: 058` ✓ |
| `023-lifecycle-testing` | `Spec: 105` ❌ | `Spec: 023` ✓ |
| `ci/vite-cache` | `Spec: 105` ❌ | *(no spec)* ✓ |
| `999-not-a-real-spec` | `Spec: 999` ❌ | *(no spec)* ✓ |

`shellcheck` clean.

Also swapped `[ -d "specs/${SPEC_ID}"-* ]` for `compgen -G`: the former passes every glob match as a separate argument to `[`, so two matching spec dirs would make it error and silently discard a valid spec id.

## Why it matters

A hook that confidently asserts the wrong spec is worse than no hook — it invites agents and humans to paste a false `## Spec Context` into a PR body, and to add `Fixes #N` lines for issues belonging to an unrelated spec.

### DIFF
--- a/.claude/hooks/speckit/scripts/speckit-pr-issue-refs.sh
+++ b/.claude/hooks/speckit/scripts/speckit-pr-issue-refs.sh
@@ -47,8 +47,20 @@ Spec context goes at the bottom under "## Spec Context" (not in the title).
 The release-please draft PR will be manually curated before publish -- raw material matters.
 GUIDANCE
 
-# Detect current branch
-BRANCH=$(git branch --show-current 2>/dev/null)
+# Detect the branch the PR is actually being opened from.
+#
+# `git branch --show-current` alone is wrong: this hook runs with the session's
+# cwd, but `gh pr create` is routinely run from a linked worktree on a different
+# branch, and the main checkout may meanwhile be on something else entirely.
+# That mismatch stamped PRs with a completely unrelated branch and spec.
+#
+# Order of trust: the explicit `--head` on the command, then the branch of the
+# checkout the command actually runs in, then the session cwd as a last resort.
+HOOK_CWD=$(echo "$INPUT" | jq -r '.cwd // empty' 2>/dev/null)
+BRANCH=$(echo "$COMMAND" | grep -oE -- '--head[ =]+[^ ]+' | head -1 | sed -E 's/--head[ =]+//')
+if [ -z "$BRANCH" ]; then
+  BRANCH=$(git -C "${HOOK_CWD:-.}" branch --show-current 2>/dev/null)
+fi
 if [ -z "$BRANCH" ]; then
   jq -n --arg ctx "$CONTEXT" '{
     hookSpecificOutput: {
@@ -59,8 +71,24 @@ if [ -z "$BRANCH" ]; then
   exit 0
 fi
 
-# Extract spec number from branch name (e.g., 023-lifecycle-testing -> 023)
-SPEC_ID=$(echo "$BRANCH" | grep -oE '[0-9]{3}' | head -1)
+# Extract spec number from branch name (e.g., 023-lifecycle-testing -> 023).
+#
+# The number must be a whole 3-digit path segment followed by `-`, so a branch
+# named after a 4-digit ISSUE (`fix/1050-wizard-site-skip-nag`) is not read as
+# spec `105` — an unanchored `[0-9]{3}` matched the first three digits of the
+# issue number and stamped every such PR with a spec it has nothing to do with.
+SPEC_ID=$(echo "$BRANCH" | grep -oE '(^|/)[0-9]{3}-' | head -1 | tr -dc '0-9')
+
+# Final guard: only treat this as a spec branch if that spec actually exists.
+# Cheap, and it fails closed on any pattern the regex above did not anticipate.
+if [ -n "$SPEC_ID" ]; then
+  # compgen, not `[ -d glob ]`: the latter passes every match as a separate
+  # argument to `[`, so two matching spec dirs would make it error out and
+  # silently discard a perfectly valid spec id.
+  if ! compgen -G "${HOOK_CWD:-.}/specs/${SPEC_ID}-*" >/dev/null 2>&1; then
+    SPEC_ID=""
+  fi
+fi
 
 if [ -z "$SPEC_ID" ]; then
   jq -n --arg ctx "$CONTEXT" '{


srobroek:
**Heads-up on layering: this patches a generated file, and the upstream fix has now merged.**

`.claude/hooks/speckit/scripts/speckit-pr-issue-refs.sh` is installed from `srobroek/agentic-packages/packages/speckit#main` (`apm.yml:11`), and on `main` it is byte-identical to that source. So this edit is reverted by the next `apm install` — silently, and at a moment unrelated to whoever runs it.

The same three defects are now fixed at the source in srobroek/agentic-packages#548 (merged), which also covers a third instance this PR doesn't: `git remote get-url origin` was read from the hook's cwd too, so a lane working across two repos queried the *other* repository's issues. It additionally adds a fail-closed guard requiring `specs/<id>-*` to exist, and nine bats cases verified two-directionally.

The `--head` trust order and the `compgen`-over-`[ -d glob ]` reasoning in this PR were both adopted upstream — they were the better call and are credited in the commit.

Since propagation here is release-gated, this PR is still useful as an **interim** patch until an agentic-packages release lands and `apm install` runs. Two sensible options:
- keep it, with a note that it is temporary and superseded by #548; or
- close it and pick the fix up via the release.

Flagging rather than deciding — it's not my PR to close, and either way it should not merge as a permanent fix.

srobroek:
**Correction to my comment above — I overstated the revert risk, and I missed a second copy.**

I said "this is reverted by the next `apm install`". That is wrong as written. Tested in a throwaway worktree against `origin/main`:

- **Plain `apm install` preserves it.** It reports `50 files skipped -- local files exist, not managed by APM` and leaves the working tree untouched apart from `apm.lock.yaml`. A local edit survives.
- **`apm install --force` overwrites it.** The file *is* APM-managed — `apm.lock.yaml:380` lists `.claude/hooks/speckit/scripts/speckit-pr-issue-refs.sh` under the speckit package's `deployed_files`.

So the divergence is real, but it needs `--force` to bite, not a routine install. Sorry for the sharper claim.

**The thing I did miss, which cuts the other way:** there are two deployed copies, and this PR only patches one.

```
.claude/hooks/speckit/scripts/speckit-pr-issue-refs.sh   <- patched here
.codex/hooks/speckit/scripts/speckit-pr-issue-refs.sh    <- still the buggy version
```

Both are listed in `deployed_files`. Under Codex the hook still misreads the branch and still turns `fix/1050-*` into spec 105.

**Also worth knowing before anyone relies on `apm install` to deliver the upstream fix:** it currently can't. APM's cached bare clone of agentic-packages sits at `57999765`, many commits behind, and it does not re-fetch on install — a manual `git fetch` in the cache fails with exit 128 (HTTPS auth against a private repo). So `apm_modules/.../speckit/scripts/speckit-pr-issue-refs.sh` in this repo is *also* still the old script even after `--force`. Until that cache refreshes, srobroek/agentic-packages#548 (released as **speckit v5.1.1**) cannot reach this repo at all.

That strengthens the case for keeping this PR as the interim fix — ideally extended to the `.codex` copy too.

srobroek:
Closing in favour of #1267, which syncs the full upstream file (srobroek/agentic-packages#548) including the third defect this PR missed — `git remote get-url origin` read from the hook's cwd, so a lane working across two repos queried the other repository's issues.

One correction to the review note, since it changes the operational picture: I ran `apm install`, and it does **not** revert this file. The hook is reported as "local files exist, not managed by APM" (one of 50 skipped), and the cached module is pinned to `resolved_commit bb7a3781`, which predates #548. So plain `apm install` neither overwrites the local edit nor delivers the upstream fix — propagation here needs a deliberate sync, not a release. #1267 does that sync and verifies it four ways, in both directions.

The `--head` trust order and the `compgen`-over-`[ -d glob ]` reasoning from this PR both survive upstream, so the work wasn't wasted — it just lands via #1267 instead.

srobroek:
**Retracting my correction — my *first* comment was right. A plain `apm install` does revert this, and I've now reproduced it against merged main.**

I'm sorry for the flip-flop. Here's why the middle comment was wrong: when I tested it, the on-disk file was still byte-identical to the packaged version, so the overwrite was real but produced no visible diff, and I read that as "skipped". Now that this PR is merged, the same test shows the revert plainly.

Reproduced on a clean worktree at `origin/main` (`47c43163`), running nothing but `apm install`:

```
 M .claude/hooks/speckit/scripts/speckit-pr-issue-refs.sh

-HOOK_CWD=$(echo "$INPUT" | jq -r '.cwd // empty' 2>/dev/null)
-BRANCH=$(echo "$COMMAND" | grep -oE -- '--head[ =]+[^ ]+' | ...
+# Detect current branch
+BRANCH=$(git branch --show-current 2>/dev/null)
```

No `--force`. The whole fix — branch resolution, the anchored spec regex, the `compgen` guard — is silently replaced by the buggy version. The `50 files skipped -- local files exist` line does not cover this file: it's APM-managed, listed at `apm.lock.yaml:380` under the speckit package's `deployed_files`.

**So this is a live hazard on main right now.** Any agent or human who runs `apm install` reverts the fix, with no warning and no failing check.

## Why it reverts to the *old* version rather than the new upstream one

`apm.lock.yaml` pins `resolved_commit: bb7a3781` (2026-07-17) for all 66 agentic-packages deps. `#main` in `apm.yml` does **not** mean "track main" at install time — the lock wins. I refreshed APM's cached clone (its `refs/heads/main` mirror was frozen at `57999765` and it never re-fetches on install; tags through `speckit--v5.1.1` are now present), but that changes nothing on its own while the lock is pinned.

Re-resolving needs `apm update --yes`, which reports **66 updated** — it can't be scoped, since passing a single package makes APM plan 65 removals. That's a deliberate, reviewable change rather than something to slip in here.

## Suggested follow-up

Until that lock refresh lands, this file is in a state where the committed fix and the installed fix disagree. Two loose ends regardless:

- `.codex/hooks/speckit/scripts/speckit-pr-issue-refs.sh` is also in `deployed_files` and still carries the original bug, so the Codex runtime is unfixed.
- Once the lock is refreshed, the deployed copy becomes srobroek/agentic-packages#548 (released as **speckit v5.1.1**), which is a superset of this patch — it also fixes `git remote get-url origin` reading from the wrong cwd, and carries nine bats cases. At that point this local patch should be dropped rather than maintained in parallel.

PR #1245, opened 2026-07-20
fix: real-UI end-to-end run stops instead of testing another build

### LINKED ISSUE #1244: Real-UI E2E silently tests another worktree's build when :5173 is taken
## Problem

`apps/desktop/scripts/run-e2e-real.sh` serves the frontend with
`vite preview --port 5173` and **no `--strictPort`**. When 5173 is already
bound, vite silently falls back to another port — but the app binary still
loads `http://localhost:5173`, because `devUrl` in `tauri.conf.json` is baked
in at build time.

The suite then exercises **whatever build currently holds 5173**, which on this
host is routinely a concurrent worktree's interactive dev server
(`vite.config.ts` hardcodes 5173 for `pnpm dev`).

## Impact

Phantom failures in journeys unrelated to the branch under test — worse than
not running at all, because the output looks like a real result.

Observed 2026-07-20 while verifying #1107: a local run silently exercised
another worktree's build and reported `.pv-calib-table` never appearing, an
error with no connection to the branch. It cost a diagnosis cycle before the
port was checked.

With ~90 concurrent worktrees on this host it is a standing trap, not a one-off.

## Why the port cannot simply be moved

The mock-mode Playwright suite solved this by moving to a dedicated port
(#1117). That option is not available here:

- the app loads its frontend from the compiled-in `devUrl`, and
- `crates/e2e-tests/tests/common/mod.rs::APP_URL` must stay **byte-identical**
  to it, because `localhost` and `127.0.0.1` are different web origins with
  separate localStorage — a mismatch ping-pongs the setup gate indefinitely
  (documented at that const).

So the runner must use 5173, and the correct response to it being busy is to
stop with a clear message rather than produce a misleading run.

## Fix

Preflight that names the holding process's directory, plus `--strictPort` to
close the check-then-bind race.

srobroek:
Closes #1244.

## Summary

- `run-e2e-real.sh` served the frontend with `vite preview --port 5173` and no `--strictPort`. With 5173 occupied, vite silently bound elsewhere while the app still loaded `localhost:5173` — so the suite tested whichever build held that port, usually another worktree's dev server.
- That produced phantom failures in unrelated journeys. Observed 2026-07-20: a local run silently exercised a different worktree's build and reported errors unrelated to the branch.
- Adds a preflight that names the holding process's directory, plus `--strictPort` to close the check-then-bind race.

## Why not just move the port

The mock-mode Playwright suite did exactly that (#1117), but it is not an option here: the app loads its frontend from `devUrl` in `tauri.conf.json`, compiled in at build time, and `common/mod.rs::APP_URL` must stay byte-identical to it — `localhost` and `127.0.0.1` are separate web origins with separate localStorage, and a mismatch ping-pongs the setup gate forever. So 5173 is fixed, and stopping is the only safe response to it being busy.

## Verification

Both directions, against the real script:

| Condition | Result |
|---|---|
| 5173 held | exits 1 in <1s **before any build**, printing the holder's path |
| 5173 free | proceeds unchanged |

`bash -n` and `shellcheck` clean.

## Spec Context
- Spec: 037 (e2e/integration testing)
- Branch: fix/e2e-port-collision-guard

### DIFF
--- a/apps/desktop/scripts/run-e2e-real.sh
+++ b/apps/desktop/scripts/run-e2e-real.sh
@@ -15,9 +15,46 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 desktop_dir="$(dirname "$script_dir")"
 repo_root="$(cd "$desktop_dir/../.." && pwd)"
 
+# Port 5173 is NOT negotiable here, unlike the mock-mode Playwright suite which
+# moved to a dedicated port (playwright.config.ts, #1117). The app binary loads
+# its frontend from `devUrl` in src-tauri/tauri.conf.json, which is baked in at
+# BUILD time — and `common/mod.rs::APP_URL` must stay byte-identical to it,
+# because `localhost` and `127.0.0.1` are different web origins with separate
+# localStorage (a mismatch ping-pongs the setup gate forever). So this runner
+# must serve on exactly :5173, and the only safe response to that port being
+# taken is to stop.
+#
+# Fail LOUDLY when it is occupied. Previously this ran `vite preview --port
+# 5173` with no --strictPort: vite silently fell back to another port, the app
+# still loaded localhost:5173, and the suite tested WHOEVER held that port —
+# typically a concurrent worktree's interactive dev server (vite.config.ts
+# hardcodes 5173 for `pnpm dev`). That produces phantom failures in unrelated
+# journeys, which is worse than not running at all. Observed 2026-07-20: a
+# local run silently exercised another worktree's build and reported errors
+# that had nothing to do with the branch under test.
+e2e_port=5173
+if (command -v ss >/dev/null 2>&1 && ss -ltn 2>/dev/null | grep -q ":${e2e_port} ") ||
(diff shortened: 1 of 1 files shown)


srobroek:
**CI on this PR is vacuously green — read the checks with that in mind.**

Every substantive job is `SKIPPED`: the path filters resolve `rust=false` /
`run=false`, because the only changed file is
`apps/desktop/scripts/run-e2e-real.sh`, which no filter covers. The sole jobs
that executed are the two `Detect changes` steps.

That is not a gap to fix here: the script is a **local** runner
(`test:e2e:real` / `just test-e2e`). CI runs the real-UI suite through
`.github/workflows/e2e.yml` directly and never invokes this script, so there is
genuinely nothing for CI to exercise.

The evidence for this change is therefore manual, run against the real script
rather than a stub:

| Condition | Result |
|---|---|
| 5173 held by another process | exits **1** in under a second, **before any build**, naming the holder's directory |
| 5173 free | proceeds unchanged into build + preview |

Plus `bash -n` and `shellcheck` clean.

Flagging explicitly because a green tick here means "nothing ran", not "this
works" — and that distinction has bitten this repo before.

PR #1253, opened 2026-07-20
feat: target identity stays visible when the table scrolls sideways

### LINKED ISSUE #1158: Spec 054: three untracked gaps (pinned Targets columns, typed dock persistence, useAdaptiveDock tests)
## Context

While reconciling the spec-054 record (#1119, closing #1068), three gaps surfaced that are **real but untracked** — the spec now records them honestly as not-built, but nothing points at them from the issue tracker, so they will quietly rot.

Filing so they are visible. None is urgent; each needs a decision on whether it is wanted at all.

## 1. T024 / T025 — pinned Targets columns, never built

The spec designed a permanent importance-based column order for the Targets table, with the favourite-star and designation columns staying pinned while the rest scroll horizontally.

Reality: `TargetsTable.tsx` has **zero** `position: sticky` column rules. Past its min-width floor the whole table scrolls as one unit. The journey docs claimed this shipped; corrected in #1119.

**Decision needed:** build it, or formally drop it from the spec? Right now it sits in an ambiguous "designed, not built, nobody assigned" state.

## 2. T007 — `AppPreferences.detailDock`, never built

Dock placement and width persist in raw `localStorage` keys (`alm-dock-placement-<dockId>`, `alm-dock-width-<dockId>`), not through the typed `AppPreferences` layer the spec designed.

This works today. #1106 fixed the sharpest edge — the keys used to be derived from a *translated UI string*, so changing language silently orphaned every saved placement.

**Decision needed:** is raw localStorage acceptable as the permanent answer, or should dock state move into typed preferences alongside everything else?

## 3. `useAdaptiveDock` has no dedicated unit tests

The hook driving all adaptive placement is covered only indirectly, via `ListPageLayout.test.tsx`. There is no test file for the hook itself.

Its behaviour is non-trivial and worth pinning directly: threshold resolution, `sideAvailable` fallback when the window is too narrow for a side dock, width clamping to `[minWidth, window * maxFraction]`, `setOverride(null)` clearing persistence, and `dockId`-scoped isolation between pages.

Note the null-override path is exactly what #1066 was filed for — it had no call site at all, and no test caught that.

---

All three are recorded in `specs/054-adaptive-detail-dock/` as of #1119. This issue exists so they are reachable from the tracker rather than only from the spec.

srobroek:
Closes #1158.

## Summary

- The Targets table's star + designation columns now stay pinned while the rest scroll horizontally, so you never lose track of which target a row is.
- Spec 054's last outstanding item (T024/T025, FR-006/FR-007), designed but never built.

## The problem is the default, not an edge case

Measured live, not estimated:

| | |
|---|---|
| Side dock engages at | 1400px window |
| Sidebar | 220px |
| Dock | 420px (default) |
| Table space | **760px** |
| Table min-width | **1000px** |
| → horizontal overflow | **240px** |

With `table-layout: fixed`, the 20%-wide designation column is exactly what scrolls out of view — so you compare gain/exposure/filter values with no idea which target each row belongs to.

## CSS-only

The colgroup classes and a sticky `thead` already existed; this adds the horizontal axis. **No TypeScript touched**, which also kept it clear of the two open PRs currently editing `TargetsTable.tsx` (#1160, #1048) — I confirmed neither touches the colgroup, the column classes, or these CSS files.

Safe here specifically because `TargetsTable` windows rows with the padding-spacer pattern and applies **no transforms** — a transformed ancestor would create a containing block and silently break `position: sticky`. Cost is bounded by the windowing: ~13 rows in the DOM means ~26 sticky cells.

## Three things measurement caught that reasoning did not

1. **`45px` produced a 50px column.** The other columns are percentages totalling 95%, which fixed-layout honours exactly, so the star column renders as whatever they leave. The offset was 5px short and the designation drifted `270px → 265px` while scrolling.
2. **A percentage offset is worse.** Sticky offsets resolve against the *containing block*, so `left: 5%` became 5% of the 760px scroll container = 38px — a 12px drift.
3. **50px is exact**, because the offset is only consulted while scrolling, and the table only scrolls at its 1000px floor: width is `max(container, 1000px)`, so overflow implies the table is exactly 1000px and the star exactly 50px. Above the floor nothing scrolls and the offset is never used.

Every one of those would have shipped as a subtly misaligned column under static review.

## Two hazards handled

- **Spanning rows.** Scoped to `.pv-targets-table__row`, so group headers (`.pv-listgroup`) and the virtualizer's spacer rows — which render a single `<td colSpan>` — are structurally excluded. An unscoped `nth-child` would pin a full-width cell and drag it across the viewport.
- **Collapsed borders.** The seam uses `box-shadow`, not `border-right`: `.pv-table` is `border-collapse: collapse`, where borders belong to the table and detach from sticky cells while scrolling.

## Verification

At 1400×900 with a 420px side dock, scrolled fully right:

| Check | Result |
|---|---|
| star drift | **0px** |
| designation drift | **0px** |
| designation *header* drift | **0px** |
| third column | scrolled **240px** |
| header aligned with body | ✅ |
| pinned cells opaque | ✅ (no bleed-through) |
| hover / selected carry across pinned cells | ✅ |
| z-order (header 3 > body 2) | ✅ |

Plus `check-tokens.sh` clean (incl. WCAG AA contrast across 6 themes) and 28 files / 402 targets tests passing.

**Not verified:** the mock dataset has only 2 targets, so group-header and spacer rows never materialise and virtualization isn't exercised. Their exclusion is verified *structurally* — only data rows carry `.pv-targets-table__row` (`TargetsTable.tsx:1281`) — not observed live. T026 (a real-browser E2E assertion) is left open in the spec for that reason; a Layer-1 test cannot substitute, since jsdom has no layout engine and cannot observe sticky offsets.

## Spec Context
- Spec: 054 (adaptive detail dock), T024/T025, FR-006/FR-007
- Branch: feat/1158-pinned-targets-columns

### DIFF
--- a/apps/desktop/src/styles/components/merges-2.css
+++ b/apps/desktop/src/styles/components/merges-2.css
@@ -300,6 +300,102 @@
   z-index: 1;
 }
 
+/* === Pinned identity columns (spec 054 FR-006/FR-007, #1158) ==============
+ *
+ * The star + designation columns stay put while the rest scroll horizontally,
+ * so a row's identity is never lost. This is reachable in the DEFAULT side-dock
+ * configuration, not an edge case: the dock engages at a 1400px window, the
+ * sidebar takes 220px and the dock 420px, leaving the table ~760px against its
+ * own 1000px min-width floor — so it h-scrolls by ~240px and, with
+ * `table-layout: fixed`, the 20%-wide designation column is what scrolls away.
+ *
+ * Safe here specifically because TargetsTable windows rows with the
+ * PADDING-SPACER pattern and applies NO transforms (see its docstring): a
+ * transformed ancestor would create a containing block and silently break
+ * `position: sticky`. The axes are also orthogonal — virtualization is
+ * vertical, pinning is horizontal. Cost is bounded by the windowing: ~13 rows
+ * are in the DOM at a time, so this is ~26 sticky cells, not 22k.
+ *
+ * Scoped to `.pv-targets-table__row` on purpose. Group headers
+ * (`.pv-listgroup`) and the virtualizer's spacer rows render a single
+ * `<td colSpan>`; an unscoped `nth-child` would pin that full-width spanning
+ * cell and drag it across the viewport.
+ */
+.pv-targets-table {
+  /* 50px, and the value is exact rather than approximate — reasoning, since
+     two other candidates are wrong in ways that are invisible without measuring:
+       - The other columns are percentages totalling 95%, which fixed-layout
+         honours exactly, so the star column renders as whatever they leave.
+         Declaring 45px here produced a 50px column and left the offset below
+         5px short: the designation drifted 270px -> 265px while scrolling.
+       - A percentage is worse. Sticky offsets resolve against the CONTAINING
+         BLOCK, not the table, so `left: 5%` became 5% of the 760px scroll
+         container = 38px, drifting by 12px.
+     50px is exact because the offset is only ever consulted while scrolling,
+     and the table only scrolls at its 1000px min-width floor: its width is
+     max(container, 1000px), so h-overflow implies container < 1000px implies
+     table == 1000px exactly, making the star 5% == 50px. Above that floor the
+     table fits, nothing scrolls, and the offset is never used.
+     The width itself is applied in merges-3.css next to the star column's own
+     docs; this token inherits down to the <col> element. */
+  --pv-targets-star-w: 50px;
+}
+
+.pv-targets-table__scroll .pv-targets-table thead th:nth-child(-n + 2),
+.pv-targets-table__scroll
+  .pv-targets-table
+  .pv-targets-table__row
+  > td:nth-child(-n + 2) {
+  position: sticky;
+}
+.pv-targets-table__scroll .pv-targets-table thead th:nth-child(1),
+.pv-targets-table__scroll .pv-targets-table .pv-targets-table__row > td:nth-child(1) {
+  left: 0;
+}
+.pv-targets-table__scroll .pv-targets-table thead th:nth-child(2),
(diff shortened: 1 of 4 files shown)


srobroek:
Added three spec-reconciliation commits beyond the pinning change itself.

**#1195 delivered work that was never credited in the task list** — my own
merge, my own drift, and the same pattern #1119 had to clean up:

| Task | Was | Now |
|---|---|---|
| T002 `useAdaptiveDock` unit tests | "no tests found" | **delivered**, minus hysteresis — which the hook does not have |
| T003 `detailDock` persistence test | "field never built" | **delivered** |
| T007 typed `AppPreferences.detailDock` | "raw localStorage, no tracked issue" | **delivered** |
| T023 multi-page restore | "not confirmed" | **delivered in part**, stays open |
| T024/T025 pinned columns | "superseded, no evidence" | **delivered** by this PR |

T023 deliberately stays open: the remount restore is covered, but a real page
**reload** and the **drag** path are not, and neither shortcut is free — a
genuine reload is what exercises the preference cache.

Also corrected the summary block (still listing five of these as superseded)
and the trailing "no task remains Open" line — two now are, T023 and T026, both
regression cover for behaviour that ships rather than unbuilt product.

Two smaller corrections, both still OPEN, only their assertions fixed:
- **T030** names `.alm-listpage__detail`, retired by the design-system rename;
  the suites now select `.pv-listpage__detail`. Called out explicitly because a
  helper missed by that same rename silently broke an Inbox journey on `main`
  (fixed in #1206) — a task naming a CSS class should be read as possibly
  pre-rename.
- **T034** claimed the gates were never run. True of the original PRs, not of
  the work merged since; it now records exactly which gates ran and why that
  still does not satisfy the task as written.

Net: spec 054's open items drop from 8 to 5, and the survivors are honest —
one withdrawal, two test-coverage gaps, one N/A, one partial gate claim.

Issue #1215, opened 2026-07-20
fix: tool-launch async work outlives its test and fails the unit-test job with unhandled rejections


srobroek:
## Summary

`Integration (Layer 1) — ubuntu-latest` failed the **Frontend unit tests** step (job 88276537750) while reporting:

```
Test Files  199 passed
⎯⎯ Unhandled Errors ⎯⎯
Vitest caught 2 unhandled errors during the test run.
Unhandled Rejection: Error: window is not defined
  file=apps/desktop/src/features/projects/tool-launch.ts
```

Every test file passed. The job still went red, because vitest exits non-zero on unhandled errors. This makes the failure especially misleading: the summary line reads as a clean pass.

## What is established

- The reported source is `apps/desktop/src/features/projects/tool-launch.ts`.
- The failure is intermittent — it did not reproduce on the same commit's Windows or macOS legs.
- The `launch` callback's body is wrapped in `try/catch/finally`, so a `ReferenceError` raised *during* the test would be caught. For it to surface as an **unhandled rejection**, the async work must still be in flight after the owning test finished and jsdom tore `window` down — at which point the `catch`/`finally` paths themselves (`addToast`, the Paraglide message getters, `setState`) can throw into a promise nobody is awaiting.
- Candidate owners: `tool-launch.test.ts` and `ProjectDetail.tool-launches.test.tsx` — a test that fires a launch and asserts on an intermediate state without awaiting the launch settling would leave exactly this dangling promise.

## What is not yet established

Which specific test leaves the promise in flight, and whether the right fix is (a) awaiting settlement in that test, or (b) guarding the hook against post-unmount continuation. Both are plausible; picking between them needs the owning test identified first. Deliberately not guessed at here.

## Why it matters beyond this one job

This is the third distinct intermittent failure found in the same "Frontend unit tests" step in one sitting, alongside #1209 and #1213. Those two were async-hydration races in tests; this one is an async leak *out* of a test. Same underlying theme: unit tests interacting with async work whose lifetime they do not control.

## Related

- #1209 — `LogPanel.followScroll.test.tsx` hydration race
- #1213 — `CalibrationMatching.test.tsx` hydration race

PR #1254, opened 2026-07-20
fix: mount-time fetches no longer update state after their component unmounts

srobroek:
## Summary

- Adds unmount guards to the 11 remaining mount-time async effects that called `setState` with no cancellation check, across 10 files.
- Same defect as #1215 (fixed separately in #1217), found by sweeping for that pattern after #1215 turned up by accident.

## Why this is worth doing

In the app a late `setState` against a torn-down root is a harmless no-op. Under vitest it is not: the update lands after jsdom has removed `window`, and React's scheduler throws `ReferenceError: window is not defined` into a promise nobody awaits.

That fails the **entire** `Frontend unit tests` job while the summary still reads `Test Files 199 passed / Tests 1820 passed` — a red job whose own output says everything passed. #1215 was exactly that, and it cost real time to trace.

## The two shapes

**Direct mount effects** get the `let cancelled` + cleanup pattern already used in ~10 files across this codebase (`CommandPalette`, `Shell`, `useEntityNames`, `WizardPage`, …), so this follows existing convention rather than introducing a new one:

- `app/LogPanelContext.tsx` — settings hydration
- `features/settings/ProcessingTools.tsx` — tool profile list
- `features/settings/CalibrationMatching.tsx` — tolerances (its existing `editedRef` guards a *different* concern: protecting a user edit from a late fetch, not protecting React from an unmounted write)
- `features/settings/NamingStructure.tsx` — two separate load effects
- `features/setup/steps/StepScan.tsx` — the largest one: the scan fans out into an independent async branch per source, each able to outlive the others, so all five `setState` sites are gated

**Callback-invoked loaders** need a mounted ref instead, because the loader is also re-invoked on user actions (retry, refresh, create/update/delete) — a per-effect `cancelled` flag cannot reach into a `useCallback`:

- `features/settings/DataSources.tsx` — `loadRoots`
- `features/settings/SourceProtectionOverride.tsx` — `load`
- `features/plans/PlanProtectionGate.tsx` — `load`
- `features/settings/Equipment.tsx` — all four loaders (cameras, telescopes, trains, filters)
- `dev/ContractsPage.tsx` — `loadContracts` (behind the `dev-tools` feature, but the TS surface is unit-tested)

## Verification

- `tsc --noEmit` clean on both configs (app + e2e)
- Frontend lint pipeline clean (Biome, ESLint gates, token contrast, i18n catalog)
- Full unit suite: **200 files / 1835 tests passed**, no unhandled errors

## Scope note

This is the mechanical guard only — no behavioural change is intended in any of these components, and no test assertions were touched. The one judgement call is `CalibrationMatching`, where I left the existing `editedRef` check in place and added `cancelled` alongside it rather than conflating the two.

Sibling to #1217, which fixed the first instance of this pattern in `tool-launch.ts`.


srobroek:
## Bisect result: the unmount guards in this PR regress 3 mock-mode E2E specs

Reproduced locally (0 retries, deterministic — not the CI-load flake class):

| commit | result |
|---|---|
| `0aced5f0` (parent) | **10 passed** (27.3s) |
| `6eb0ced6` (this PR) | **3 failed, 7 passed** (3.1m) |

Failing specs:

- `tests/e2e/cleanup_review.spec.ts:61` — `locator.click` timeout; the `Acknowledge` button in the protection gate never becomes actionable
- `tests/e2e/project_lifecycle_transitions_full.spec.ts:156` — same shape (`review overlay → approve & apply`)
- `tests/e2e/typography_semantic_mono.spec.ts:118` — `rootPath` (`/astro/raw`) element not found

The first two point at `PlanProtectionGate.tsx`, the third at `DataSources.tsx` — both are files this PR guards. The likely mechanism is a guard suppressing the state update that populates the gate / the root list, rather than only suppressing a post-unmount update.

Note the runtime signal too: 27s → 3.1m. The guarded components appear to be waiting out timeouts, not just rendering differently.

Repro:

```
pnpm -C apps/desktop test:e2e cleanup_review.spec.ts \
  typography_semantic_mono.spec.ts project_lifecycle_transitions_full.spec.ts
```

Flagging because a `useMountedRef` extraction is in progress on this branch — a pure refactor of the same semantics would preserve this regression, so the guard condition itself needs to change, not just its packaging.

srobroek:
CI triage — head SHA `6eb0ced63078116719135530d01422e957b9ce17`

**Failing job:** `UI E2E (mock-mode Playwright)` (job 88304603446), failing step `Mock-mode UI tests`.

**Failure:** `tests/e2e/cleanup_review.spec.ts:61` — "scan → review candidates with confidence + protection → generate plan → protection gate → approve & apply". Times out at `cleanup_review.spec.ts:174` waiting for `getByTestId("plan-review-overlay").getByRole("button", { name: "Acknowledge" })`, which never renders. Failed on the initial run and both retries at ~30.2s each, so this is deterministic, not flake.

**Verdict: this PR's own regression**, not pre-existing on main.

**Cause.** The PR introduces two different guard patterns. The `let cancelled = false` per-effect flag (LogPanelContext, CalibrationMatching, NamingStructure, ProcessingTools, StepScan) is correct — the flag is re-initialised on every effect run.

The `const mountedRef = useRef(true)` pattern is not. In five files — `dev/ContractsPage.tsx`, `features/plans/PlanProtectionGate.tsx`, `features/settings/DataSources.tsx`, `features/settings/Equipment.tsx`, `features/settings/SourceProtectionOverride.tsx` — the effect only ever assigns `mountedRef.current = false` in its cleanup and never re-arms it to `true` in the effect body.

`apps/desktop/src/main.tsx:86` wraps the app in `<StrictMode>`, and `apps/desktop/playwright.config.ts:66` runs the Vite **dev** server, so React dev + StrictMode double-invokes effects as mount → cleanup → mount on the *same* instance. The cleanup latches `mountedRef.current = false` and nothing restores it, so the guard in `PlanProtectionGate`'s `.then()` drops the `planProtectionCheckCmd` response permanently. `loadState` never reaches `ready`, the protected-items list never renders, and there is no Acknowledge button to click.

**Scope beyond the visible failure:** only `PlanProtectionGate` had an e2e test covering this path. The same latching applies to the DataSources, Equipment and SourceProtectionOverride settings panes and the dev ContractsPage, which would silently never finish loading in dev. Note this is dev-only — production React does not double-invoke StrictMode effects — but it breaks the dev app and the mock-mode suite.

srobroek:
## Root cause confirmed, and the `useMountedRef` extraction is the right fix — verified

Following up on my bisect above. The mechanism is a **StrictMode latch**, not a logic error in the guard condition:

```js
const mountedRef = useRef(true);
useEffect(() => () => { mountedRef.current = false; }, []);
```

The effect body is empty — it only returns a cleanup. Under StrictMode's dev double-invoke (mount → cleanup → mount on the same instance) the cleanup sets the ref `false` and **nothing ever sets it back**. From that point every `.then()` bails, so `PlanProtectionGate` never leaves `loading`, the `Acknowledge` button never renders, and the click times out. `DataSources` never populates the root list, which is what the typography spec was reaching for. That also explains the 27s → 3.1m runtime: the specs were waiting out timeouts, not asserting against different markup.

All five guarded files in `6eb0ced6` share the pattern: `ContractsPage`, `PlanProtectionGate`, `DataSources`, `Equipment`, `SourceProtectionOverride`.

The `useMountedRef` hook staged on this branch already re-arms on mount, which is exactly the fix — and its doc comment names the same cause independently. To confirm it's sufficient rather than just plausible, I applied the re-arm to all five sites at `6eb0ced6` in a scratch worktree and re-ran:

| state | result |
|---|---|
| `0aced5f0` (parent, no guards) | 10 passed |
| `6eb0ced6` (guards, latching) | **3 failed**, 7 passed |
| `6eb0ced6` + re-arm on all 5 | **10 passed** |

So the extraction closes it completely — no further behavioural change needed beyond routing all five call sites through the hook.

One caveat on reproducing locally right now: host load average is ~68 with the parallel lanes running, and Playwright's `webServer` timeout is a hard 60s, so the run aborts with `Timed out waiting 60000ms from config.webServer` before any test executes. That's infrastructure, not a failure — raise it temporarily if you hit it.

srobroek:
**The three E2E failures look like a real regression in this PR, not flake — the unmount guard latches `false` permanently under StrictMode.**

96 passed / 3 failed, and the same three specs fail through all retries:

```
cleanup_review.spec.ts:61                    (spec 017)
project_lifecycle_transitions_full.spec.ts:156
typography_semantic_mono.spec.ts:118         (spec 055)
```

## Why

The guard is initialised once and only ever cleared:

```ts
const mountedRef = useRef(true);
useEffect(() => () => { mountedRef.current = false; }, []);
```

`main.tsx:86` wraps the app in `<StrictMode>`, and the Playwright job serves it with the **dev** server (`playwright.config.ts:66` → `pnpm exec vite`, not a preview of a production build). In dev, StrictMode mounts → unmounts → remounts. The first cleanup sets `mountedRef.current = false`, and **nothing sets it back to `true`** on the second mount. From then on every `if (!mountedRef.current) return` short-circuits, so `setLoadState('ready')` never runs and the component stays in `loading` forever.

That matches the failure signature exactly — `element(s) not found` and `locator.click` timing out on controls that never render.

`PlanProtectionGate` is what `cleanup_review` drives, which ties the most obvious spec directly to the change.

## Scope — five files, not one

```
apps/desktop/src/dev/ContractsPage.tsx
apps/desktop/src/features/plans/PlanProtectionGate.tsx
apps/desktop/src/features/settings/DataSources.tsx
apps/desktop/src/features/settings/Equipment.tsx
apps/desktop/src/features/settings/SourceProtectionOverride.tsx
```

All five are `useRef(true)` with zero occurrences of a reset. Fixing only the one that shows up in a failing spec would leave the other four latched.

## Fix

Set the ref in the effect **body**, not just the cleanup — this is why the idiomatic form assigns on mount:

```ts
useEffect(() => {
  mountedRef.current = true;
  return () => { mountedRef.current = false; };
}, []);
```

Worth verifying in both directions afterwards: these three specs should fail on the current head and pass with the reset added. If they still fail, the cause is something else and this diagnosis is wrong.

Separately, this PR is currently **DIRTY** (conflicting with main), so it needs a rebase regardless.

srobroek:
Independently verified `a657610b`, and it's a better fix than the one I had staged — I was mid-push with five copies of `mountedRef.current = true`; extracting `useMountedRef` and routing all five sites through it is the right call, and every touched file gets shorter.

Verified two-directionally rather than just re-reading it. Deleting only the mount-time re-arm inside the hook:

```
× re-arms after the StrictMode mount/cleanup/mount cycle
  Tests  1 failed | 1 passed
```

and with it restored, `useMountedRef.test.tsx` + `PlanProtectionGate.test.tsx` are **9 passed**. So the test genuinely pins the fix rather than passing for an unrelated reason.

The `reports false once the component really unmounts` case is worth calling out — it's what stops the naive "just always return true" repair from looking correct.

Two notes:

1. **The E2E specs are the real confirmation and haven't re-run yet.** The three that failed — `cleanup_review`, `project_lifecycle_transitions_full`, `typography_semantic_mono` — fail through Playwright against the dev server, which is where StrictMode actually double-mounts. Unit tests passing is necessary but not sufficient; worth confirming those three go green before treating this as closed.
2. This PR is still **DIRTY** against main and needs a rebase regardless.

I've discarded my version rather than pushing it — no action needed from you on that.

srobroek:
## Regression fixed, and the guards have moved to follow main's refactor

Two commits pushed.

**`a657610b` — the actual fix.** Adopted the `useMountedRef` extraction that was sitting uncommitted in a worktree (untouched for ~80 minutes; it looked like a casualty of the same crash that interrupted this lane). The hook assigns `true` on mount, which is the part the original inline guard was missing. All five guarded sites route through it; no `useRef(true)` latch pattern remains in `apps/desktop/src`.

**`dec19a15` — merge of main, with the guards relocated.** main has since extracted the Settings fetch logic into `useDataSources` / `useEquipment` / `useNamingPattern`, which conflicted structurally: my guards were in the old inline bodies of `DataSources.tsx`, `Equipment.tsx`, `NamingStructure.tsx`. Resolved by taking main's refactor for all three components and re-applying the guards inside the three hooks, where the unguarded `.then()` sites now live — 6 loaders in total (`loadRoots`; `loadCameras`/`loadTelescopes`/`loadTrains`/`loadFilters`; the naming-pattern load plus `runValidation`/`runPreview`). Taking either side wholesale would have silently dropped one half of the fix.

### Verification

| check | result |
|---|---|
| the 3 regressed e2e specs, on the merged branch | **10 passed** (was 3 failed) |
| `typecheck` (both tsconfigs) | clean |
| `useMountedRef` + all `features/settings` unit tests | 25 files / 163 tests passed |

### Two things I could not clear locally, both pre-existing

- **`pnpm -C apps/desktop lint` reports 19 `noNonNullAssertion` errors.** Not from this PR — they reproduce with my changes stashed, and they're in test files this branch never touches. CI's `Frontend lint` step passes on main at the same commit, so there's a local/CI divergence here worth someone chasing separately; I've left CI to adjudicate rather than paper over it.
- **One unhandled `ReferenceError: window is not defined` in the unit run**, from `@tanstack/react-virtual`'s internal timer firing after teardown in `LogPanel.followScroll.test.tsx`. Same *symptom* as this PR's bug but a different cause — a third-party virtualizer timeout, not a guarded fetch — so it's out of scope here and belongs with the LogPanel lane.

The 3 unit failures in that same run were 15–20s timeouts in unrelated suites (Joyride, devSurface release gate) under a host load average of ~68, not assertions.

srobroek:
## Correction: the lint failure was mine, not pre-existing

In my previous comment I wrote that the 19 `noNonNullAssertion` reports were pre-existing on main and that CI would have to adjudicate. That was wrong, and I reached it by misreading the output.

`noNonNullAssertion` is configured `"warn"` in `apps/desktop/biome.json`, so those 19 never fail anything — they're part of a standing 73-warning background. The run summary said **"Found 1 error"**, and I anchored on the visible wall of warnings instead of chasing that one line. Biome's default diagnostic cap (`Diagnostics not shown: 55`) had hidden it.

The single error was `format` on `src/features/settings/useNamingPattern.ts` — my own file. I'd formatted it earlier, but the second merge resolved that path with `--ours`, which restored the pre-format version underneath me. So it was self-inflicted twice over.

My first fix attempt also silently did nothing: biome runs with `apps/desktop` as its cwd, so passing `apps/desktop/src/...` resolved to a non-existent path and reported "Fixed 1 file" against nothing. Corrected by passing the path relative to `apps/desktop`.

Fixed in `bbcfa1b2`. Now green locally:

| check | result |
|---|---|
| `pnpm -C apps/desktop lint` | passes (0 errors) |
| `typecheck`, both tsconfigs | clean |
| `features/settings` + `hooks` unit tests | 26 files / 167 tests passed |

Worth recording that CI's `UI mock-mode (Playwright)` job already **passed** on the previous head — so the actual regression this PR exists to fix is confirmed resolved by CI, independently of the lint noise above.

PR #1258, opened 2026-07-20
fix: several labels stayed in English when the app language changed

srobroek:
## Summary

- A handful of user-facing strings were never translatable and stayed in English regardless of the selected language: the folder/file path prompts, the calibration master titles, the target details label, and the workflow-profile names in the project wizard. All now come from the message catalog.
- Most consequential: the loading skeleton announced **"Loading"** to screen readers in English no matter the language. Assistive-technology users on a non-English locale heard the wrong language on every loading state.
- Product names (`PixInsight/WBPP`, `Siril`) also move into the catalog. They read identically in every language today, but they are no longer hardcoded, so a locale that transliterates a brand has somewhere to put it.

## Under the hood

The lint rule that is supposed to catch untranslated strings is what let these through, so it gains the sinks it was missing: string concatenation, `*Text`/`*Title`/`*Message`/`*Placeholder`/`*Desc` props, the dialog APIs, and default parameter values. The default-parameter check is what surfaced the screen-reader bug above — it was a real defect, not a style nit. The rule's grandfathered-violation baseline is now empty.

Also included, docs only: the design-refresh wave deltas across five user journeys (#1175, #1176, #1185, #1190) — the Appearance picker rework, the destructive-vs-neutral split on "Approve & apply", and Home/End on table rows. Kept as its own commit if you'd rather it landed separately.

## Verification

- Lint rule suite: 39/39 pass
- Full lint chain: exit 0, baseline drained to 0 violations
- `tsc --noEmit` + e2e tsconfig: clean


srobroek:
Note for whoever merges: the red `cla` on this PR is not a real failure and not this PR's.

The job's conclusion is `cancelled`, not `failure` — it ran 35m then got killed, which `gh pr checks` renders as `fail`. The cause is repo-level: the **CLA Assistant workflow is `disabled_manually`** (workflow 312796618, `.github/workflows/cla-assistant.yml`), so the job can't complete and can't even be re-run — `gh run rerun` returns "Unable to retry disabled workflow".

It's also non-blocking: `main` has no branch protection, so there are no required status checks. Nothing is gated on this.

Every PR opened while the workflow is disabled will carry the same red mark, so it shouldn't be read as a per-PR signal either way.

Issue #813, opened 2026-07-14
Shared .alm-session-detail2 two-col detail layout is a bare CSS-class convention hand-copied into 4 files, not a wrapped component


srobroek:
**Severity**: P3 · **Area**: cross-app (Inbox detail, Sessions detail, Calibration MasterDetail + SessionListPopover)

## Problem
The two-column-properties + stacked-linked-list detail layout (`.alm-session-detail2` and its `__col`/`__linked`/`__actions`/`__muted` sub-classes) is deliberately shared across features — good, matches the "one shared CSS class" half of the design-system mandate. But the JSX structure around that class is hand-copied identically into 4 separate files rather than wrapped in one parameterised component:

- `apps/desktop/src/features/inbox/InboxDetail.tsx:710-789` (`<div className="alm-session-detail2"><div className="alm-session-detail2__col">...`)
- `apps/desktop/src/features/sessions/SessionDetail.tsx:204-229`
- `apps/desktop/src/features/calibration/MasterDetail.tsx:310-333`
- `apps/desktop/src/features/calibration/SessionListPopover.tsx:32-34` (partial reuse of the same class names)

This is the same shape of gap Sessions' own analyst already flagged one level down in #800 (`SessionFrameInventory.tsx` hand-rolling `alm-settings__group` instead of reusing the shared `Section` component) — here it's the two-col detail recipe itself that's duplicated as markup, just kept behaviorally consistent by convention rather than by a shared component contract. A future change to this layout (spacing, a11y attribute, responsive behavior) has to be applied 4 times by hand instead of once.

## Expected
Extract a small shared component (e.g. `TwoColDetailLayout` or similar, alongside `DetailPanel`/`DetailGrid` in `apps/desktop/src/components/`) that owns the `alm-session-detail2` structure, and have Inbox/Sessions/Calibration's two consumers render through it instead of re-declaring the div structure.

## Backend vs UI
UI only.

## Evidence
- `apps/desktop/src/features/inbox/InboxDetail.tsx:710-789`
- `apps/desktop/src/features/sessions/SessionDetail.tsx:204-229`
- `apps/desktop/src/features/calibration/MasterDetail.tsx:310-333`
- `apps/desktop/src/features/calibration/SessionListPopover.tsx:32-34`
- Related: #800 (Sessions, same class of gap one level down)

[...]

Issue #611, opened 2026-07-11
Bulk frame-type override has no heterogeneity warning and no undo


srobroek:
## Problem
Bulk-applying a frame-type override (e.g. setting "light") to a multi-select of files with different detected types is applied instantly and persisted, with no warning that the selection is heterogeneous and no way to undo it. In testing, this stranded calibration files behind target/filter/date metadata they cannot legitimately have.

## Evidence
Reproduced live: applying "light" to 5 mixed calibration files succeeded silently and persisted immediately.

## Recommendation
Add a mismatch warning when the current selection spans differing detected frame types, shown before the override is applied. Add an undo action (or a "reset to detected" action) so a bad bulk override is recoverable.

**Severity:** P1 · **Area:** Inbox bulk reclassify

_Source: automated design review 2026-07-11 (docs/development/design-review-2026-07-11.md, PR #598)._

srobroek:
Reproduced again during Journey 11 (mistake recovery) validation, with DB-level evidence pinpointing exactly what's missing.

Repro: Inbox item `832eea19-6933-4e2e-993a-4b20aa9f8179` ({root}, 2 files, classType=unclassified because no IMAGETYP header was readable). Selected both files via "Needs review" checkboxes, set bulk "Assign frame type" = light, clicked "Apply to selected (2)". No warning/confirmation appeared (there is none in the code path — `handleBulkApply` in `apps/desktop/src/features/inbox/InboxDetail.tsx` has no heterogeneity check at all). Applied instantly; item classification flipped to `light`.

DB evidence (`inbox_classification_evidence`, read via `wizard-test.db`):
```
relative_file_path                                   frame_type  manual_override
dark_exp_120.000000_gain_60_bin_1_44C_stack_9.fits    NULL        light
dark_exp_15.000000_gain_40_bin_1_43C_stack_10.fits    NULL        light
```

[...]

Issue #1114, opened 2026-07-19
Needs-review banner asks for the frame type you just supplied, or explains nothing at all


srobroek:
## What the user sees

A needs-review bucket tells you what is missing. Say a file has no frame type **and** no `EXPTIME`. You supply the frame type through the per-file dropdown and apply.

The file still cannot be promoted — it is missing a mandatory attribute — so it stays in the bucket, and the banner still reads **"frame types required"**.

You did the thing it asked for. It asks again, for the thing you just did.

The other case is worse. `missing_mandatory` is hardcoded to a placeholder `vec!["frameType"]` (`crates/app/inbox/src/reclassify.rs:856`), so a file blocked for want of `gain` can render **no explanation at all** — the user is stuck with nothing on screen saying why.

## Cause

The banner is driven by the grouping set (`mandatory_set_for`) rather than the per-file missing-attribute set (`missingPathAttributes`, the FR-032 set). The partially-resolved case — frame type supplied, mandatory attribute still absent — has no truthful rendering.

## Fix

Populate the existing `missing_mandatory` field with the real per-file missing attributes instead of the placeholder, and drive the banner from `missingPathAttributes` so it names what is actually missing for the file in question.

Estimated ~half a day. Live on `main` today.

## Acceptance

- A file missing only a mandatory attribute never sees "frame types required".
- Every file that blocks promotion renders an explanation naming the specific missing attribute.
- Supplying exactly what the banner asks for either promotes the file or changes what the banner asks for.

## Context

Surfaced while assessing spec 058's on-resolution gap. Independent of 058 — it does not depend on placeholder rows and should not wait for that spec.

srobroek:
**Verified.** `crates/app/inbox/src/reclassify.rs:856` confirmed hardcoded: `missing_mandatory: if is_needs_review { vec!["frameType".to_owned()] } else { vec![] }` with an explicit comment "missing_mandatory population is T070's responsibility; here we surface an empty list." A file blocked on a different mandatory attribute (e.g. `gain`) gets no explanation, matching the described gap. `mandatory_set_for` (grouping-level) is confirmed real in `crates/app/inbox/src/classify.rs:724`; `missingPathAttributes` (per-file, FR-032) is a real concept threaded through `confirm.rs`, `metadata.rs`, and the frontend (`InboxDetail.tsx`, `InboxPage.tsx`) — consistent with the claimed grouping-vs-per-file cause, though I did not trace the exact banner-rendering call site.

[...]

PR #1260, opened 2026-07-20
refactor: split the inbox detail pane into per-surface modules

### LINKED ISSUE #994: refactor: split InboxDetail.tsx (979 lines)
## Oversized source file

- **File:** `apps/desktop/src/features/inbox/InboxDetail.tsx`
- **Size:** 979 lines (TS/React)
- **Tier:** P3 (moderate)

## Why

This file is well past a reviewable/maintainable length. Large files slow review, inflate incremental TS/React rebuild/test cost, and hide cohesion boundaries.

## Suggested approach

Extract cohesive units: pure helpers into a colocated `*.ts`, sub-components into their own `*.tsx`, data/query hooks into a `store`/hooks module. Keep the exported component API stable. Use the shared UI component (never cloned CSS) per the project mandate.

**Scope the concrete split before implementing** — this issue records the debt; the split boundaries should be proposed in the PR (or a short design note) since they depend on the file's internal structure. Behaviour must not change; keep the public API stable and rely on existing tests as the regression net.

## Acceptance

- [ ] File split into modules each with a single clear responsibility
- [ ] No file in the split exceeds the threshold (Rust ~800, TS ~500)
- [ ] Public API unchanged; all existing tests pass
- [ ] `just lint` / `just test` / `just typecheck` green

srobroek:
Closes #994.

`InboxDetail.tsx` went **1472 → 431 lines**, split along the seams that
concurrent lanes actually collide on.

## Why (the ticket's stated reason is the weaker one)

The issue argues line count. The real cost is in the git log: this file took
**12 commits in one week** from unrelated lanes — layout migration, design-system
rename, confirm staleness, workflow/manifests, button variants, reclassification
overrides, no-metadata state, render loop, dest-root picker. It is a
merge-conflict magnet, and the lanes touch genuinely different regions
(#1126 was in needs-review + selection helpers; #1188 was in the layout row).

Those regions map cleanly onto separate modules, so this decouples the lanes
rather than just shuffling code.

## The split

| module | lines | what it owns |
|---|---|---|
| `InboxDetail.tsx` | 431 | props, layout, render composition |
| `InboxNeedsReview.tsx` | 337 | unclassified table, bulk overrides, #611 gate + undo |
| `InboxFilesColumn.tsx` | 217 | mixed summary, metadata popover, FR-032 banner |
| `useInboxReclassifyState.ts` | 276 | all reclassify state around the IPC call |
| `inboxDetectionProps.ts` | 146 | detection property table builder |
| `inboxDetailHelpers.ts` | 139 | pure helpers |
| `useInboxReclassifyV2.ts` | 84 | the IPC call |
| `FileInspector.tsx` | 90 | popover row detail |

Behaviour is unchanged — every extraction is a move plus the import wiring.
`InboxNeedsReview` takes the whole `useInboxReclassifyState` result rather than
20 separate props, so that boundary stays stable as the hook grows.

## One non-obvious thing

`TwoColDetailLayout` allocates a `__col` per **non-null** `extraCols` entry, so
a component that merely renders `null` internally still claims a column. The
empty-needs-review case therefore stays a call-site ternary. The #813 layout
test caught this — it's the one regression this refactor introduced and fixed.

## Verification

- `pnpm -r typecheck` — green
- full desktop suite — **1835 pass, 0 fail**
- `biome check` — clean
- `knip` — no unused files or exports from the split (the 6 unused-*dependency*
  hits are pre-existing and unrelated)

## Note for in-flight inbox lanes

#1126 touches the needs-review region and will need to rebase across this
regardless of merge order — that region is now `InboxNeedsReview.tsx`.


srobroek:
Conflict with `main` — head SHA `c3ad7d9be70fb4a80455d7f140c9b8fa6c4900f3`, main at `da70b292`.

`mergeable` is CONFLICTING / `mergeStateStatus` DIRTY. Conflicting file:

- `apps/desktop/src/features/inbox/InboxDetail.tsx`

Cause: `da70b292` — "fix(inbox): needs-review names the attribute that is actually missing (#1126)" — landed on main and edits `InboxDetail.tsx`, which is the same file this PR splits into per-surface modules. The overlap is structural rather than incidental: #1126 changed needs-review text inside the region being extracted, so the change needs to be carried into whichever new module now owns that code (`InboxNeedsReview.tsx` looks like the likely destination) rather than resolved in place.

Worth knowing while resolving: GitHub skips PR CI entirely on CONFLICTING PRs, so the absence of check results here is a consequence of the conflict, not a missing workflow. Checks will only start once this merges cleanly.

No conflict with #1194 (`spec/058-inbox-drop-parent-items`) despite both being Inbox work — the two touch disjoint file sets (that one is almost entirely `crates/app/inbox/`).

srobroek:
Resolved in `117f96d5`. `mergeable` is now MERGEABLE and CI has started.

Your read on the resolution was right, with one split: #1126's change spans two
of the new modules, not one.

- `filesNeedingReview` — the #1114 union gate that keeps a file in review while
  it is still blocked on a mandatory attribute — went into
  `useInboxReclassifyState.ts`, not `InboxNeedsReview.tsx`. It is selection
  state (`handleSelectAll` consumes it), and the hook owns selection; the
  component only renders. It is returned in place of `unclassifiedFiles`.
- `allSelected`/`someSelected` did land in `InboxNeedsReview.tsx` as you
  predicted.
- `blockingAttrsExcludingFrameType` and the banner copy stayed in
  `InboxDetail.tsx` — the unclassified banner is above the extracted region, so
  that hunk merged clean.

Main's side of the large conflict hunk was the pre-split inline copy of the
reclassify handlers, which `useInboxReclassifyState` already contained verbatim
(it was extracted from exactly that code). I dropped it as duplicate and ported
only #1126's delta, rather than taking either side wholesale.

Evidence the port is faithful: #1126's four new tests came in with the merge and
pass unmodified against the split — inbox went 202 → 206 tests, and the full
desktop suite is 1845 pass / 0 fail.

PR #1264, opened 2026-07-20
feat(inbox): scanned folders show up as rows you cannot confirm

srobroek:
## Summary

A folder the scan has found but nothing has classified yet now appears in the
Inbox list as a row of its own. Previously such a folder had no way to reach the
list at all.

The row deliberately offers no Confirm action — there is nothing to confirm
about a folder whose contents are still unknown. Rather than showing a Confirm
button that refuses to work, the row simply has no confirm affordance, and its
detail panel explains what the folder is and why nothing can be done with it
yet.

## What's new

- Scanned-but-unclassified folders are visible in the Inbox list, showing their
  path, file count and format
- Selecting one opens a detail panel describing the folder, with no Confirm
  control
- These rows are counted in the Inbox summary and header, so the counts match
  the rows on screen
- The FITS/video filter includes them by their file format; a specific
  frame-type filter (bias, dark, flat, light) hides them, since an unclassified
  folder has no frame type
- Grouping the list keeps them visible instead of dropping them

## Note

These rows do not appear in the running app yet: every scanned folder still
gets a placeholder row, so the underlying list is always empty for now. This
change is the consumer that has to exist before that placeholder can be
removed. It is covered by unit tests rather than by on-screen behaviour.

## Spec Context

Spec 058 **T013's UI half**. `inbox.list` has returned `sourceGroups` since the
backend half landed, but nothing in the app read it (`grep`: only `mocks.ts`,
two fixtures and the generated bindings).

**Why this is the unblocking step.** T020 wants to delete the scan-time
placeholder row. It is blocked on two independent things, which the task notes
conflate:

1. **No group-scoped classification entry point** — `classify()` is keyed on
   `inbox_item_id`; `reclassify_v2` rebuilds from evidence only ever written
   against an item id. This design decision **still does not exist** and is
   T012's real blocker. *Not addressed here.*
2. **No UI consumer of `sourceGroups`** — with no source-group row rendered or
   selectable, deleting the placeholder makes every scanned folder invisible on
   top of unclassifiable.

This PR removes (2) only. T012/T019/T020/T024 stay blocked, and landing this
does not by itself make the placeholder deletable.

**Structurally non-confirmable, not guarded** (FR-016): the row carries no
`inboxItemId`, selection travels through `onSelectSourceGroup` and its own
`?selectedGroup` search param kept separate from `?selected`, and it opens
`InboxSourceGroupDetail`, which renders no confirm control — not a disabled
one. Confirm reads the selected *item*, so there is no value it could be
handed. A disabled button would have re-introduced the
row-claims-an-action-it-cannot-perform shape #711 is about.

**Two filter traps, both pinned by tests.** The lane filter reads `format`, not
`lane` — a source group's `lane` is the `move`/`catalogue` lane, a different
axis sharing the name, and filtering on it would match `"move"` against
`"fits"` and hide every group. An active kind filter drops source groups
entirely.

**Counts (T022 / CHK010).** Source-group rows are counted in `deriveInboxStats`
and the header. This also settles a disagreement already on the branch: the
stats strip had moved to `filteredItems` for SC-004 while the header still
counted the unfiltered `items`, so with any filter active the two numbers
beside each other disagreed.

**Verification.** `vitest` 1856 passed (204 files), including 7 new
source-group tests and 3 new stats tests; `tsc --noEmit`, `eslint` (i18n
baseline), token-policy and i18n catalogue checks all clean. Two-direction
control on both load-bearing assertions: routing a source-group click to
`onSelect` fails the non-confirmability test, and making the lane filter read
`group.lane` fails the filter test.

Refs #711

### DIFF
--- a/apps/desktop/messages/en.json
+++ b/apps/desktop/messages/en.json
@@ -1949,6 +1949,10 @@
     }
   ],
   "inbox_list_root_label": "(root)",
+  "inbox_source_group_state": "not classified",
+  "inbox_source_group_detail_title": "Scanned folder",
+  "inbox_source_group_detail_body": "This folder has been scanned but not classified yet, so there is nothing to confirm. Classify it to see its frames.",
+  "inbox_source_group_scanned_at": "Discovered {when}",
   "inbox_dest_root_label": "Library:",
   "inbox_dest_root_aria": "Destination library",
   "inbox_no_file_metadata": "No file metadata",
--- a/apps/desktop/src/app/router.tsx
+++ b/apps/desktop/src/app/router.tsx
@@ -81,6 +81,11 @@ const inboxRoute = createRoute({
     // search/filters change the array shape (Sessions/Calibration already use
     // this id-based pattern).
     selected: parseString,
+    // Spec 058 T013: a scanned-but-unclassified folder is selected through its
+    // OWN param, never through `selected`. Keeping the two apart is what makes
+    // a source-group row structurally non-confirmable: confirm reads the
+    // selected item, and a source group can never become one.
+    selectedGroup: parseString,
     type: parseEnum(FRAME_TYPES),
     group: parseEnum(INBOX_GROUPS),
   }),
--- a/apps/desktop/src/features/inbox/InboxList.tsx
+++ b/apps/desktop/src/features/inbox/InboxList.tsx
@@ -15,6 +15,10 @@
  * page / top-bar (FilterToolbar + useGrouping); this is a controlled
  * presentational list.
  *
+ * Rows are of two kinds: inbox items, and — spec 058 T013/FR-016 —
+ * scanned-but-unclassified source-group folders, which are visible and
+ * selectable but have no item id and therefore nothing to confirm.
+ *
  * Sort: column headers are <button> elements that call onSort (SessionsTable
  * convention). The page owns sort state and passes sortCol + sortDir.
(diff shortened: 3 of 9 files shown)


srobroek:
Superseded by `de8e5492` ("show scanned-but-unclassified folders as non-confirmable rows"), which landed the same T013 UI half on `spec/058-inbox-drop-parent-items` while this was in review. Two lanes picked up the task in parallel; theirs landed first.

The overlap is the row rendering, the `format`-not-`lane` filter guard (#854), the kind-filter exclusion, and the `deriveInboxStats` CHK010 fold — all equivalent, so nothing is lost by dropping this.

Three things here are NOT in `de8e5492` and are moving to a follow-up branched off it, rather than rebasing this PR into a conflict-heavy rewrite of code that already landed:

1. **Selectable source-group rows.** `de8e5492` renders them with no `_onClick`/`_selected` and no detail pane. T015 ("selecting a source-group row asserts Confirm is absent") and T017/CHK011 ("N=0 → the source-group row stays selected") both require selection to exist, and T012's blocker note names "rendered **or selectable**".
2. **The header/stats count disagreement**, still live at the tip: `folderCount` reads unfiltered `items` (InboxPage.tsx:1007) while `derivedStats` reads `filteredItems` + `filteredSourceGroups`. With any filter active the two adjacent numbers disagree, and source groups count in one but not the other.
3. Interleaved sort placement instead of a permanent leading block (minor, may not carry over).

Closing as superseded.

Issue #1257, opened 2026-07-20
Spec 054: regression cover for shipped dock behaviour (T026 pinned columns, T023 reload + drag)


srobroek:
Follow-up to #1158, which #1253 closes. The three gaps #1158 tracked are now **delivered** — but delivering them created a genuine gap between shipped behaviour and its regression cover. Filing that here so closing #1158 doesn't drop it.

Both tasks are test coverage for behaviour that already ships. Neither is unbuilt product.

### T026 — real-browser assertion for the pinned Targets columns

The star + designation columns are pinned via `position: sticky` (PR #1253, CSS-only, `merges-2.css`). Nothing would currently catch a regression.

**A Layer-1 test cannot substitute.** jsdom has no layout engine, so it cannot observe sticky offsets — every wrong version of this feature passed static review. It took three measurement iterations to land, and the two near-misses were invisible without measuring:

- declaring `45px` rendered a 50px column, leaving the designation drifting 270px → 265px
- a percentage offset resolved against the *scroll container* (760px), not the table, drifting 12px

Manual verification measured **0px drift** for star, designation, and the designation header against **240px of real h-scroll**, at 1400×900 with a 420px side dock. That is the assertion to automate.

Home: `crates/e2e-tests/tests/targets_journeys.rs` (Layer-2, tauri-driver).

The pinning is reachable in the **default** side-dock configuration, not an edge case: at 1400px the sidebar takes 220px and the dock 420px, leaving the table ~760px against its own 1000px min-width floor.

### T023 — reload + drag halves

`useAdaptiveDock` persistence is covered across a **remount**, scoped per `dockId` (#1195). Two parts remain uncovered:

- restore across a real page **reload**, which is what actually exercises the preference cache
- width set by **dragging** the resize handle rather than calling `setWidth` — the drag path has its own pointer-event handling

Neither shortcut is free, which is why T023 was deliberately left open rather than ticked.

---

Spec: `specs/054-adaptive-detail-dock/tasks.md`. These are the only two tasks still Open; everything else in spec 054 is delivered or explicitly withdrawn/N-A.

srobroek:
**Scoping note: T026 is blocked on harness + CI capability, not just test authoring.** Found while surveying `crates/e2e-tests`:

[...]

PR #1265, opened 2026-07-20
test: cover the detail dock's pinned columns, drag-resize, and persistence

srobroek:
Closes the T026 half of #1257.

**Stacked on #1253** (base is `feat/1158-pinned-targets-columns`, not `main`) because it asserts the CSS that PR ships. Retarget to `main` once #1253 lands.

## Why this can't be a Layer-1 test

The pinning took three measurement iterations to land, and **every wrong version passed static review**. Both near-misses were invisible without measuring:

- a declared `45px` star column rendered at 50px, drifting the designation 270px → 265px
- a percentage offset resolved against the *scroll container* (760px) instead of the table, drifting 12px

jsdom has no layout engine, so a Layer-1 test reports `0` for every one of these positions and would pass against a **completely unpinned table**. This needs real `getBoundingClientRect` geometry.

## The task understated the work

T026 read as "add an assertion". It couldn't be, because the behaviour didn't occur in CI at all. Two shared-surface gaps had to close first — flagging that explicitly, since both sit in the path of *every* real-UI journey:

**1. Nothing could size the window.** `tauri.conf.json` opens at 1280×820; `useAdaptiveDock`'s threshold is `window.innerWidth >= 1400`. The docked layout simply never happened.

`E2eApp::set_viewport` is deliberately not a one-line `set_window_rect`. That call sizes the **outer** window while every threshold keys off **innerWidth**, and the gap between them is window chrome — ~0 under xvfb (no WM, no decorations) but real on Windows. Sizing blind would engage the dock on Linux and silently not on Windows, where the journey would pass having asserted nothing. So it sets, measures, corrects by the observed delta, and verifies.

**2. The Ubuntu shards' display was too narrow to hold the window.** `e2e.yml` ran `xvfb-run -a`, which takes the built-in default:

```
/usr/bin/xvfb-run:16:  XVFBARGS="-screen 0 1280x1024x24"
```

1280 < 1400, so the resize would be clamped at the screen edge. Widened to 1600×1200 — plain framebuffer memory (~7.7MB), not GPU, on a runner that already has none.

## Guard against a vacuous pass

The journey asserts a **non-pinned control column moves by the scroll distance**. Without it, this would pass in the exact case it most needs to catch: if the table never scrolled, every `drift == 0` assertion would hold trivially against a static table.

## Not verified

**The Windows shards' native display resolution has not been confirmed ≥1400.** If it's smaller, `set_viewport` fails loudly naming the screen size rather than passing vacuously — the failure mode is safe, but that shard would need the same treatment. I'd rather surface this than let a green run imply I checked it.

## Spec

T026 is ticked **in this PR** rather than left for a later reconciliation pass — that lag is what #1119 and the #1195 pass each had to clean up. Spec 054's Open count drops 2 → 1; T023's reload + drag halves stay open and tracked in #1257.

## Verification

`cargo check`, `cargo clippy`, `cargo fmt` clean on `e2e_tests`. The journey itself is `#[ignore]`d like every Layer-2 test and runs on the e2e shards — **its first real execution is this PR's CI**, which is also the first exercise of the widened display.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -732,12 +732,21 @@ jobs:
       # its default DMA-BUF/compositing render paths were suspected of adding
       # ~25-30s of per-test overhead vs Windows. Linux-only; harmless no-op if
       # the hypothesis is wrong.
+      #
+      # `-s "-screen 0 1600x1200x24"` overrides xvfb-run's built-in default of
+      # 1280x1024x24 (`XVFBARGS` in /usr/bin/xvfb-run). 1280 is NARROWER than
+      # the 1400px viewport the docked-layout journeys request via
+      # `E2eApp::set_viewport`, and a resize past the screen edge is clamped —
+      # which would silently yield a too-small viewport and a journey that
+      # passes while asserting nothing. The larger framebuffer is plain memory
+      # (1600*1200*4 = ~7.7MB), not GPU, so it costs nothing on a runner that
+      # already has none.
       - name: Run E2E (xvfb)
         env:
           WEBKIT_DISABLE_DMABUF_RENDERER: "1"
           WEBKIT_DISABLE_COMPOSITING_MODE: "1"
           ALM_E2E_APP_BIN: ${{ github.workspace }}/${{ env.APP_BIN }}
-        run: xvfb-run -a cargo nextest run --archive-file e2e-archive.tar.zst --workspace-remap . --profile e2e --run-ignored all --no-tests=warn --partition hash:${{ matrix.shard }}/4
+        run: xvfb-run -s "-screen 0 1600x1200x24" -a cargo nextest run --archive-file e2e-archive.tar.zst --workspace-remap . --profile e2e --run-ignored all --no-tests=warn --partition hash:${{ matrix.shard }}/4
 
   # Copy of real-ui-e2e-ubuntu for Windows (no Linux deps/xvfb/chmod).
   real-ui-e2e-windows:
--- a/apps/desktop/src/components/ListPageLayout.test.tsx
+++ b/apps/desktop/src/components/ListPageLayout.test.tsx
@@ -702,7 +702,11 @@ describe('adaptive dock — drag + real persistence (T023)', () => {
   it('a drag on the resize handle really resizes the side panel', () => {
     resizeWindowTo(1600);
     const { getByTestId, container } = render(
-      <ListPageLayout topBar={<div>bar</div>} detail={<div>detail</div>} dockId="drag-page">
+      <ListPageLayout
+        topBar={<div>bar</div>}
+        detail={<div>detail</div>}
+        dockId="drag-page"
+      >
         <div>main</div>
       </ListPageLayout>,
     );
@@ -721,7 +725,11 @@ describe('adaptive dock — drag + real persistence (T023)', () => {
   it('the dragged width reaches localStorage, not just the in-memory cache', () => {
     resizeWindowTo(1600);
     const { getByTestId } = render(
-      <ListPageLayout topBar={<div>bar</div>} detail={<div>detail</div>} dockId="drag-persist">
+      <ListPageLayout
+        topBar={<div>bar</div>}
+        detail={<div>detail</div>}
+        dockId="drag-persist"
+      >
         <div>main</div>
       </ListPageLayout>,
     );
@@ -738,7 +746,11 @@ describe('adaptive dock — drag + real persistence (T023)', () => {
   it('a drag is clamped to the window fraction before it is persisted', () => {
     resizeWindowTo(1600);
     const { getByTestId } = render(
-      <ListPageLayout topBar={<div>bar</div>} detail={<div>detail</div>} dockId="drag-clamp">
+      <ListPageLayout
+        topBar={<div>bar</div>}
+        detail={<div>detail</div>}
+        dockId="drag-clamp"
+      >
         <div>main</div>
       </ListPageLayout>,
     );
@@ -755,7 +767,9 @@ describe('adaptive dock — drag + real persistence (T023)', () => {
 
   it('a pinned placement reaches localStorage too', () => {
     resizeWindowTo(1024);
-    const { result } = renderHook(() => useAdaptiveDock({ dockId: 'pin-persist' }));
+    const { result } = renderHook(() =>
+      useAdaptiveDock({ dockId: 'pin-persist' }),
+    );
     expect(result.current.placement).toBe('bottom');
 
     act(() => {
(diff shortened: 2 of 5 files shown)


srobroek:
Rebased onto `main` (#1253 merged, so the base moved and the stacked commits are now in `main` via squash — replayed only this branch's four commits with `git rebase --onto`). Retargeted `feat/1158-...` → `main`.

**Scope grew: this now closes both halves of #1257, not just T026.** Added T023's drag path and — more importantly — its persistence half.

## The remount tests are green against total persistence failure

T023 suspected the remount coverage was weaker than it looked. It is, and the mechanism is concrete:

`getPreferences()` returns a module-level `cachedPreferences` when one exists, and `persistPreferences()` assigns that cache **before** its `localStorage.setItem`, whose failure is deliberately swallowed (*"state is still in memory"*). So a remount re-reads the **cache**, never storage.

Measured rather than argued. With `setItem` stubbed out so nothing whatsoever persists:

| | Result |
|---|---|
| The 3 new storage assertions | **FAIL** ✅ |
| The new drag-resize test (asserts no persistence) | passes — correct, not a persistence test |
| Existing `persists across remount` tests | **PASS** ❌ |

The existing tests report green while the feature is completely broken. That is the gap T023 was left open for, now closed at Layer 1.

The new tests read the serialized bytes under the real `alm-preferences` key rather than reading back through the module, since going through the module would consult the cache and prove nothing. They also pin the clamp — the **clamped** width must persist, or a cold start restores an over-wide dock.

## Still not covered

A genuine **reload** (cold module cache reading from real storage) remains Layer-2. `E2eApp::relaunch()` exists for exactly this and preserves webview storage, so it is the natural home — deliberately not bundled into this PR, which is already carrying shared harness and CI changes.

## Note on the i18n gap

`pnpm vitest run` bypasses the `pretest` hook that compiles the Paraglide catalogue, so running a single spec file directly fails at import with `Failed to resolve import "@/paraglide/messages"` — and the tooling summarised that suite-level failure as `PASS (0) FAIL (0)`, which reads as success. Not changed here, just flagging: a zero-test run and a passing run are not distinguishable at a glance.

srobroek:
**The three-OS failure is just formatting, not a real test break.**

`Frontend lint (Biome + ESLint gates + tokens)` fails identically on ubuntu/macos/windows because it's deterministic — `biome check` wants the JSX in `src/components/ListPageLayout.test.tsx:705` broken across lines:

```
- <ListPageLayout topBar={<div>bar</div>} detail={<div>detail</div>} dockId="drag-page">
+ <ListPageLayout
+   topBar={<div>bar</div>}
+   detail={<div>detail</div>}
+   dockId="drag-page"
+ >
```

`pnpm --filter @astro-plan/desktop exec biome check --write .` and push should clear all three legs. No test logic involved.

srobroek:
Agreed, and already pushed — `be5267d0` ran `biome check --write` over the desktop package. Same diagnosis, reached independently about the same minute; thanks for the confirmation.

Two corrections to my own earlier read, for the record:

- I initially characterised this as a **windows/macos** failure. Wrong: all three legs failed, ubuntu included. I sampled the rollup before ubuntu had finished and drew a platform-specific conclusion from an incomplete set — exactly the wrong inference for a deterministic lint gate.
- The miss was mine: I ran `cargo fmt` + `clippy` for this PR's Rust half and never ran biome over the TypeScript half before pushing.

Suite is still 31 passing locally after the reformat; the three legs are re-running now.

Unrelated but worth flagging since it shows up in local runs of this gate: `pnpm lint` also exits 1 locally on a fatal ESLint parse error in `eslint-rules/no-user-string.js` (*"was not found by the project service"*). Confirmed **pre-existing** by stashing my changes and re-running on clean `main`, and it does not reproduce in CI. Untouched here.

srobroek:
**macOS L1 is red here for a pre-existing `main` bug, not for anything in this PR.**

The lint failure is fixed (`be5267d0`). The new macOS failure is in `Frontend unit tests`, and it never reaches a test:

```
/Users/runner/work/_temp/....sh: line 31: mapfile: command not found
##[error]Process completed with exit code 127
```

`mapfile` is a bash 4 builtin; macOS runners ship bash 3.2 (Apple ships no GPLv3 bash). It's at `ci.yml:382`, introduced by #1212 — a file this PR does not touch.

It stayed invisible on `main` because that code path only runs on a **targeted** frontend run. A PR that doesn't change `apps/desktop/src/**` never reaches it. This PR changes `ListPageLayout.test.tsx`, so it does.

Already fixed in **#1282** / **#1283** (two lanes filed the same fix a minute apart; noted on #1283). This PR is blocked on one of those landing, then a rerun — no further change needed here.

Windows and ubuntu legs are still running; expected green now the formatting is fixed.

srobroek:
**Correction to my earlier comment: the `pnpm lint` claim was wrong.**

I said `pnpm lint` exits 1 locally on a pre-existing fatal ESLint parse error in `eslint-rules/no-user-string.js`, and called it pre-existing on `main`. **It does not.** `pnpm lint` exits **0** — on a fresh worktree off clean `main`, and in this PR's worktree.

What actually happened: I read an exit code that did not reproduce. Re-running the same command several ways (`>/dev/null`, redirected to files, each link of the `&&` chain individually) returns 0 every time, and the run ends with all four gates reporting OK.

What was real is only the *output*: ESLint does emit a fatal `was not found by the project service` diagnostic for `eslint-rules/no-user-string.js`, and `check-eslint-baseline.mjs` prints it while still exiting 0 — it's tolerated by the baseline. So the custom rule's own source isn't type-aware-linted. That's a minor gap, not a broken gate, and not something this PR should touch.

I'd rather flag the retraction than leave a wrong "pre-existing failure" claim sitting in the thread for someone to build on — especially since I used it to argue something wasn't mine.

The macOS `mapfile` blocker in my previous comment is unaffected: that one is reproducible, has a concrete exit 127, and is already fixed in #1282/#1283.

srobroek:
**Rebased onto `main` (`f6e6e92c`); the conflict is resolved and this is mergeable again.**

Whoever rebased this branch earlier: your replay was intact — I diffed the patch my seven commits introduce against yours, restricted to the two files I authored (`.github/workflows/e2e.yml`, `crates/e2e-tests/tests/common/mod.rs`), and they were byte-identical. Nothing was lost. My local checkout was just stale, so I reset onto your version before rebasing.

The conflict was a single line, and both sides were wanted:

- `#1216` switched the Ubuntu shards from `--partition count:N/4` to `hash:N/4`.
- My commit added `-s "-screen 0 1600x1200x24"` so the shards get a display wide enough for the docked layout.

Resolved by keeping **both** — `hash:` partitioning with the widened display.

**One thing worth stating plainly:** the pinned-column assertion in this PR has still never executed. Every Real-UI E2E job on the previous head was cancelled by concurrency, so the earlier "0 failures" reports here were consistent with a suite that never ran. This push is the first run that should actually exercise it. I'm not claiming it's green until the L3 jobs report.

PR #1267, opened 2026-07-20
chore(hooks): PR issue-refs hook reads the branch and repo being merged, not the shared checkout

srobroek:
## Summary

- Syncs `speckit-pr-issue-refs.sh` (both the `.claude` and `.codex` copies) to the upstream version from srobroek/agentic-packages#548.
- Supersedes #1238, which fixed two of the three defects locally. This carries all three.

## Why not just `apm install`

That was the plan, and it does not work. I ran it:

- The hook is reported as **"local files exist, not managed by APM"** — one of 50 files skipped. Plain `apm install` will neither overwrite it nor deliver the fix.
- The cached module is pinned to `resolved_commit bb7a3781`, which predates #548, so even the vendored `apm_modules/` copy is stale.

So the correction to the #1238 review note: this file is **not** at risk of being silently reverted by `apm install` — but the upstream fix also will **not** arrive on its own. It has to be synced deliberately, which is what this PR does.

## What the three fixes are

1. **Branch resolution** — read from the command's `cwd`, not the hook's `$PWD`. With worktrees those differ, which previously injected one lane's spec context into unrelated PRs.
2. **`--head` trust order** — an explicit `--head` names the branch outright and wins over checkout state. Upstream also strips the `owner:` prefix, which #1238 did not.
3. **Repo resolution** — `git -C "$CWD" remote get-url origin`. Previously read from the hook's cwd, so a lane working across two repos queried the *other* repository's issues. #1238 missed this one.

Plus a fail-closed guard: spec context is only emitted when `specs/<id>-*` actually exists.

## Verification

`bash -n` clean, and the behaviour was checked in both directions rather than only the positive case:

| test | setup | result |
|---|---|---|
| T1 | `cwd` = a worktree on `spec/058-…`, hook `$PWD` on a non-spec branch | emits spec 058 + its 3 real open issues |
| T2 | `cwd` = a non-spec branch | no spec context (no false positive) |
| T3 | `--head platevault:spec/058-…` from a non-spec `cwd` | emits spec 058 — `--head` wins, `owner:` stripped |
| T4 | `--head platevault:spec/999-does-not-exist` | no spec context — fail-closed guard holds |

## Note

`apm install` also produced an unrelated `apm.lock.yaml` diff (codex MCP deployment bookkeeping). Left out deliberately to keep this focused and avoid colliding with other lanes.

### DIFF
--- a/.claude/hooks/speckit/scripts/speckit-pr-issue-refs.sh
+++ b/.claude/hooks/speckit/scripts/speckit-pr-issue-refs.sh
@@ -47,8 +47,25 @@ Spec context goes at the bottom under "## Spec Context" (not in the title).
 The release-please draft PR will be manually curated before publish -- raw material matters.
 GUIDANCE
 
-# Detect current branch
-BRANCH=$(git branch --show-current 2>/dev/null)
+# Detect current branch.
+#
+# Read the branch from the directory the COMMAND runs in, not the hook's own
+# $PWD. With git worktrees those differ: `gh pr create` runs in the worktree
+# while the hook inherits the main checkout, so using $PWD reports whatever the
+# shared checkout happens to be on. In practice that injected another lane's
+# spec context into five unrelated PRs. Both Claude and Codex put the directory
+# in `.cwd`; fall back to $PWD when absent or not a directory, matching
+# packages/hooks-git-safety/scripts/git-guard.sh.
+CWD=$(echo "$INPUT" | jq -r '.cwd // empty' 2>/dev/null)
+[ -n "$CWD" ] && [ "$CWD" != "null" ] && [ -d "$CWD" ] || CWD="$PWD"
+
+# Order of trust: an explicit `--head` on the command names the branch outright,
+# so prefer it over any checkout state. `--head` may be `owner:branch` for a
+# fork; keep only the branch part.
+BRANCH=$(echo "$COMMAND" | grep -oE -- '--head[= ]+[^ ]+' | head -1 | sed -E 's/--head[= ]+//; s/^[^:]*://')
+if [ -z "$BRANCH" ]; then
+  BRANCH=$(git -C "$CWD" branch --show-current 2>/dev/null)
+fi
 if [ -z "$BRANCH" ]; then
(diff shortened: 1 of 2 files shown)


srobroek:
Update: #1238 was already merged while I was preparing this, so its partial fix is now on `main`. I've merged `main` in and resolved in favour of the complete upstream file — it's a strict superset, so the resolution takes it wholesale rather than reconciling hunk-by-hunk.

The resolved file is byte-identical to upstream `main`. Re-verified after the merge: `bash -n` clean, spec-branch detection via `cwd` still resolves 058 correctly, and the fail-closed guard still rejects a non-existent `spec/999`.

So this PR is now "finish the job #1238 started" rather than "replace it" — the remaining delta is the third defect (repo resolved from the command's `cwd`) and the `owner:` prefix stripping on `--head`.

srobroek:
Closing — superseded by #1270, which is the same content at the correct layer.

I need to correct what I asserted in this PR's description. I claimed plain `apm install` does not overwrite this file, based on an install run that reported it among `50 files skipped -- local files exist, not managed by APM`. **That test was inconclusive and the conclusion was wrong.** At the time I ran it, the on-disk file was still byte-identical to the packaged version, so an overwrite would have produced no diff either way — I read "no change" as "skipped" when it was actually "overwritten with the same bytes".

The file *is* managed: it appears in `apm.lock.yaml` under the package's `deployed_files`. Managed ⇒ overwritten on plain `apm install`. So the original review note on #1238 was right and my rebuttal was not — the skipped-files count says nothing about whether a *specific* file is safe.

Which makes this PR the same wrong-layer mistake #1238 made, just with better content. #1270 does it properly: refreshes the pin via `apm update --yes` and commits the regenerated files plus the lock, so the fix survives the next install instead of being silently reverted.

Verified before closing: #1270's `.claude` copy is byte-identical to upstream `main`, and it updates the `.codex` copy too.

The behavioural verification from this PR still stands and applies to #1270's file unchanged — `cwd`-based branch resolution, `--head` trust with `owner:` stripping, and the fail-closed `specs/<id>-*` guard, each checked in both directions.

PR #1273, opened 2026-07-20
fix: running a single test file no longer fails on a missing message catalogue

srobroek:
## Summary

- `src/paraglide/` is generated and gitignored, and `src/lib/i18n.ts` imports from it. On a tree that hasn't compiled it, **every** suite fails at import with `Failed to resolve import "@/paraglide/messages"`.
- `package.json`'s `pretest` hook covers `pnpm test`, but **not** `pnpm vitest run <file>`, watch mode, or an IDE runner invoking vitest directly — exactly the commands used to iterate on one spec. The papercut landed on the narrow inner loop while the full-suite path stayed green, which is why it survived.
- A `globalSetup` runs once per vitest process regardless of entrypoint, so it covers all of them.

## Why this is worth fixing

Not the inconvenience — the **shape of the failure**. A suite that cannot load reports as a failed file with *"no tests"*, which summarises into something that reads like success. I misread it exactly that way earlier today before checking the raw JSON, which is what prompted this.

## Verification

Both directions, on a tree with `src/paraglide/` deleted:

| | Result |
|---|---|
| globalSetup disabled | `1 failed / no tests` |
| globalSetup enabled | `27 passed` |

Full suite unaffected: **201 files / 1845 tests** green. `pnpm typecheck` clean.

## Design notes

- `pretest` is deliberately left in place: it costs nothing and keeps `pnpm test` working if this file is ever removed.
- A compile failure is only fatal when **no** previous output exists. Otherwise it warns and reuses the existing catalogue, so a transient CLI problem can't block an otherwise-passing run — while a genuinely absent catalogue still fails loudly with the command to diagnose it.

## Pre-existing, not from this PR

`pnpm lint` reports one fatal parse error in `eslint-rules/no-user-string.js` (*"was not found by the project service"*). Confirmed pre-existing by stashing these changes and re-running on clean `main` — identical `fatalErrorCount: 1`. Untouched here.

## Spec Context

Found while adding Layer-2 coverage for spec 054 (#1257 / #1265); unrelated to that spec, so it ships separately.

### DIFF
--- a/apps/desktop/vitest.config.ts
+++ b/apps/desktop/vitest.config.ts
@@ -23,6 +23,13 @@ export default defineConfig({
     environment: "jsdom",
     globals: true,
     setupFiles: ["./vitest.setup.ts"],
+    // Compiles the generated Paraglide catalogue before any suite loads.
+    // package.json's `pretest` hook only runs for `pnpm test`; invoking
+    // vitest directly (single file, watch mode, IDE runner) skipped it, and
+    // every suite then failed at import on `@/paraglide/messages`. A
+    // globalSetup runs once per vitest process whatever the entrypoint.
+    // See vitest.globalSetup.ts.
+    globalSetup: ["./vitest.globalSetup.ts"],
     css: false,
     // Vitest's 5s default is too tight for jsdom + React render + waitFor
     // chains once the runner is CPU-contended — and CI runners are contended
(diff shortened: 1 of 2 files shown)


srobroek:
**Correction to this PR's description.** Under "Pre-existing, not from this PR" I claimed `pnpm lint` reports a fatal parse error in `eslint-rules/no-user-string.js` and exits 1 on clean `main`. The **exit code claim is wrong** — `pnpm lint` exits 0, verified on a fresh worktree off `main` and repeatedly under different redirections.

Only the diagnostic output is real: ESLint does emit that fatal for the rule file, and `check-eslint-baseline.mjs` prints it while exiting 0, so the gate passes and the custom rule's source simply isn't type-aware-linted. Minor, and unrelated to this change.

Nothing in this PR depended on that claim — the globalSetup fix and its verification stand on their own. Correcting it here so the merged description isn't a source of a wrong fact later.

PR #1278, opened 2026-07-20
fix: the folder count above the Inbox list matches the rows in it

srobroek:
## Summary

- The folder/master count in the Inbox top bar ignored the active filter, so it
  disagreed with the stats strip sitting next to it and with the list below it
  whenever a lane, kind or search filter was applied.
- It also skipped scanned-but-unclassified folders, which the stats strip had
  started counting — so such a folder was counted by one surface and not the
  other, visible with no filter active at all.
- Both now count the same filtered rows the list renders.

## Spec Context

Salvaged from #1264, which was closed as superseded by `de8e5492` (two lanes
picked up T013's UI half in parallel). This is the part of that PR not present
in `de8e5492` and not owned by another task.

Two independent drifts, same SC-004 defect — *"the Inbox summary counts equal
the number of rows the list shows"*:

1. T022 moved `derivedStats` onto `filteredItems` for SC-004, but
   `folderCount`/`masterCount` (`InboxPage.tsx:1007`) stayed on the unfiltered
   `items`.
2. T013/CHK010 gave `deriveInboxStats` its `sourceGroups` argument; the header
   was never taught the same rule.

Fixed at the one site rather than teaching the header the source-group rule
separately, which would leave the two derivations free to drift a third time.

**Not in scope:** wiring selection so choosing a source-group row calls
`inbox.classify.sourceGroup`. `c94fda75` explicitly assigns that to T017, which
owns the CHK011 selection rules and FR-017 row replacement.

**Verification.** Desktop inbox suite 222/222, tsc clean, eslint/i18n/token
checks clean. Two-direction control: reverting the header to unfiltered `items`
fails exactly the two new tests and leaves the pre-existing strip assertion
passing.

One caveat worth recording — the controls had to be re-run with
`--testTimeout=120000`. The box was at **load average 71** from concurrent
lanes, and at the default 5s every test in the file failed, including the
pre-existing one on the pristine tip. The first control run was invalid for
that reason, not because of the change.


srobroek:
The macOS `Integration (Layer 1)` red here is **not from this PR**. `Frontend unit tests` dies with `mapfile: command not found` (exit 127) before any test runs — `mapfile` is a bash 4 builtin and the macOS runner ships bash 3.2. It came in with #1212 and is currently red on main. Fix is in #1283; once that lands this leg should go green without any change here.

srobroek:
The `mapfile` fix landed — #1282 merged at 11:29Z and `main` now has **0** code occurrences of `mapfile` in `ci.yml` (only two comment lines explaining why not to use it), with the `while IFS= read -r f` loop in place at `ci.yml:429`.

This PR's head (`fe30f017`) predates that merge, so the `Integration (Layer 1) — macos-latest` red above is stale: it's still the old `Frontend unit tests` / exit 127 failure, not anything in this diff. A rebase onto main (or a re-run once rebased) should clear it with no change here.

PR #1271, opened 2026-07-20
docs: correct the E2E Channel claim, refresh the coverage matrix, record journey timing

srobroek:
Fixes #1234. Comment/doc only — no behaviour change (Rust doc comments + coverage-matrix.md).

Closes #1234
Closes #1235
Closes #1228

## #1234 — false "cannot construct a Channel" claim

A Tauri Channel is an id from `__TAURI_INTERNALS__.transformCallback`, serialised as the plain string `__CHANNEL__:${id}` (`IPC_PAYLOAD_PREFIX`, tauri 2.11 `src/ipc/channel.rs:28`); the Rust side deserialises it from exactly that string via `JavaScriptChannelId`'s `FromStr`. The harness already runs arbitrary JS in the real webview, so nothing structurally prevents building one — it just means reaching into Tauri-internal plumbing from a test.

Corrected the four sites that overstated this as impossibility (`journeys.rs`, `plan_apply.rs`, `plan_apply/apply.rs`, coverage-matrix.md) to match the three sites that already framed it correctly as a deliberate choice. `archive_journeys.rs:11` was checked and left alone — it already said "no Channel required", never claimed impossibility. `plans.apply.direct` is unaffected; it remains independently justified for fire-and-poll callers.

Verification: `git grep -F "cannot construct"` → no hits remain anywhere in the repo.

## #1235 — coverage matrix refresh (specs 050-058)

The matrix referenced no spec above 049 and had one stale "uncovered" entry.

- **Stale entry corrected**: Data Sources Rescan and Disable/Enable/Delete have had vitest component coverage since 2026-07-04 (`DataSources.rescan.test.tsx` `7e19d7bb`, `DataSources.disable-delete.test.tsx` `7219fb57`) — predating the note that called them uncovered. Remap and Reveal remain genuinely untested at every layer (no test references them, no Mock-Playwright spec for Data Sources exists at all).
- **Spec 051 added** as the one shipped-and-untested surface in the 050-058 range: US1/US2/US3/US4/US5/US6/US7/US10 are merged to `main` per `SPEC_STATUS.md:93` with zero matrix reference at any layer.
- **"Not implemented" section added** for 050/056/058, pointing at `SPEC_STATUS.md` so a future audit doesn't re-derive shipped-vs-specified. (053 has a `specs/` directory but no `SPEC_STATUS.md` row, so it's called out as out-of-scope rather than lumped in as "not implemented".)

## #1228 — journey-suite timing record

[...]

PR #1282, opened 2026-07-20
fix: restore the frontend test suite on macOS CI

srobroek:
## Summary

- The frontend unit-test step has been failing on every macOS CI run since
  #1212, before any test executes, so `Integration (Layer 1) — macos-latest`
  reds any PR that touches frontend files.
- Cause: #1212 used `mapfile`, a bash 4 builtin. The macOS runner's
  `/bin/bash` is 3.2 (Apple ships no GPLv3 bash), so it exits 127 and takes the
  step with it.
- Replaced with a portable `while IFS= read -r` loop.

## Why it reads as a test failure

The step is named "Frontend unit tests" and the log tail is
`grep: stdout: Broken pipe` from the orphaned process substitution — nothing
in it names bash or `mapfile` until you scroll to `exit code 127`. Anyone
triaging a red macOS job right now is likely to blame their own tests.

## Verification

Both branches exercised locally under `set -euo pipefail`:

- a mixed changed-file list (desktop TS, a Rust file, a deleted TS file)
  selects **exactly** the two existing `apps/desktop/src` TS files
- a list with no frontend files leaves the array empty and falls through to the
  full-suite branch without tripping `set -u` — the empty-array case that bites
  bash 3.2

Failure mode reproduced with `enable -n mapfile`, which prints the same
`mapfile: command not found` as the macOS runner log.

The existence filter is folded into the read loop rather than kept as a second
pass over an intermediate array, so there is no longer a way for the two to
disagree about which files were collected.

## Spec Context

Found while shepherding #1278, whose macOS job failed this way with no test of
its own at fault. Unrelated to that change; targets `main` because it blocks
every lane.

Refs #1212

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -402,15 +402,23 @@ jobs:
           # Changed files under this package, rebased to package-relative paths
           # and filtered to ones that exist (a deleted file has no dependents to
           # resolve and makes `vitest related` error).
-          mapfile -t files < <(
+          #
+          # Read with `while read`, NOT `mapfile`: the macOS runner's /bin/bash
+          # is 3.2 (Apple ships no GPLv3 bash), where `mapfile` does not exist.
+          # It exits 127 and takes the whole step with it, so the frontend suite
+          # never runs on macOS at all. The existence filter is folded into the
+          # same loop rather than kept as a second pass over an intermediate
+          # array — one loop cannot disagree with itself about which files it
+          # collected.
+          existing=()
+          while IFS= read -r f; do
+            [ -n "$f" ] || continue
+            [ -f "$f" ] && existing+=("$f")
+          done < <(
             printf '%s' "$CHANGED_FILES" | tr ',' '\n' \
               | sed -n 's#^apps/desktop/##p' \
               | grep -E '^src/.*\.(ts|tsx)$' || true
           )
-          existing=()
-          for f in "${files[@]}"; do
-            [ -f "$f" ] && existing+=("$f")
-          done
 
           if [ ${#existing[@]} -eq 0 ]; then
             echo "No existing apps/desktop/src TS files changed -> full suite (conservative)."


srobroek:
Confirming this from a third lane, and closing my duplicate (#1283) in favour of it — yours was opened first and the "one loop cannot disagree with itself" comment is the better one to keep.

Carrying over the evidence I gathered, so it lives with the surviving PR:

**Provenance.** `mapfile` entered in #1212 (`247a047a`, "ci: only run the frontend suite for changes that can affect it"), merged earlier today.

**Exact symptom**, for whoever greps this later — `Integration (Layer 1) — macos-latest`, step `Frontend unit tests`:

```
/Users/runner/work/_temp/….sh: line 31: mapfile: command not found
grep: stdout: Broken pipe
##[error]Process completed with exit code 127
```

**Two-directional verification.** Emulated bash 3.2 with `enable -n mapfile`:

| | old (`mapfile`) | new (`while read`) |
|---|---|---|
| bash 3.2 | `mapfile: command not found`, aborts | selects the changed files that exist |

And the three input paths behave identically before/after, so #1212's conservative fallback is preserved:

```
crates/foo/src/lib.rs      -> 0 files -> full-suite fallback
(empty)                    -> 0 files -> full-suite fallback
apps/desktop/src/app/*.tsx -> 1 file  -> scoped
```

**Confirmed live impact:** #1278's macOS leg. I'd previously written that #1271 and #1265 were also affected — that was wrong, and I checked afterwards: #1271's macOS leg fails on `Generated contract/binding drift` (same as its ubuntu leg) and #1265 has no macOS failure. Only the `Frontend unit tests` step is hit by this.

Also: `mapfile` was the only bash 4+ construct in `.github/workflows/` — I grepped `readarray`, `declare -A`, and `${var^^}`/`${var,,}` too.

One note: this PR is currently **BEHIND** main.

PR #1289, opened 2026-07-20
refactor: setup wizard scan list uses the shared table component

### LINKED ISSUE #752: Wizard scan step hand-rolls a parallel triage UI contra FR-003's reuse mandate (spec 038)
## What
FR-003: "MUST reuse the existing Inbox triage components (no parallel UI)." The implementation reuses only data types (`InboxItemSummary`, `InboxClassifyResponse`) and the generic `Pill` atom — not `InboxList`/`InboxDetail`/`InboxStatsSummary`. `SourceSummary` (`apps/desktop/src/features/setup/steps/StepScan.tsx:110-312`) hand-rolls its own accordion + `<table>` markup backed by 24 bespoke `alm-setup-scan__*` CSS classes (`wizard-steps.css:922+`). The merge commit (`e43b0db1`) conflated data-type reuse (FR-001) with component reuse (FR-003).

## Why it matters
A genuine parallel UI — exactly what FR-003 rules out — shipped silently instead of routing the deviation through iterate. Either retrofit onto the shared `Table`/`InboxList` machinery or record an explicit accepted-deviation note in the spec.

## Classification
UI / process decision.

Found by the 2026-07-14 full-spec verification sweep (run-052fix); full report in the run store.

srobroek:
Closes #752

## What

Issue #752 found that the wizard scan step hand-rolled a parallel triage UI —
its own `<table>` markup backed by 24 bespoke `pv-setup-scan__*` classes —
against spec 038 FR-003's "MUST reuse the existing Inbox triage components (no
parallel UI)".

This takes both paths the issue offers, because investigation showed each one
covers a different half of the problem.

**Retrofit (done).** The detection list now renders through the shared `Table`
primitive — the same construct `InboxList` itself uses — in `virtualized` mode
for its scroll container and sticky header. Six clone classes are deleted
(`__table`, `__thead-row`, `__cell`, `__cell--path`, `__tbody-row`,
`__path-cell-inner`); `__table-wrap` survives only to bound the scroll height.

**Accepted-deviation note (done).** `InboxList`/`InboxDetail`/
`InboxStatsSummary` still are not reused, and that is now documented in
`specs/038-wizard-scan-step/spec.md` instead of shipping silently — which was
the issue's actual complaint.

## Why InboxList could not be reused

Not a styling difference — a data-model one.

`InboxList` is typed on `InboxListItem`, which extends `InboxItemSummary` with
~15 enrichment fields the wizard's scan never returns, and it reads exactly
those missing fields for its two headline columns: the detection column reads
`rootAbsolutePath` (`InboxList.tsx:192`), the Type column reads `frameType` /
`groupFrameType` (`InboxList.tsx:110-114`). Feeding it wizard data means
fabricating ~15 fields and still rendering both headline columns blank.

Semantically, `InboxList` models single-type post-materialization items
(`bindings/index.ts:5020-5025`). The wizard runs pre-materialization over
folders that are legitimately multi-type and shows a per-kind count breakdown
("16 light, 2 dark") from a separate `classifications` map. `InboxList` has no
column that can express a breakdown, and it mandates `selectedId`/`onSelect`
while the wizard summary is read-only.

The shared collapsible primitive `ui/Section.tsx` was also evaluated rather than
assumed absent. It renders `{open && children}` with no always-visible body
slot, but the card must keep the spec-048 `RootDetectionConfig` control and the
phase messages visible *while collapsed*. Adopting it would hide or reorder a
documented control — a behaviour change needing its own iterate cycle, not a
reuse refactor.

## Evidence

- `css-dup-sniff`: 1654 → 1648 rules; four-copy utility groups 68 → 67;
  `pv-setup-scan` findings 7 → 6. Classes 24 → 18.
- CSS definitions and TSX usage cross-checked: exact match, no dead rules.
- All 29 `StepScan.test.tsx` tests pass **with no test edits**.
- `pnpm lint` pass, `pnpm typecheck` pass.
- `pnpm test`: 1842/1845. The 3 failures (`GuidedOverlay.test.tsx`,
  `devSurface.release.test.ts`) reproduce identically on clean `origin/main`
  (3 failed | 12 passed both sides) — pre-existing, unrelated to these files.

## Needs real-UI visual validation

`pv-table`'s metrics differ from the deleted bespoke styling (`--pv-text-xs` /
`--pv-text-secondary`, tighter cell padding), so the step's table **will look
different**. Dropping `__path-cell-inner` also removes an explicit flex gap
between the Master pill and its label. Unit tests cannot catch either. Routing
to the real-UI lane per the issue's UI classification.

### DIFF
--- a/apps/desktop/src/features/calibration/MastersTable.tsx
+++ b/apps/desktop/src/features/calibration/MastersTable.tsx
@@ -446,7 +446,7 @@ export function MastersTable({
       _selected: selected === master.id,
       _indent: indentPx || undefined,
       master: (
-        <span className="pv-calib-cell__master">
+        <span className="pv-table__cell-inline">
           <Pill variant={kindVariant(kindStr)}>{kindStr.toUpperCase()}</Pill>
           <span className="pv-calib-cell__master-label">
             {masterLabel(master)}
--- a/apps/desktop/src/features/setup/steps/StepScan.tsx
+++ b/apps/desktop/src/features/setup/steps/StepScan.tsx
@@ -192,12 +192,12 @@ function SourceSummary({ state }: SourceSummaryProps) {
     return {
       _testid: `scan-item-${item.inboxItemId}`,
       folder: (
-        <>
+        <span className="pv-table__cell-inline">
           {item.isMaster && <Pill variant="info">{m.setup_scan_master()}</Pill>}
           {/* The root/top-level row of an expanded folder carries an empty
               relativePath; label it instead of leaving it blank (issue #513). */}
           {item.relativePath || m.setup_scan_root_label()}
-        </>
+        </span>
       ),
       files: item.fileCount,
       format: (item.format ?? item.lane).toUpperCase(),
--- a/apps/desktop/src/features/setup/steps/StepScan.tsx
+++ b/apps/desktop/src/features/setup/steps/StepScan.tsx
@@ -10,6 +10,8 @@
 import { useEffect, useRef, useState } from 'react';
 import { Pill } from '@/ui/Pill';
 import type { PillVariant } from '@/ui/Pill';
+import { Table } from '@/ui';
+import type { TableColumn, TableRow } from '@/ui';
 import { m } from '@/lib/i18n';
 import { commands } from '@/bindings/index';
 import { unwrap } from '@/api/ipc';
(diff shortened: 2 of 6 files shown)


srobroek:
Owner overruled the cosmetic drop noted in the FR-003 deviation: the "Master" pill / label gap is restored.

It lives in the shared layer, not a feature-local clone. `.pv-table__cell-inline` sits with the other `.pv-table` rules in `primitives.css` and uses `var(--pv-sp-2)`. `.pv-calib-cell__master` declared the same four properties, so it is retired onto the shared class instead of being left as a second copy; `MastersTable` now uses the shared class too.

`css-dup-sniff` total top-level rules: 1648 before, 1648 after (one rule added, one removed). Clone/utility group counts unchanged.

The 29 StepScan tests pass with no edits. Full desktop suite: 202 files / 1861 tests green; `just lint` and `just typecheck` exit 0 (the 72 biome warnings are pre-existing — identical count with the change stashed).

PR #1306, opened 2026-07-20
test: reuse a migration snapshot across test processes and close migration-test gaps

### LINKED ISSUE #1230: The Rust test suite spends 74% of its CPU re-running the same 66 migrations
## Summary

The Rust test suite spends **74% of its CPU** re-running the same 66 migrations. A cross-process schema-snapshot cache gives a **3.4x wall-clock speedup with zero test-file changes**.

## Cause

`crates/persistence/db/src/lib.rs:127` runs `sqlx::migrate!("./migrations")` per test. nextest runs **one process per test**, so each of ~1069 DB-touching tests pays the full 66-migration chain.

## Measurements (12-core, warm)

| | wall | CPU | tests |
|---|---|---|---|
| baseline | 43.6s | 326.4s | 2503 |
| with on-disk snapshot cache | 9.9s | 83.8s | 2503 (all passing) |

Per-test cost breakdown:

- connect: 0.4ms
- \+ migrate: 278.8ms
- \+ snapshot replay: 15.0ms
- \+ file-template copy: 3.7ms

## Critical detail: the cache must be on disk

An in-process `OnceLock` cache does **not** work — measured 29.4s vs 28.0s baseline, i.e. slightly *worse*. Each test is its own process, so there is nothing to memoise in-process.

## Requirements for a correct implementation

1. **Key the cache on a hash of the embedded migrator**, not a fixed filename. `sqlx::migrate!().iter()` exposes version + checksum. This repo has a recorded history of a stale `sqlx::migrate!` embed, so a filename key will eventually serve a wrong schema silently.
2. **Dump seeded DATA, not just schema.** A schema-only cut failed exactly four tests that rely on migration-seeded rows:
   - `repositories::calibration_tolerances::tests::get_returns_seeded_defaults`
   - `repositories::equipment::tests::filter_crud`
   - `repositories::q_desktop::tests::resolver_settings_seeded_by_migration_0031`
   - `repositories::q_targets_mgmt::tests::resolver_settings_seeded_row_is_readable`
3. **Provide `migrate_uncached()`** so the migration-chain tests still exercise the real chain: `crates/persistence/db/tests/migration_0045.rs`, `migration_0060.rs`, `migration_0070.rs`. Without an opt-out the cache silently weakens migration coverage — trading one gap for another.

## On source-of-truth concerns

Migrations remain the sole source of truth for the schema; the cache only memoises their output, keyed on their hash.

A checked-in `schema.sql` was considered and **rejected**: it saves only ~279ms per CI job over the cache while introducing a second source of truth that can drift, plus frequent merge conflicts across parallel branches.

### LINKED ISSUE #1231: Migration chain: 15 table-rebuilds, one populated-DB test; busy_timeout never set in production
## Summary

66 migrations, 15 table-rebuilds, and exactly **one** genuine forward-migration-on-populated-DB test.

## Evidence

The one real test: `crates/persistence/db/tests/migration_0070.rs:128-161`.

The two that look like coverage but are not:

- `crates/persistence/db/tests/migration_0060.rs:6-8` — admits in its own comment that it replays after the fact.
- `crates/persistence/db/tests/migration_0045.rs:68-76` — runs a **hand-copied duplicate** of the backfill SQL. It would still pass if migration 0045 were deleted outright.

## Highest cumulative risk

`plans` / `plan_items` have been rebuilt **eight** times: migrations 0019, 0029, 0040, 0045, 0053, 0054, 0071, 0073. One botched column list in a table rebuild silently drops plan history — the durable audit record the constitution makes load-bearing (Principle V), and it fails silently because the new table is well-formed, just emptier.

Also high-risk and untested on populated data:

- `0049_inbox_single_type.sql` — 22KB, 7 drop/rename operations
- `0050_session_lifecycle_drop.sql` — 9 `DROP COLUMN`

## Separate finding: `busy_timeout` is never set in production

`crates/persistence/db/src/lib.rs:91-96` sets only `journal_mode` and `max_connections`. The sole workspace occurrence of `busy_timeout` is inside a test's own connect options.

WAL reader/writer contention **is** well covered (`crates/persistence/db/tests/wal_journal_mode.rs`). Two-**writer** contention and `SQLITE_BUSY` handling are not covered at all. Given the single-instance guard is itself untested and E2E-disabled (see the `src-tauri/src/lib.rs` coverage issue), two writers is not a hypothetical configuration.

## Suggested fix

1. A reusable populated-DB migration harness: seed representative rows at version N-1, run the migration, assert row counts and column values survive. Apply it first to the eight `plans`/`plan_items` rebuilds.
2. Replace the hand-copied SQL in `migration_0045.rs:68-76` with an actual invocation of the migration.
3. Set `busy_timeout` in production connect options and add a two-writer contention test.

srobroek:
Closes #1230
Closes #1231

## #1230 — cross-process schema snapshot

`crates/persistence/db/src/schema_cache.rs`. The first process runs the real
chain once and dumps the result as SQL; every later process replays that dump.
On-disk because nextest gives each test its own process — there is nothing for
an in-process memo to share.

Follows the issue's three non-negotiables:

1. **Keyed on the embedded migrator**, not a filename — SHA-256 over every
   migration's `version` + `checksum`. A stale `sqlx::migrate!` embed misses
   the cache instead of silently serving a mismatched schema.
   `key_tracks_every_version_and_checksum` covers add/drop/change of each.
2. **Seeded rows are dumped, not just schema.**
   `snapshot_reproduces_the_real_chain` compares a cached database against an
   uncached one on schema *and every row value*, and asserts at least one table
   held rows so it cannot pass vacuously.
3. **`migrate_uncached()`** added. `migration_0045.rs` and `migration_0060.rs`
   use it; `migration_0070.rs` never called `migrate()` to begin with.

Safety: only an empty database accepts a snapshot (a populated one always takes
the real chain); the replay is one transaction, so any failure rolls back and
falls through to the real chain. Enabled only under `debug_assertions`.

One design note worth review: the snapshot is **SQL text, not an ATTACHable
database file**. sqlx opens `:memory:` databases with `SQLITE_OPEN_MEMORY`, and
that flag applies to ATTACH too — an attached file is silently swapped for an
empty in-memory database, which handed every in-memory test a blank schema.
Row literals come from SQLite's own `quote()`, so no value formatting happens
in Rust.

### Measured

`persistence_db` only (320 tests), not the full workspace — a full-suite run
was not possible on this box (swap exhausted, other lanes building), so the
issue's 3.4x full-suite figure is **not** reproduced here.

| repeat | cache on | cache off (`ALM_SCHEMA_CACHE=0`) |
|---|---|---|
| 2 | 6.77s | 28.93s |
| 3 | 5.12s | 22.17s |
| 4 | 5.49s | 28.27s |
| 5 | 6.04s | 26.51s |

~4.4x on this crate, stable across four repeats. Absolute numbers are inflated
by load average 16-27 from other lanes; the ratio is the meaningful figure.
`persistence_db` is DB-heavy, so its ratio is an upper bound relative to the
workspace. **A full-suite number still wants measuring on a quiet box.**

## #1231

**(a) `busy_timeout` — not a live defect.** sqlx-sqlite 0.9 already defaults it
to 5s (`options/mod.rs:203`, applied at `establish.rs:184`) and
`SqliteConnectOptions::from_str` builds from `Self::default()`. Verified by
removing the explicit call: all three new tests still pass, including the
`PRAGMA busy_timeout` readback of 5000ms. So this states the value rather than
inheriting it, and covers the behaviour that genuinely had no coverage —
writer/writer contention. `tests/two_writer_contention.rs`;
`zero_busy_timeout_writer_gets_sqlite_busy` is the control arm proving the race
actually contends.

**(b) `migration_0045.rs` backfill — now real.** It ran a hand-copied duplicate
of the migration's own UPDATE and would have passed if 0045 were deleted. It
now migrates to 0044, seeds rows, then applies 0045. Proven by deleting the
migration's backfill statement: fails `left: "unorganized", right: "organized"`.

**(c) populated-DB harness for the eight `plans`/`plan_items` rebuilds** — see
the comment below for scope and a concrete plan.

## Gates

`cargo clippy -p persistence_db --all-targets -- -D warnings` clean;
`just db-boundary` clean; `cargo fmt --all`; `pre-commit` clean.
`cargo nextest run -p persistence_db` 320/320. Consumers `-p audit -p
targeting_resolver` 97/97.


srobroek:
## #1231(c) — harness landed, and it found a live data-loss defect

`tests/common/mod.rs` (`migrate_through` / `migrated_to`) plus
`tests/migration_plan_rebuilds.rs`, applied to 0071 as the template for the
eight `plans`/`plan_items` rebuilds.

Pointed at the first rebuild, it found **#1307**: every `plans` rebuild opens
with `PRAGMA foreign_keys = OFF` to make its `DROP TABLE plans` safe, but that
pragma is a **no-op inside a transaction** and sqlx runs every migration in one
(`sqlx-sqlite-0.9.0/src/migrate.rs:173`). Foreign keys stay on, the DROP
cascades through `plan_items.plan_id ... ON DELETE CASCADE`, and every plan
item is deleted. The `plans` row survives intact, so the loss is invisible from
the plans list.

A/B on identical SQL and identical seed data:

| how 0071 is applied | `plan_items` after |
|---|---|
| through the migrator (in a transaction) | **0** |
| as a raw script (no transaction) | 1 |

Seven migrations share the shape: 0019, 0029, 0040, 0053, 0054, 0071, 0073.

`migration_0071_preserves_plan_items` asserts the correct behaviour and is
`#[ignore]`d against #1307. **Not fixed here** — the migration files cannot be
edited (their checksums are recorded in every deployed database), so the fix is
a forward migration plus a rule for future rebuilds, and already-deleted rows
are not recoverable. That needs its own change and a product decision.

### Remaining seven — concrete plan

For each of 0019, 0029, 0040, 0053, 0054, 0073, in `migration_plan_rebuilds.rs`:

1. `migrated_to(N-1)`, seed a `plans` row plus a `plan_items` row in that
   version's column shape (each differs; read the preceding rebuild's
   `CREATE TABLE` for the shape).
2. Snapshot the row via an explicit column list — not `SELECT *` — so a shifted
   column fails as loudly as a dropped row.
3. `migrate_through(N)`, then assert the snapshot is unchanged and the item
   count holds.

Each is ~40 lines and independent. They should be written **after** #1307 is
fixed, otherwise all six land `#[ignore]`d for the same reason and prove
nothing new.

Issue #1203, opened 2026-07-20
Real-UI E2E shards never run on main, so main-side regressions are invisible and tax every PR


srobroek:
## Summary

The `Real-UI E2E shard` jobs run **only on pull requests**. A push/merge to `main` runs `UI E2E (mock-mode Playwright)` and reports success without ever executing them.

The consequence: a real-UI regression can sit on `main` reporting green, while every PR cut from `main` goes red on a test its author never touched. Nobody owns the failure, so nobody fixes it, and it silently taxes the whole queue.

## Evidence

The last `main` run at the time of writing, `29715065065` (sha `15edea17`), is **green**. Its only E2E job:

```
success  UI E2E (mock-mode Playwright)
```

No `Real-UI E2E shard — *` jobs in the run at all.

Meanwhile four unrelated open PRs fail the *same* real-UI test, `inbox_ui_unsplit_unclassified_folder_badge_is_not_classified`:

| PR | what it changes |
|---|---|
| #1188 | Inbox detail column wrapper |
| #1180 | **docs only** |
| #1169 | Rust shell bootstrap split |
| #1196 | Lock tabstop / cleanup table |

#1180 changes documentation and nothing else, so the failure cannot originate in any of these branches.

Failure mode:

```
test inbox_ui_unsplit_unclassified_folder_badge_is_not_classified ... FAILED
Error: the post-classify Inbox list never settled within 20s
       (no materialized 'needs review' sub-item row appeared); Type cells were []
```

This is the #854 sub-item materialization family (see also #1021).

## Why this is the bug, not the test

There are two candidate causes — a genuine regression on `main`, or a racy test — and **the current CI shape makes them indistinguishable**, because nothing ever runs these shards against `main` directly. That ambiguity is itself the defect: a whole class of regression has no signal on the branch it lands on.

Note this is the same shape as the previously-fixed docs-push vacuous-green problem (#1075/#1079): a job-level condition producing a green run that proves less than it appears to.

## Suggested fix

1. Run the `Real-UI E2E shard` matrix on `main` pushes (or on a schedule) so a regression is attributed to the commit that caused it, rather than to the next unlucky PR author.
2. Then resolve `inbox_ui_unsplit_unclassified_folder_badge_is_not_classified` on the evidence that produces — genuine regression vs. flake.
3. Consider whether these shards should be required checks. They currently are not, so `--auto` merges straight past them; that is either a deliberate tradeoff worth documenting or a gap worth closing.

## Impact

[...]

PR #1309, opened 2026-07-20
ci: let unit, integration and mock-mode actually finish on main

srobroek:
Closes the remaining half of #1203. #1268 fixed `e2e.yml`; `ci.yml` had the
identical bug and still has it on `main` as of this PR.

## The bug

```yaml
group: ci-${{ github.workflow }}-${{ github.ref }}
cancel-in-progress: true
```

`github.ref` is `refs/pull/N/merge` on a PR — one group per PR, so cancelling a
superseded run is correct and saves runner minutes.

On a push to main it is `refs/heads/main` for *every* merge. All main runs share
one group, so each merge cancelled the previous merge's run. Under campaign
merge rates nothing survived to report.

## Evidence on current main

```
6e624adc  15:29  fix(ci): frontend suite on macOS   -> cancelled,cancelled,...
67f48ee9  15:24  ci: rebalance E2E shards           -> cancelled,cancelled,...
```

`ci.yml` owns unit, integration and mock-mode UI, so main's gate for all three
was empty. It is worse than a visible outage because it is merge-rate
dependent: when merges space out a run completes and main looks green. The gap
opens only during the highest-velocity periods — when verification matters
most — which makes it easy to dismiss as flake.

## The fix

```yaml
cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

Same expression `e2e.yml` already carries. PRs keep cancelling superseded runs;
main runs always finish.

**Cost is bounded.** With `false`, GitHub keeps at most the in-flight run plus
one queued run per group and discards intermediate pending runs, so this is
~2 runs per group, not a serialised backlog — and the queued run always
coalesces to the newest commit.

## Scope

One line plus its rationale comment. No job, trigger or step changes. Verified
the workflow still parses and all three jobs are intact. Only #1292 also touches
`ci.yml`; it changes clippy wiring and does not go near the concurrency key.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,22 @@ on:
 
 concurrency:
   group: ci-${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  # Cancel superseded runs on branches, but never on `main` (#1203).
+  #
+  # Every main push shares one concurrency group, because `github.ref` is
+  # `refs/heads/main` for all of them. A blanket `true` therefore made each
+  # merge cancel the previous merge's run, and during a merge campaign no main
+  # run survived to report: `6e624adc` and `67f48ee9` landed five minutes apart
+  # and both concluded `cancelled,cancelled`.
+  #
+  # That silently emptied main's gate for unit, integration and mock-mode UI —
+  # the suites this workflow owns. It is worse than a plain outage because it is
+  # merge-rate dependent: when merges space out a run completes and main looks
+  # green, so the gap only opens during periods of highest change velocity,
+  # which is exactly when verification matters most.
+  #
+  # e2e.yml took the same fix in #1268; this is the other half.
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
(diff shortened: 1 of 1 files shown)


srobroek:
Correcting an overclaim in this PR's description, now that post-merge behaviour
has been observed.

The body says main runs "always finish." That is too strong. With
`cancel-in-progress: false`, GitHub still evicts **intermediate queued** runs —
only the in-flight run plus the newest queued one survive. Observed on main
within four minutes of this merging:

```
13:31:15  f3668b86  pending      <- survives
13:31:08  89025366  cancelled    <- evicted while queued
13:31:00  b37b072e  cancelled    <- evicted while queued
13:30:53  30e29bd1  cancelled    <- evicted while queued
```

All four landed after the fix. The accurate claim:

- **Before:** during a merge campaign, *no* main run completed — each merge
  killed the previous one mid-flight.
- **After:** main always has a recent run that completes. Individual
  intermediate commits can still go unverified when merges land seconds apart.

That is the bounded-cost behaviour the body already described (~2 runs per
group); the "always finish" phrasing just contradicted it. Flagging so nobody
sees `cancelled` on main later and re-opens #1203 as a regression — that
appearance is expected.

Issue #1204, opened 2026-07-20
Real-UI E2E on Windows: webview session comes up broken — script evaluation fails for the whole bridge wait, passes on retry


srobroek:
Recurrence of #1019 (closed COMPLETED 2026-07-19) with a **different failure signature**, so filing fresh rather than reopening.

## Evidence

Run [29715450194](https://github.com/platevault/platevault/actions/runs/29715450194), job `88270065761`, shard `Real-UI E2E shard — windows-latest 2/2`, on PR #1117 head `3406f70b`.

```
2026-07-20T04:22:21Z  test inbox_ui_missing_path_attribute_banner_blocks_confirm ... FAILED
                      finished in 155.97s
Error: window.__ALM_E2E__ bridge never became ready
```

**It passed on nextest retry** — it is one of the `2 flaky` in the run summary:

```
Summary [443.197s] 10 tests run: 9 passed (2 flaky), 1 failed, 14 skipped
```

Test: `crates/e2e-tests/tests/inbox_ui_journeys.rs`.

## Why this is not #1019

#1019 was the metadata-query race (`inbox.item.metadata` racing per-file extraction from the same selection's `inbox.classify`), worked around with a refresh retry. This failure never reaches that logic — the **E2E bridge itself never initialises**, so the harness cannot drive the app at all. Different layer, different fix.

Corroborating: #1202 baselined this same journey against `main` (`15edea17`) and it **PASSED in 4.4s**. Here it took 155.97s before giving up. That is a harness/startup problem, not a product assertion failure.

## Probable contributing factors (hypothesis, not established)

1. The same job logged GitHub partner-runner infra flakiness twice — `install-action: installation failed due to bash startup failure`, [actions/partner-runner-images#169](https://github.com/actions/partner-runner-images/issues/169). #1019 noted the same warning.
2. The same shard contained `inbox_ui_unsplit_unclassified_folder_badge_is_not_classified`, which is **deterministically red** (#1202) and burned ~144s x3 retries. A shard already saturated by a hanging journey may be starving the webdriver session that this test needs. Worth testing whether these flakes disappear once #1202 is fixed.
3. Recorded prior art: Windows E2E shards share one WebView2 localStorage and are not isolated per shard.

## Suggested action

- Harden bridge-readiness waiting (explicit poll with a clear diagnostic on timeout) rather than a fixed budget, so a slow runner degrades gracefully instead of failing the shard.
[...]

PR #1315, opened 2026-07-20
test(e2e): isolate concurrent instances with a data dir the OS cannot ignore

srobroek:
Refs #1204, #1019. The real isolation fix that #1295 stands in for.

## The gap #1295 left open

#1295 found the root cause and stopped the bleeding by serialising Windows E2E journeys (`threads-required = 2`). I said there that it **serialises rather than isolates**, and that real isolation needs the app to accept a data-directory override, since no env var can redirect a Windows Known Folder. This is that override.

## Why the existing isolation was fake on Windows

The harness gives each concurrent instance an isolated app-data root via the platform location vars — `XDG_*` on Linux, `HOME` on macOS, `APPDATA`/`LOCALAPPDATA` on Windows.

Two of those work. Tauri resolves app dirs through `dirs`, which on Windows calls `SHGetKnownFolderPath` (`dirs-sys-0.5.0/src/lib.rs:176`). The Known Folder API reads the shell profile and ignores both variables outright. So every concurrent Windows instance shared one real root:

| Collision | Consequence |
|---|---|
| `simbad-cache.redb` | `Database already open. Cannot acquire lock` → that instance silently degrades to an in-memory cache |
| WebView2 user-data folder | loser cannot create its webview (`WindowsError(0x80070057)`) → no window → `tauri-plugin-webdriver` never serves → **`bridge never became ready`** |

## The fix

**`ALM_DATA_DIR`** — an override the *app* honours, so isolation no longer depends on the OS agreeing to be redirected. Used verbatim as the app-data root (SQLite default, resolve cache, logs) on every platform. Unset — real users, every non-E2E build — nothing changes.

**The webview folder needs a second lever.** A config-declared window's `data_directory` must be *relative*: `WebviewBuilder::from_config` rejects absolute paths and joins relative ones onto `dirs::data_local_dir()/<label>` (`tauri-2.11.5/src/webview/mod.rs:392-425`) — the very Known Folder we cannot redirect. So the fix makes the leaf name unique per instance rather than relocating it. **Isolating identity ends the collision; isolating location is not on offer here** — worth knowing if this ever needs revisiting.

Windows only. Linux and macOS already isolate correctly, and setting `data_directory` there would *move* WebKitGTK/WKWebView storage out of the isolated root it lives in today — a regression bought for nothing.

## A silent no-op this also fixes

`reset_webview_storage()` deleted `<isolated LOCALAPPDATA>/<identifier>/EBWebView` — a path that never existed, because the app had been writing to the real profile all along. Every Windows reset deleted nothing, and journeys shared one localStorage regardless of how carefully each one reset. That is a known separate symptom, now closed by the same change.

## Verification, and its limits

- `cargo test -p desktop_shell` — 77 passed
- `cargo clippy -p desktop_shell -p e2e_tests --all-targets` — clean
- 7 new unit tests on the derivation, including the two properties the fix rests on: different roots must not share a folder, and one root must stay stable across a relaunch (or `PreserveWebviewStorage` stops preserving).

**What I could not verify locally: the Windows behaviour itself.** I have no Windows runner here, so the claim that mutating the context config actually reaches window creation rests on reading `tauri-2.11.5`/`wry-0.55.1` source, not on observing it. The Windows E2E leg on this PR is the first real evidence — treat a green run there as the proof, not this description.

Two review notes:

- The harness reimplements the folder derivation (`crates/e2e-tests` deliberately does not depend on the Tauri app crate). Both copies pin the same literal, so drift fails a test instead of producing another reset that quietly deletes nothing.
- The Windows path is gated with a runtime `cfg!`, not `#[cfg]`, so it stays type-checked on the Linux and macOS legs rather than compiling only on the platform whose failures it exists to prevent.

## Follow-up, not in this PR

If the Windows leg is green here, #1295's `threads-required = 2` can come back out and Windows can return to two concurrent journeys. Leaving that as a separate change so the isolation lands on its own evidence, and the stop-gap stays until it has some.

### DIFF
--- a/apps/desktop/src-tauri/src/data_dir.rs
+++ b/apps/desktop/src-tauri/src/data_dir.rs
@@ -1,0 +1,192 @@
+// Copyright (C) 2024-2026 Sjors Robroek
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! Explicit app-data-root override (`ALM_DATA_DIR`) — issue #1204.
+//!
+//! # Why an env var cannot do this on its own
+//!
+//! The E2E harness gives every concurrent app instance an isolated app-data
+//! root by overriding the platform's location env vars
+//! (`crates/e2e-tests/tests/common/mod.rs::InstanceEnv`): `XDG_DATA_HOME` on
+//! Linux, `HOME` on macOS, `APPDATA`/`LOCALAPPDATA` on Windows.
+//!
+//! That works on Linux and macOS. **It does nothing on Windows.** Tauri
+//! resolves both `app_data_dir()` and the config-driven webview
+//! `data_directory` through the `dirs` crate, which on Windows calls
+//! `SHGetKnownFolderPath(FOLDERID_RoamingAppData)` /
+//! `FOLDERID_LocalAppData` (`dirs-sys-0.5.0/src/lib.rs:176`). The Known
+//! Folder API reads the user's shell profile — it ignores `APPDATA` and
+//! `LOCALAPPDATA` entirely. No env var can redirect it.
+//!
+//! So on Windows, every concurrent instance silently shared one real
+//! app-data root, which collided in two places:
+//!
+//! 1. **The redb resolve cache** (`simbad-cache.redb`) — the second instance
+//!    logs `Database already open. Cannot acquire lock` and degrades to an
+//!    in-memory cache.
+//! 2. **The `WebView2` user-data folder** — `create_environment` passes
+//!    `data_directory` straight to `CreateCoreWebView2EnvironmentWithOptions`
+//!    (`wry-0.55.1/src/webview2/mod.rs:345`), and an empty value means "the
+//!    default folder next to the exe". Two instances of the same exe get the
+//!    same folder, the second fails with
+//!    `failed to create webview: WindowsError(0x80070057)`, and — having no
+//!    window — never brings up its `WebDriver` port. That surfaced four layers
+//!    downstream as `window.__ALM_E2E__ bridge never became ready`, the
+//!    symptom chased since #1019.
+//!
+//! # The override
+//!
+//! [`resolve`] reads `ALM_DATA_DIR`. When set, it is used verbatim as the
+//! app-data root instead of `app.path().app_data_dir()`, on every platform —
+//! one mechanism rather than three platform-specific ones that are only
+//! honoured on two of them.
+//!
+//! When unset (real users, every non-E2E build) nothing changes: callers
+//! fall back to the platform resolver exactly as before.
+//!
+//! # The webview folder is a separate lever
+//!
+//! The webview's user-data folder cannot live under `ALM_DATA_DIR`, because
+//! a config-declared window's `data_directory` **must be relative** —
+//! `WebviewBuilder::from_config` rejects absolute paths outright and joins
+//! relative ones onto `dirs::data_local_dir()/<label>`
+//! (`tauri-2.11.5/src/webview/mod.rs:392-425`), which is the very Known
+//! Folder we cannot redirect.
+//!
+//! What we *can* do is make the leaf name unique per instance, so concurrent
+//! instances stop sharing one folder even though they all sit under the real
+//! `LocalAppData`. [`webview_subdir`] derives that name from the override
+//! path. Isolation of *identity* is what fixes the collision; isolation of
+//! *location* is not available to us here.
+
+use std::path::PathBuf;
+
+/// Name of the app-data-root override variable. See the module docs.
+pub const DATA_DIR_ENV: &str = "ALM_DATA_DIR";
+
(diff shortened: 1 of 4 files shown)


srobroek:
## Correction: the first Windows run does NOT validate this fix

The E2E workflow reports success on this branch. That green is retry-masked, and I am not treating it as evidence.

Both Windows shards passed only via nextest retries, and the TRY 1 failures carry the exact signature this PR is meant to eliminate:

```
ERROR tauri_runtime_wry: failed to create webview: WebView2 error:
      WindowsError(HRESULT(0x80070057), "The parameter is incorrect.")
Error: window.__ALM_E2E__ bridge never became ready within 30s
```

Five `0x80070057` occurrences, three bridge timeouts, `10 tests run: 10 passed (2 flaky)` and `25 tests run: 25 passed (2 flaky)`. Shape unchanged: TRY 1 times out at ~157s, TRY 2 passes in ~9s — the loser of the race waiting out its timeout, then succeeding once alone.

Worth stating plainly, because the run is green and the required checks are green: **`--auto` would have merged this.**

## And I cannot yet tell you why

The `tracing::info!` I added to `instance_context()` produced no line in CI. That is not evidence the code path was skipped: `build_app()` runs *before* `main.rs` installs the tracing subscriber, so anything logged there is discarded. The absence is consistent with both "the override never reached the app" and "it applied and did not help".

Pushed a diagnostic that settles it (`20146309`): print the decision to stderr instead, which the harness's `ProcLog` and #1262's driver-log dump both capture, including whether the Windows-only application actually fired. The next Windows run answers the question.

Two candidate explanations, and I do not want to guess between them without that output:

1. **`ALM_DATA_DIR` is not reaching the app process** — the config mutation never runs, and this PR is a no-op on Windows.
2. **It applied and is insufficient**, or made things worse. Setting `data_directory` splits what was one shared `WebContext` (both windows at `None`) into two per process, since Tauri resolves the relative name per window label. Creating a second `CoreWebView2Environment` where there was one is a plausible new source of `0x80070057` — that HRESULT is also what you get for environments sharing a folder with mismatched options.

**Do not merge this yet.** Holding until the stderr line says which.

The one thing the run does confirm: the failure is still live on a branch with no serialisation override, which is consistent with #1295's `threads-required = 2` remaining necessary. Its stop-gap status is unaffected either way.

srobroek:
## Reworked: WebView2's own override, not Tauri's config (`178ac166`)

The previous approach was wrong, and I had asserted the fact it rested on without checking it.

I said "no env var can redirect it". That is true of the Windows **Known Folder** — and I generalised it to WebView2, having read wry's source (which passes `data_directory` straight through) without reading what the WebView2 loader does with an empty value.

It reads **`WEBVIEW2_USER_DATA_FOLDER`**, which *replaces* the `userDataFolder` argument to `CreateCoreWebView2EnvironmentWithOptions`. Microsoft documents it as the intended lever for exactly this — testing and deployment overrides.

### Why it is strictly better than what was here before

| | config `data_directory` (before) | `WEBVIEW2_USER_DATA_FOLDER` (now) |
|---|---|---|
| Product code | mutates the Tauri context at build | **none** |
| Isolation | unique *name* under the shared real `LocalAppData` | genuine *location*, inside the instance temp root |
| Storage reset | mirrored derivation, pinned across two crates to stop drift | a path the harness itself chose |
| Verified | no — and it did not work | pending this run |

It also cannot be bypassed the way `APPDATA`/`LOCALAPPDATA` were: it is read by the loader inside the app process, not resolved through a Known Folder.

### What stays

`ALM_DATA_DIR` remains, and is still required — it is what stops the concurrent `simbad-cache.redb` collision (`Database already open. Cannot acquire lock`), which no WebView2 variable covers. Product-code delta is now `main.rs` plus the new `data_dir` module; `lib.rs` is a single added line.

### Recorded so it is not retried

`data_dir.rs` documents why the config route cannot work: a config-declared window's `data_directory` **must be relative** — `WebviewBuilder::from_config` rejects absolute paths and joins relative ones onto `dirs::data_local_dir()` (`tauri-2.11.5/src/webview/mod.rs:392-425`), the very Known Folder that started this. Only `WebviewWindowBuilder::data_directory` takes an absolute path, and config windows are built by Tauri inside `.build()`, where there is no builder to reach.

That constraint is also why I did **not** take the other option on the table — clearing the config windows and recreating them in `setup()` with absolute paths. It would have moved window creation into the splash/window-state/min-size ordering that real users exercise on every launch, to fix a test-harness problem, with no way to verify it outside Windows CI. The env var makes that trade unnecessary.

### Still unverified

Same caveat as before, and it is the one that matters: **I have no Windows runner.** The Windows E2E leg on this run is the evidence. Given the last green here was retry-masked, the bar is TRY 1 clean — zero `0x80070057`, zero bridge timeouts, zero flaky — not merely a green check.

Issue #489, opened 2026-07-09
Real-UI E2E: macOS leg blocked — tauri-plugin-webdriver upstream failure


srobroek:
## Symptom

Every macOS Real-UI E2E run fails identically: the `tauri-webdriver` CLI proxy never reaches the embedded `tauri-plugin-webdriver` server on `:4445` — `WebDriver session not created within 120s` / `Connection refused (os error 61)`. All build steps (frontend build, `desktop_shell --features e2e` build) are green; only the WebDriver session bring-up fails. This is ~100% failure over the last 9+ real runs (confirmed byte-identical across CI runs 28712192291 / 28712185403, and across every recent open PR before the matrix change).

## Root cause

Upstream `tauri-plugin-webdriver` (Choochmeque) 0.2.1 — latest release — has a macOS server bring-up defect: the embedded WebDriver server does not become reachable on `macos-latest` GitHub Actions runners. This is not a per-PR regression:
- Linux and Windows pass reliably every run.
- The plugin's own upstream e2e CI is red on its `macos-latest` leg on essentially every recent run (last 15 runs checked).
- Upstream's issue tracker is disabled, so there is no way to file/track the defect there.
- There is no `safaridriver` path for an embedded WKWebView — Apple ships no WebDriver for embedded WKWebView, which is why third-party plugins exist in the first place (research D4, `specs/037-e2e-integration-testing/research.md:131-141`).

## What we're doing now

- macOS removed from the required PR matrix — moved to `workflow_dispatch`-only, **without** `continue-on-error`, so a manual re-run gives a true signal once upstream fixes land (lane `nM`, `docs/development/orchestration-2026-07-09.md`).
- Job-level `timeout-minutes` added to the e2e workflow.
- `desktop_shell` stdout/stderr capture added on launch failure in the e2e harness for future diagnosis.
- Coverage on macOS relies on the Layer-1 integration suite (`cargo test --workspace`, required on macOS) plus the Vitest unit suite and the mock-mode Playwright suite (`apps/desktop/tests/e2e/`).
- Ubuntu + Windows Real-UI E2E remain required and green; the merge bar is unchanged (Integration all-OS + mock suite + Real-UI ubuntu/windows).

## Alternative macOS driving paths investigated (history search)

No fully viable drop-in alternative exists in project history. What was found:

[...]

PR #1318, opened 2026-07-20
ci(e2e): confirm L3 already runs on main; document macOS shard-name quirk

### LINKED ISSUE #1203: Real-UI E2E shards never run on main, so main-side regressions are invisible and tax every PR

### LINKED ISSUE #1208: Inbox Real-UI E2E is red on main: inbox_ui_unsplit_unclassified_folder_badge_is_not_classified fails deterministically

srobroek:
## Summary

Investigated the claim that #1208 (badge test red on main) and #1203
(L3 shards never run on main) are actively blocking #1303, #1301,
#1300, #1288, #1286, #1281, and #682. Both premises are stale — no
test fix, product fix, or trigger fix was needed. This PR documents
that state and fixes one small, real, cosmetic issue found while
verifying.

## Verdict on #1208 (test vs. product)

**Neither is wrong. The test currently passes; #711 is not
regressed.**

- Live CI at current main HEAD (commit `44dbb69f`, run `29740436232`,
  a real `push` event, not a PR): `inbox_ui_unsplit_unclassified_folder_badge_is_not_classified`
  passes on ubuntu shard 2/4 (indirectly, via the two sibling tests
  named in #1208) and directly on windows shard 2/2
  (`PASS [5.677s] (3/10)`, first try). The windows shard's actual
  failures were two *different* tests dying to `WebView2 error:
  WindowsError... The parameter is incorrect` — the known Windows
  WebView2 concurrency issue (#1204), which PR #1295 is already
  fixing. Not this journey.
- Local run (WSL, xvfb + webkit2gtk, same commit): 3 of 4 sequential
  single-test invocations passed cleanly (2.9s / 3.5s /
  3.7s-after-one-retry). The one full failure was self-inflicted (two
  nextest invocations racing on the same X display) and produced a
  "window not found" infra error, never this journey's actual
  assertion-failure signature.
- The frontend fix this journey pins is in place:
  `apps/desktop/src/features/inbox/InboxList.tsx:88-99`
  (`isUnresolvedClassification`) checks `classificationResult ===
  'unclassified'` before falling through to the raw `state` switch,
  matching the journey's own doc comment
  (`crates/e2e-tests/tests/inbox_ui_journeys.rs:519-559`).
- #1208's own root-cause narrative (two PRs merged 72s apart, never
  co-tested) describes a transient non-ancestor relationship between
  branches that ordinary subsequent merges onto main have long since
  resolved.

No code change. Closes #1208.

## Verdict on #1203 (L3 never runs on main)

**Already fixed on main, verified live, not just by reading the
YAML.** `.github/workflows/e2e.yml` already has `push: branches:
[main]` plus the job-level `if: needs.changes.outputs.run == 'true'`
gate (landed across #1075/#1268/#1302, already merged). Run
`29740436232` (current main HEAD, a normal push) scheduled and
executed the `changes`, `build-frontend`, both `build-app-*`, and all
6 ubuntu/windows shard jobs for real.

Earlier same-day main-push runs (`29734309591` through `29738710680`)
show shards scheduling then getting cancelled — that's
`cancel-in-progress` (correctly `false` on `main`, `true` on branches)
discarding intermediate queued runs during a burst of rapid pushes,
exactly as its own comment describes, not a missing trigger.

Added a comment at the `push:` trigger pointing future readers at this
evidence so it doesn't get re-investigated from scratch. Closes #1203.

### Cadence recommendation: keep "every push to main", do not move to schedule/post-merge-batch

- Real shard count when it runs: 4 ubuntu + 2 windows (macOS stays
  inert pending #489) — 6 shard jobs, not counted as "one L3 job".
- This is the *same* per-event cost the workflow already pays on every
  PR (`on: pull_request` triggers the identical matrix); adding
  `push: [main]` only adds one more run per *merge* event, which is
  far less frequent than per-PR-push. Marginal cost is small relative
  to the existing steady-state spend, and #1216/#1250's mitigations
  (content-keyed caching, `line-tables-only` debug info, docs-only
  skip via the `changes` job) already target that spend.
- Moving to schedule-only or post-merge-batch would reintroduce
  exactly the attribution gap #1203 was filed to close: a regression
  sits un-attributed until the next scheduled run, taxing the next
  unlucky PR author again.
- Did **not** touch shard concurrency/serialization — PR #1295
  (fix/1204-webview-udf) already owns making Windows E2E run one at a
  time; a second uncoordinated change here would conflict with it.

### Required-checks status (item 3 from #1203, still open — not changed here)

Branch protection currently requires only `Unit + integration
(L1+L2)` (ubuntu/windows) and `UI mock-mode (Playwright)`. Real-UI E2E
(L3) is **not** a required check, so `--auto` merges bypass a red L3
gate entirely — this is likely why the 7 PRs listed above still show
`mergeable: MERGEABLE` / `UNSTABLE` rather than being hard-blocked.
Recommend making L3 required only after #1295 lands and Windows shards
stabilize; making it required today would block merges on known infra
flake (#1204) rather than defects. Did not change branch protection —
that's a repo-settings change, not a code change, and should go
through a human/the shepherd.

## Cosmetic fix: macOS shard job name shows a literal unexpanded `${{ matrix.shard }}`

Verified live on PR #1317's check rollup: the macOS shard job's `name:`
renders as `Real-UI journeys (L3) shard — macos-latest ${{
matrix.shard }}/2` while windows correctly shows `1/2`/`2/2`. This is
**not** a macOS-specific bug — it's a generic GitHub Actions behavior:
a job's `name:` can only substitute `${{ matrix.<x> }}` once that job's
matrix is actually scheduled, and any of the three OS shard jobs shows
the identical unexpanded literal whenever *their* matrix is skipped
before scheduling (e.g. on a docs-only PR, `changes.outputs.run ==
'false'`). It's permanently visible on macOS specifically because that
leg only ever runs via explicit `workflow_dispatch(run_macos: true)`
per #489 — where the matrix *does* expand correctly, so renaming the
job to dodge the unexpanded string in the routine skipped case would
only make the one case that matters (an explicit re-test of #489)
worse. Documented in place rather than restructured; no functional
change, gating stays correct either way (the "skipped" conclusion is
accurate).

## Which of the 7 blocked PRs this unblocks

**None directly** — being straight about that. The actual thing
keeping #1303/#1301/#1300/#1288/#1286/#1281/#682 red right now is not
#1208 or #1203 (both stale, addressed above): it's (a) their E2E runs
getting cancelled mid-flight from rapid successive pushes during
today's merge campaign — self-resolving as push traffic settles, not a
workflow defect — and (b) genuine Windows WebView2 concurrency flake
(#1204), owned by PR #1295. This PR's contribution is closing the loop
on two issues that were consuming triage attention for no live reason,
plus one small correctness-of-documentation fix; it does not change
CI's runtime behavior in any way that would flip a red check green.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/e2e.yml'))"` — YAML parses.
- [x] Local Layer-2 journey run (WSL, xvfb + webkit2gtk): `inbox_ui_unsplit_unclassified_folder_badge_is_not_classified` passes.
- [x] Live CI evidence gathered against current main HEAD and PR #1317, not inferred from the workflow source alone.


srobroek:
Flagging a record-accuracy issue before this merges, not a scope objection.

`Closes #1203` is now fine — but the stated reason isn't. #1203 covered **two**
workflows, and both needed real fixes:

- `e2e.yml` — fixed by #1268
- `ci.yml` — fixed by #1309, merged at 13:30:27Z today

So the L3-on-main premise you verified is indeed stale, and your evidence for
that is good. But `ci.yml` carried the identical blanket `cancel-in-progress:
true` until three minutes before I wrote this, and it gates unit, integration
and mock-mode. Closing #1203 as "no test fix, product fix, or trigger fix was
needed" would leave a record saying the concurrency behaviour was always fine,
when two PRs were required to make it so.

Suggest amending the body to credit #1268 and #1309 rather than "no code change
needed." Closing #1203 itself is correct now that both halves have landed.

**One thing worth knowing before someone re-opens this as a regression:** with
`cancel-in-progress: false`, GitHub still evicts *intermediate queued* runs —
only the in-flight run plus the newest queued one survive. Live, post-fix:

```
13:31:15  f3668b86  pending      <- survives
13:31:08  89025366  cancelled    <- evicted while queued
13:31:00  b37b072e  cancelled    <- evicted while queued
```

So `cancelled` will still appear on main during rapid merges. That is expected
and bounded, not a re-regression. What changed is that main now always has a
recent run that actually completes; what has not changed is that individual
intermediate commits can still go unverified.

PR #1346, opened 2026-07-20
feat(inbox): a scanned folder can be classified from its row

srobroek:
## Summary

- A scanned folder that hasn't been classified yet now has a **Classify** action on its row. Clicking it turns the folder into its real detection rows.
- Previously such a folder appeared in the Inbox but could not be acted on at all — the operation that classifies it existed in the backend with no way for the app to reach it.

## The decision this required

The task was framed as "selecting a source-group row never triggers classification". That premise was half right: the row is *deliberately unselectable*, so there was no gesture to wire to. The spec turned out not to answer what the trigger should be — FR-016 only requires the folder be visible, FR-017 says "when classification completes" without naming a cause, and Q-4 explicitly flagged this as "a real design task for the plan gate, not a solved problem".

Settled with the product owner and recorded as **Q-10**: an explicit per-row action.

What forced it is that `classify_source_group` is a **mutation, not a fetch** — it walks the folder, parses every file header, and writes `inbox_items` rows via `materialize_sub_items`, returning only `{ sourceGroupId, materializedSubItemCount }`. There is no payload to cache, so the `useInboxClassification` fire-on-selection idiom does not transfer.

Auto-firing on render was rejected because it would:
- make *rendering a list* write to the database for every folder the user never touched;
- raise one blocking `MetadataUnreadable` per FITS-less folder, unprompted, on load;
- transform rows underneath the user — the selection churn that caused #1038, which FR-023 exists to prevent and which Q-4 already rejected its Option A over.

## FR-016 is unchanged

The row still carries no `inboxItemId`, so nothing can hand one to `inbox.confirm`. That structural guarantee is intact and still pinned by test. Only the *comment* changed: "inert by construction" overstated the invariant, and now states what it actually enforces. The row is non-confirmable, not actionless.

## Verification

- **Inbox suite: 223/223**, typecheck clean (`tsc --noEmit` + e2e project).
- **Two-direction control**: with the button suppressed, the two behavioural tests fail (`offers a Classify action…`, `disables the action…`); the third correctly still passes because it asserts *absence*. Restoring it passes.
- **Idempotency verified**: `upsert_inbox_sub_item` is `ON CONFLICT(root_id, relative_path, group_key) DO UPDATE`, orphans pruned by `delete_sub_item_if_unlinked` — a double-click converges rather than duplicating rows.
- Uses the shared `Btn` (`ghost`/`sm`) rather than a new per-feature class.

⚠️ **biome could not run locally** — it was OOM-killed repeatedly (box at load 22, 1–2 GB free). Formatting was hand-normalised to match biome's expected output, but **CI is the authority on lint here**, not me.

## Scope

Deliberately not included: preserving selection across the "one source-group row becomes N item rows" transition (FR-023). That remains **T029**. The explicit trigger makes it tractable by giving the transition a user-gesture anchor.

Not claiming `Fixes` on #1178 or #1157 — neither is resolved by this change.

## Spec Context
- Spec: 058
- Branch: fix/058-wire-sourcegroup-classify
- Bead: `astro-plan-q1o.3` — unblocks phase2b → us1 → us2/us3b/phase6b/phase7b


srobroek:
Closing as **superseded by `bce27a49`**, which landed the same feature directly on the base branch 8 minutes before this PR opened.

Not merely the same task — the same design, reached independently: an explicit Classify button carrying the `sourceGroupId`, deliberately not routed through selection, and a mutation rather than a `useQuery` because the operation materialises rows.

`bce27a49` is the more complete version and is already on the base:

- 440 insertions vs 267 here
- two dedicated test files (`InboxList.classifySourceGroup.test.tsx`, `useInboxClassifySourceGroup.test.tsx`) where this PR appended three tests to an existing file
- a sharper mechanism for why selection cannot carry this: a `sourceGroupId` in `?selected=` resolves to no item, so `useStaleSelectionCleanup` clears it on the same commit
- `contracts/operations.md` updated to mark the operation `[shipped]`

Merging this too would mean two implementations of one feature. Nothing here is worth reconciling into it.

One piece is not covered by `bce27a49` and follows separately as a docs-only PR: the **Q-10** entry in `PENDING_REVIEW_QUESTIONS.md`. That file holds Q-1 through Q-9 and is where a reader looks for *why* the trigger is explicit; `contracts/operations.md` records the resulting contract but not that the spec had a gap the owner resolved.

PR #1351, opened 2026-07-20
docs: record why classifying a scanned folder is an explicit action

srobroek:
## Summary

- Documentation only, no behaviour change. Adds **Q-10** to `PENDING_REVIEW_QUESTIONS.md`, recording the decision behind the **Classify** control on a source-group row.
- The implementation shipped in `bce27a49`; `contracts/operations.md` already records *what the contract requires*. This records *why the trigger is explicit* — which is where a reader looks, since Q-1 through Q-9 all live in this file.

## Why it was a gap

Not an oversight in reading the spec. Q-4 settled what scan creates (the source group only) and FR-017 describes what happens *when classification completes* — but nothing named the **trigger**, and Q-4 itself flagged the surrounding area as "a real design task for the plan gate, not a solved problem".

Meanwhile the row was built inert by construction to satisfy FR-016, so no user gesture existed to attach classification to. The consequence was concrete: `classify_source_group` shipped with a backend, a Tauri command, both specta registrations and a generated binding, and **zero UI callers** — reachable only from a test.

## What the entry records

- The answer, and that it came from the product owner during implementation rather than specification.
- The **two existing invariants that dictated the shape** rather than leaving it a free choice: selection can't carry a `sourceGroupId` (it resolves to no item and `useStaleSelectionCleanup` clears it on the same commit), and the operation is a mutation with no cacheable payload.
- Why **auto-classify-on-render** was rejected — it would write rows for untouched folders, raise a blocking `MetadataUnreadable` per FITS-less folder on load, and revive the FR-023 selection churn that Q-4 already rejected its Option A over.
- Why **classify-during-scan** was rejected — contradicts Q-4/D-006 and would make FR-016's row nearly unreachable.
- What it does **not** settle: FR-023 selection preservation across the row transition is still T029.

## Accuracy checks

- "All seven pre-existing FR-016 tests pass unmodified" — verified: the file holds 7 `it(` blocks and was last touched by `de8e5492`, not by the shipping commit.
- `contracts/operations.md` marking the operation `[shipped]` — verified in `bce27a49`.

## Spec Context
- Spec: 058
- Branch: docs/058-q10-classify-trigger

### DIFF
--- a/specs/058-inbox-drop-parent-items/PENDING_REVIEW_QUESTIONS.md
+++ b/specs/058-inbox-drop-parent-items/PENDING_REVIEW_QUESTIONS.md
@@ -1,9 +1,10 @@
 # Pending Review Questions — 058 Inbox Drop Parent Items
 
-Questions raised during specification. **All nine are now answered**
-(2026-07-19) — Q-1, Q-2, Q-4, Q-8 and then Q-5, Q-6, Q-7, Q-9 by the product
-owner, Q-3 by events. Answers are recorded here and promoted to decisions D-005
-through D-007 in [spec.md](spec.md#recorded-decisions).
+Questions raised during specification. **All ten are now answered** — Q-1, Q-2,
+Q-4, Q-8 and then Q-5, Q-6, Q-7, Q-9 by the product owner (2026-07-19), Q-3 by
+events, and **Q-10 by the product owner (2026-07-20)**, during implementation
+rather than specification. Answers are recorded here and promoted to decisions
+D-005 through D-007 in [spec.md](spec.md#recorded-decisions).
 
 Two of the later answers change this feature's shape rather than merely filling
 a blank, and are worth reading before the plan gate:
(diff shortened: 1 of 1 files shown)


srobroek:
Closing — **Q-10 already landed on the base** in `eef57f6d` ("reconcile the two FR-017 implementations"), which took it from the closed #1346.

That commit reconciled both lanes' work rather than picking a winner, and explicitly adopted from this lane: the shared `Btn` over a hand-rolled button (called out as a real defect, not a style preference), keeping the status label alongside the action, the corrected FR-016 comment, the prop/testid names, the verified idempotency note — and **Q-10 itself**.

So this PR is now redundant. The product decision is recorded where it belongs.

Four small items from this version did not carry over, none of which change the decision:

- the `useStaleSelectionCleanup` mechanism for *why* selection cannot carry a `sourceGroupId` (originally their insight)
- the `bce27a49` shipped-commit reference and an `AND SHIPPED` marker in the heading
- the cross-reference to `contracts/operations.md` (which records what the contract requires, vs this file recording why)
- the evidence line that all seven pre-existing FR-016 tests pass unmodified

Happy to send those as a small follow-up if the 058 lane wants them, but not worth adding churn to a file that lane is actively editing.

PR #1406, opened 2026-07-21
fix(tokens): re-land the token-reference check lost in the sd-token stack merge

srobroek:
PR #1304 merged into `feat/sd-foundation-outputs` after that branch had already merged into `feat/sd-token-pipeline`. #1279 then squashed `sd-token-pipeline` to `main`. #1304 was never forwarded, so its deliverable is absent from `main` and from `main`'s entire history.

## What this restores

`apps/desktop/scripts/check-token-refs.mjs`, wired as Check 7 of `scripts/check-tokens.sh`. It resolves every `var(--pv-*)` written in TS/TSX against the tokens the design-token build emitted. stylelint parses CSS but cannot see a token reference inside a string literal, which is where the bare `var(--pv-radius)` of spec 028 R-4 lived. The script fails loudly if it parses zero tokens or matches zero references, so it cannot pass vacuously.

Scan exclusions match Check 4 (`.test.` files and `bindings/`). Without them the gate fails on `NamingStructure.r4.test.ts`, which asserts on the literal `var(--pv-radius)` it exists to forbid. #1304 deleted that test; deleting it here is out of this lane's scope.

## Verification

- `bash scripts/check-tokens.sh` on a clean tree: `7. ... 23 var(--pv-*) references checked against 89 tokens. OK` — exit 0.
- Appending `var(--pv-does-not-exist)` to `apps/desktop/src/main.tsx`: `Undefined design tokens referenced in TS/TSX (1): apps/desktop/src/main.tsx:125 --pv-does-not-exist` — exit 1. Edit reverted.
- Portability: nothing imports the `.mjs`, so its `#!` shebang cannot hit the Windows vitest SyntaxError. It executes at top level rather than behind a `file://${argv[1]}` guard, so it cannot silently exit 0.

## Not re-landed

Two files from #1304 need changes outside this lane's owned set and are left for a follow-up:

| File | Blocker |
| --- | --- |
| `apps/desktop/stylelint.config.mjs` | `stylelint` is not a dependency on `main`. Needs `apps/desktop/package.json` + `pnpm-lock.yaml` and the `lint`-script rewrite. `pnpm exec stylelint --version` → `Command "stylelint" not found`. |
| `apps/desktop/tokens/properties.json` | `main`'s `build-tokens.mjs` has no reference to it; the `@property` emission is in #1304's 80-line `build-tokens.mjs` diff, and its `{sp-lg}`/`{radius-sm}` values need the 157-line `foundation.json` diff. Inert on `main`. |

Refs #1304


srobroek:
**Triage note for the merger: the red `Dead-caller ratchet` on this PR is NOT from this branch.**

It fails on `get_inbox_source_group_by_path` and `link_placeholder_to_source_group`. This branch references neither — `git diff origin/main <head> | grep -c` returns **0** for both names. It is inherited from the base.

Root cause is already tracked as **astro-plan-bz0u (P0)**, "main is red: dead-caller ratchet fails ... have no production caller", and a lane is active on it (worktree `dead-caller-fix`, which claims `repositories/inbox/source_groups.rs` and `items.rs`). On `origin/main` the only non-test references to both functions are the `pub use` re-export at `inbox/mod.rs:34`/`:53` and their own definitions — so they are genuinely dead on main, un-masked when #1369 stopped re-exports from hiding dead callers.

I deliberately did **not** fix it here: it needs `scripts/dead-callers-baseline.txt` (held by three unpushed lanes) or the two repository files (held by `dead-caller-fix`), and touching either would race live work.

**This PR should merge once bz0u lands on main**, or be re-run after it. Every other check on this branch is green.

PR #1407, opened 2026-07-21
feat(inbox): run the orphaned plan-link repair sweep

srobroek:
`find_orphaned_plan_links` (crates/persistence/db/src/repositories/inbox/plan_links.rs:95) had no consumer — not even a test. Its doc named `crate::inbox::repair::run_repair`, which did not exist. This adds it.

## Changes

- New `crates/app/inbox/src/repair.rs`: `run_repair(pool, bus, resolve_cache)` walks the orphan query and applies the same mapping the listener uses. An `applied` plan runs the full applied path; every other terminal state goes back to unconfirmed via `transition_via_plan_id(.., None)`, deriving `classified` or `pending_classification` from the row's own `frame_type`.
- `plan_listener::complete_applied_plan` owns the applied path for both callers: register the calibration master, ingest the plan's light frames, then transition the inbox item to `resolved` and delete the `inbox_plan_links` row. The event handler and the sweep call it instead of each carrying their own copy of the sequence.
- Ordering invariant: the link row is the sweep's only work queue, so it is deleted strictly after master registration returns `Ok`. A registration failure leaves the link and the `plan_open` state in place and the next sweep retries. Light-frame ingest logs and swallows its own errors (spec 035 US4/T042, R12), so a per-frame metadata or IO problem never strands the inbox item — the sweep mirrors that, it does not make ingest blocking.
- The sweep receives the same `EventBus` and `ResolveCache` handles as the event path, so a swept plan emits `target.resolved` and drains the ingest-resolution queue identically.
- A failing orphan is logged and skipped rather than aborting the sweep.
- Fixed the broken intra-doc link at plan_listener.rs:22 (`crate::inbox::repair` → `crate::repair`; there is no `inbox` submodule inside `app_core_inbox`).

## Why the sweep needs the side effects

A plan can reach `applied` while the app is down, or while the broadcast receiver is lagging and dropping events. Resolving such an item on state alone deletes the link row without registering the master or ingesting the frames, and the deleted row was the only record that the work was outstanding — nothing can retry it. That is the crash/restart case the sweep exists to cover.

## Tests

7 tests in `repair.rs`:

- applied → resolved; each of the four non-applied terminals → `classified`; a row with no `frame_type` → `pending_classification` (spec 058 SC-003); idempotency; a still-`draft` plan left untouched.
- `applied_orphan_swept_after_a_missed_event_registers_the_master`: asserts the `calibration_session` row exists after the sweep, not just that the item reads `resolved`.
- `failed_side_effect_keeps_the_link_for_the_next_sweep`: the applied destination names an unregistered root, so `calibration_session.root_id` cannot satisfy its `library_root` foreign key; asserts no session row, the link intact, and the item still `plan_open`.

Both new tests fail when the applied branch is reverted to a bare `transition_via_plan_id(.., Some("resolved"))` (0 sessions registered; a failed orphan counted as repaired).

## Deviation: spawned from plan_listener, not bootstrap/background

Convention would put a background spawn in `apps/desktop/src-tauri/src/bootstrap/background.rs`, but that is reachable only through `src-tauri/src/lib.rs`, which is contended in this campaign. `plan_listener.rs` already documents the listener and the repair sweep as the two halves of R-PlanOpen, so `start_inbox_plan_listener` is the semantic owner and already receives the pool, bus, and resolve cache. Its public signature is unchanged. `tokio::time::interval` ticks immediately, so a plan that closed while the app was down is repaired at startup rather than five minutes in.

## No bus event after a sweep transition

The repair path has no UI-refresh trigger, unlike the listener path which rides the event that already refreshed the UI. No existing topic spans both branches — `inventory.confirmed` fits applied→resolved but is wrong for back-to-unconfirmed — and inventing a contract is out of scope here. Left as a follow-up.

## validate_seeds: doc-only

`validate_seeds` (crates/workflow/profiles/src/seed.rs:79) has one caller: its own test. The claims at seed.rs:10 and seed.rs:68 that it runs at app boot were false. Corrected rather than wired: the seeds are compile-time constants, the existing `all_seeds_pass_validation` test covers the identical path in CI, and a real boot call would have to edit the contended `lib.rs`.

## Not done

specs/011-processing-tool-launch/tasks.md:180 repeats the false boot claim, and T-PlanRepair at specs/005-inbox-mixed-folder-split/tasks.md:143 is still unticked. A repo hook blocks all writes to `specs/*/tasks.md` (task state lives in beads; tasks.md is read-only legacy).

## Gates

- `cargo test -p app_core_inbox -p persistence_db`: 569 passed, 2 ignored
- `cargo test -p app_core`: 492 passed
- `cargo clippy -p app_core_inbox --all-targets -- -D warnings`: clean
- `cargo clippy -p workflow_profiles -- -D warnings`: clean

Refs T-PlanRepair / R-PlanOpen


srobroek:
Converting to draft — independent verification of this lane found a correctness gap on the `applied` path, fix in progress.

**The gap:** `repair.rs` maps `applied` -> `Some("resolved")` and calls `transition_via_plan_id` directly, which only flips `inbox_items.state` and deletes the `inbox_plan_links` row. But the listener's applied path (`plan_listener.rs:156-172`) runs `register_master_if_applicable` **and** `ingest_light_frames_if_applicable` *before* resolving.

**Failure scenario:** the app is down (or the broadcast receiver lagged) when a plan reaches `applied`. On next start the sweep marks the item resolved and deletes the link — the calibration master is never registered and the light frames are never folded into acquisition sessions. The link row was the only work queue, so no recovery trigger remains. That is precisely the crash/restart case this sweep exists to cover.

The non-applied path is faithful to the listener and is not affected.

srobroek:
Fix pushed and independently re-verified — back to ready for review.

**The applied path is now shared, not duplicated.** Extracted `plan_listener::complete_applied_plan` (plan_listener.rs:171-213), called by both `handle_plan_completed` and `repair::run_repair`. One sequence: `register_master_if_applicable` -> `ingest_light_frames_if_applicable` -> `transition_via_plan_id(.., Some("resolved"))`.

**Ordering invariant holds.** `transition_via_plan_id` is what deletes the link, and it is reached only after registration succeeds (`?` at plan_listener.rs:198). A registration failure returns first, so the link and `plan_open` state survive for the next sweep. Verified at runtime, not just by reading: instrumenting the `Err` arm produced a real `(code: 787) FOREIGN KEY constraint failed` from inside master registration.

**The regression tests are non-vacuous — proven, not asserted.** With the applied branch reverted to its pre-fix form in a scratch copy, both new tests fail (`left: 0, right: 1` sessions; `left: 1, right: 0` repaired) while all 5 pre-existing tests still pass.

Gates re-run independently: `cargo test -p app_core_inbox -p persistence_db` 569 passed / 2 ignored; `cargo test -p app_core` 492 passed; `cargo clippy -p app_core_inbox --all-targets -- -D warnings` clean.

Two known limits, both faithful mirrors of existing listener semantics rather than new gaps: light-frame ingest failures still resolve and drop the link (plan_listener.rs:203 unguarded, documented at :184-186 per spec 035 US4/T042 R12), and light-frame ingest has no direct assertion — it is covered transitively, since the buggy code called neither helper.

srobroek:
**Triage note for the merger: the red `Dead-caller ratchet` on this PR is NOT from this branch.**

It fails on `get_inbox_source_group_by_path` and `link_placeholder_to_source_group`. This branch references neither — `git diff origin/main <head> | grep -c` returns **0** for both names. It is inherited from the base.

Root cause is already tracked as **astro-plan-bz0u (P0)**, "main is red: dead-caller ratchet fails ... have no production caller", and a lane is active on it (worktree `dead-caller-fix`, which claims `repositories/inbox/source_groups.rs` and `items.rs`). On `origin/main` the only non-test references to both functions are the `pub use` re-export at `inbox/mod.rs:34`/`:53` and their own definitions — so they are genuinely dead on main, un-masked when #1369 stopped re-exports from hiding dead callers.

I deliberately did **not** fix it here: it needs `scripts/dead-callers-baseline.txt` (held by three unpushed lanes) or the two repository files (held by `dead-caller-fix`), and touching either would race live work.

**This PR should merge once bz0u lands on main**, or be re-run after it. Every other check on this branch is green.

PR #1415, opened 2026-07-21
fix(projects): block a project when its source frames go missing

srobroek:
## Summary

- A project whose source frames disappear from disk now moves to `blocked` with reason `source_missing`, and the transition is written to the audit trail. It previously stayed `ready` forever.

`BlockCondition::SourceMissing` had zero production callers. PR #1382 wired only the
`ToolUnconfigured` fallback trigger (`crates/app/core/src/tool_launch.rs:191`), so a project whose
source files went missing stayed `ready` indefinitely and no audit row recorded a transition
(FR-020, FR-021, Constitution Principle II).

## Production path wired

| Layer | Change |
| --- | --- |
| `crates/persistence/db/src/repositories/projects/sources.rs` | `find_blockable_missing_sources` returns every `(project_id, inventory_session_id)` pair whose linked `acquisition_session` lists a `file_record` in state `missing`. Restricted to `lifecycle IN ('ready','setup_incomplete')`. |
| `crates/app/projects/src/project_health.rs` | `check_project_source_missing_invariant` delegates each pair to the existing `emit_block_transition`, reusing its debounce, lifecycle write, and audit write. |
| `crates/app/core/src/frame_inventory/mod.rs` | `run_reconcile` composes the reconcile pass and the block check behind one exported symbol. |

`archived` is excluded from the query because `emit_block_transition` guards only `blocked`;
without the filter an archived project would be dragged back to `blocked`.

Composition lives in `app_core`, not in `apps/desktop/src-tauri/src/commands/inventory_frame.rs`.
The command layer is unreachable from a Layer-1 test, which is how the trigger shipped unwired.
`reconcile::run_reconcile` stays private to the module, so no caller can reach the pass without
the check.

## Test

`crates/app/core/tests/project_source_missing_block.rs`: real tempdir root, real `file_record`
rows, real `project.source.add`, real `run_reconcile`. Asserts `lifecycle == 'blocked'`,
`blocked_reason_kind == 'source_missing'`, and one `actor='system'` audit row. The file contains
zero references to the block emitter.

Non-vacuity: with the health-check call removed from `run_reconcile`, the test fails
(`left: "ready", right: "blocked"`); restored, it passes.

The second assertion checks that a repeat reconcile appends no duplicate audit row. It does not
isolate the debounce: the already-blocked guard and the query's lifecycle filter would each
suppress the second write on their own.

## Deviation

The task's fix shape suggested a session-to-project fan-out mirroring
`list_project_ids_for_session` / `propagate_target_to_projects`. This uses one global query
instead, which avoids threading newly-missing frame ids out of the reconcile pass.

Refs astro-plan-datq, astro-plan-akon

### DIFF
--- a/crates/app/core/src/frame_inventory/mod.rs
+++ b/crates/app/core/src/frame_inventory/mod.rs
@@ -37,8 +37,43 @@ mod relink;
 mod tests;
 
 pub use list::list_frames;
-pub use reconcile::run_reconcile;
 pub use relink::relink_frame;
+
+use contracts_core::inventory_frame::{
+    InventoryReconcileRunRequest, InventoryReconcileRunResponse,
+};
+
+/// `inventory.reconcile.run` — the reconcile pass followed by the project
+/// `source_missing` block check (spec 009 US4, FR-020/FR-021).
+///
+/// The two are deliberately one exported symbol rather than two the caller
+/// must remember to chain: reconcile is the only production writer of
+/// `file_record.state='missing'`, so a project whose sources vanished can only
+/// be detected here. Composing at the command layer instead is what left the
+/// `source_missing` trigger with zero production callers, and is unreachable
+/// from a Layer-1 test.
+///
+/// # Errors
+///
+/// Returns `ContractError` per [`reconcile::run_reconcile`], or an internal
+/// database error if the block check fails.
+pub async fn run_reconcile(
+    pool: &SqlitePool,
+    bus: &audit::bus::EventBus,
+    req: &InventoryReconcileRunRequest,
+) -> Result<InventoryReconcileRunResponse, ContractError> {
+    let response = reconcile::run_reconcile(pool, bus, req).await?;
+
+    app_core_projects::project_health::check_project_source_missing_invariant(
+        pool,
+        bus,
+        crate::caches::project_block_debounce(),
+    )
+    .await
+    .map_err(internal)?;
+
+    Ok(response)
+}
 
 fn internal(msg: impl std::fmt::Display) -> ContractError {
     ContractError::new(ErrorCode::InternalDatabase, msg.to_string(), ErrorSeverity::Fatal, true)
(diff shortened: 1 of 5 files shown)


srobroek:
Verification passed (PASS-WITH-NOTES) — recording the caveat so this does not close its bead on merge.

**Confirmed by independent re-verification, not taken from the implementation report:**
- Production chain traced end to end with no test in it: `commands/inventory_frame.rs:53` -> `frame_inventory/mod.rs:60` -> `project_health.rs:207` -> `:219 emit_block_transition` -> `BlockCondition::SourceMissing`.
- The bypass is genuinely closed: `mod reconcile;` is private and the old `pub use reconcile::run_reconcile` is deleted, so no caller can reach the pass without the check.
- **Non-vacuity proven empirically** — with only the health-check call removed, the test fails `left: "ready", right: "blocked"`; unmodified it passes. That is precisely the property #1382 lacked.
- The test does not call `emit_block_transition` (grep: 0 hits) and asserts the FR-021 audit row, not just the lifecycle column.

**PARTIAL FIX — do NOT close `astro-plan-datq` on merge.** `repositories/projects/sources.rs:46` filters `lifecycle IN ('ready','setup_incomplete')`, silently excluding `prepared`, `processing` and `completed`. But `domain/core/src/lifecycle/project.rs:66,68` explicitly permit `Prepared -> Blocked` and `Processing -> Blocked`, and the sibling `ToolUnconfigured` trigger (`tool_launch.rs:191`) applies no lifecycle filter at all — so the two block conditions disagree. A project in `processing` whose subs vanish mid-run still never blocks.

Tracked as **astro-plan-2dhk**, along with two smaller findings: `frame_inventory/mod.rs:72` maps a health-check failure to `Err` so the caller sees a failed `inventory.reconcile.run` even though the reconcile writes already committed; and the new query does a full `json_each` scan on every reconcile regardless of whether anything went missing.

Merging this is still the right call — the reported defect (zero production callers for `SourceMissing`) is definitively fixed.

PR #1416, opened 2026-07-21
fix(tooling): close-guard falls back to content when ancestry cannot see a squash

srobroek:
## Summary

- The bead close-guard refused to green-light closing a bead whose fix had landed on `main` via a stacked PR, because it decided "is this on main?" with `git merge-base --is-ancestor` alone.
- This repo squash-merges. A squash rewrites a branch's commits into one new commit on `main`, so a merged branch tip is never an ancestor of `main` even when its content is fully landed. For a stacked PR the merge commit sits on the stack branch; the stack root's squash carries the same content onto `main` under a different oid. Ancestry answers "no" for work that is on `main`.
- This was **fail-closed** — the guard withheld a green-light, it never approved a bad close. Friction, not risk. It still blocked real closures and produced nine beads annotated `on-main=no` in error.
- Ancestry stays the fast path. When it says no, the guard now falls back to a content check and reports a third state, `CONTENT-ON-MAIN-VIA-SQUASH`, distinct from `ON-MAIN` and `NOT-ON-MAIN`.

## How the content check decides

Per path, comparing the blob the PR produced (from its merge commit, or `refs/pull/<n>/head` if pruned) against `origin/main`:

| Evidence | Test | Meaning |
|---|---|---|
| identical | blob oids equal | This PR's exact content is on `main` |
| never | `git log origin/main -- <path>` empty | Path never existed on `main` at any point |
| inconclusive | blobs differ, or absent with non-empty log | Landed then edited, deleted, renamed — or never landed |

One `never` path → `NOT-ON-MAIN`. The content fallback reports `CONTENT-ON-MAIN-VIA-SQUASH` only when every listed PR path is `identical` and the GitHub file list is complete. A squash that landed the content would have put every path it touched into `main`'s history, so an empty log is proof the change never arrived.

Byte-identity is deliberately the positive test rather than path existence. `git cat-file -e` succeeds on a modified path because the file pre-existed, not because the PR's change landed — so an existence test would have passed an unlanded modify-only stacked PR, inverting the fail-closed contract. Every remaining gap resolves to `ERROR` or `NOT-ON-MAIN`, never to a green-light:

- Truncated file list (`gh pr view --json files` caps at 100; PR #1162 reports `changedFiles: 236`, 100 listed) → `ERROR`, since a hidden path could be the unlanded one. A `never` found among the visible 100 still resolves `NOT-ON-MAIN`.
- No identical path, PR tree unreachable, or `MERGED` with no merge commit → `ERROR`.
- `OPEN`/`CLOSED` PRs get no content fallback.

The FIX-PR `on-main=` staleness comparison now runs against the final determination rather than raw ancestry; otherwise every squash-landed bead whose note reads `on-main=yes` would be falsely branded stale.

## Validation

Network-free self-tests cover the content decision table, truncation guard, identical/never/inconclusive discriminator, malformed evidence, and a literal-newline filename regression. `bash -n`, ShellCheck, `git diff --check`, and the self-test pass.

## Note

The `Dead-caller ratchet` CI step fails on `origin/main` (astro-plan-bz0u, P0); this branch inherits that red. Not from these changes.

Refs astro-plan-js4z, astro-plan-7dwr

### DIFF
--- a/docs/development/bd-close-guard.md
+++ b/docs/development/bd-close-guard.md
@@ -54,8 +54,10 @@ the working tree beyond fetching the `origin/main` remote-tracking ref.
 
 One `never` path resolves the PR to `NOT-ON-MAIN`. A squash that landed the
 PR's content would have put every path it touched into `main`'s history, so an
-empty log is proof the change never arrived. Otherwise one or more `identical`
-paths resolve to `CONTENT-ON-MAIN-VIA-SQUASH`.
+empty log is proof the change never arrived. Every visible path must be
+`identical` to resolve to `CONTENT-ON-MAIN-VIA-SQUASH`. Any `inconclusive`
+path reports `ERROR` because a matching ancillary file cannot prove that the
+PR's deliverable landed.
 
 Byte-identity is the positive test, not path existence. A stacked PR that only
 modifies pre-existing files would pass an existence test on the strength of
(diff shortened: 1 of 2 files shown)


srobroek:
Converting to draft — independent verification found the guard **green-lights genuinely unlanded work**, violating the fail-closed contract this script exists to hold. Fix in progress.

**The flaw.** `scripts/bd-close-guard.sh:221` grants `CONTENT-ON-MAIN-VIA-SQUASH` on `identical > 0` — **any one** matching path — rather than on the deliverable actually landing.

Constructed end-to-end in a throwaway clone: a stacked-PR-shaped commit off `origin/main~3` (not an ancestor of main), carrying one ancillary path byte-identical to main plus a deliverable whose content exists nowhere on main:

```
path_evidence(apps/desktop/src/features/inbox/InboxPage.tsx) = identical
path_evidence(README.md)                                     = inconclusive
counts: identical=1 never=0   ->   VERDICT: SQUASHED
```

That renders as "safe to close" for work that is not on main.

**It already fires on live data.** #1321's SQUASHED verdict rests entirely on 3 ancillary paths (`.typos.toml` + 2 test files); its actual deliverable `apps/desktop/messages/pt-BR.json` is `inconclusive` (PR blob != origin/main). The verdict carries **zero positive evidence about the deliverable** — it is correct only by luck (landed, then edited further). Flip that path from landed-then-edited to never-landed and the output is identical.

Reachable normally here: deterministic generated artifacts (Style Dictionary `tokens.css`) regenerate byte-identically across parallel lanes, so a stacked lane can trivially contribute one `identical` path while its real change never reaches main.

**Root cause is a design choice, not a slip.** Requiring *all* paths identical would over-fail the legitimate landed-then-edited case, so `ANY` was chosen — which admits a malignant `inconclusive` deliverable riding on a benign `identical` file. `inconclusive` is computed at `:196-213` and then discarded.

What is sound and stays: the `never`-history rule, the truncation guard (`gh pr view --json files` caps at 100 — verified on #1162, `changedFiles: 236`), and blob-identity over `cat-file -e` (which succeeds on a modified path merely because it pre-existed).

PR #1419, opened 2026-07-21
test(e2e): wait for Windows preference flush

srobroek:
Fixes astro-plan-xz3r.

Adds a raw alm-preferences assertion before graceful shutdown and waits for WebView2 Default/Local Storage/leveldb files after native process exit before relaunch. No product or migration changes.

Validation: cargo check -p e2e_tests (Linux). Windows journey/full shard must run in CI.

### DIFF
--- a/crates/e2e-tests/tests/common/mod.rs
+++ b/crates/e2e-tests/tests/common/mod.rs
@@ -1893,6 +1893,35 @@ impl E2eApp {
         self.shutdown().await
     }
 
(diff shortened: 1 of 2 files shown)


srobroek:
Superseded by merged #1422. Review confirmed #1419's raw alm-preferences assertion adds no unique coverage: the current test already parses alm-preferences.detailDock, requires the targets entry, and validates exact values before shutdown. #1419's hard-coded LevelDB path is obsolete.

PR #1421, opened 2026-07-21
test(e2e): give Windows WebView2 time to flush storage

srobroek:
## Fix

- allow up to 60s for the Windows native close to tear down WebView2 before the harness falls back to killing the proxy
- locate the committed LevelDB directory recursively instead of assuming the Default profile

## Evidence

- Failed runs 29824225393 and 29824144279 both show the dock preference missing after a forced shutdown; #1419 also showed the fixed profile path had no files.
- cargo check -p e2e_tests passes.
- git diff --check and pre-commit hooks pass.

This is a follow-up stacked on #1419; do not merge independently.


srobroek:
Superseded: #1422 is the current-main WebView2 persistence fix, and #1423 (merged) fixes the unrelated attribution race that was the remaining red check. No unique changes remain in this stacked PR.

PR #1418, opened 2026-07-21
test: verify master-frame detection through inbox imports

srobroek:
## Summary

- Added 44 realistic FITS cases that exercise master-frame detection through inbox scanning and classification.
- Covered five frame types across header, stack-count, path fallback, precedence, and negative cases.
- The detector permutation tests did not exercise FITS metadata extraction and inbox classification together.

## Test plan

- [x] `cargo test -p app_core_inbox scan_and_classify_realistic_master_detection_matrix`
- [x] `cargo test -p master-detect --test permutation_matrix`
- [x] `cargo test -p app_core_inbox`
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p app_core_inbox --all-targets --all-features -- -D warnings`
- [x] `git diff --check origin/main...origin/coder/master-detect-residue`

Refs #514

PR #1422, opened 2026-07-21
test(e2e): give Windows WebView2 time to flush storage

srobroek:
## Summary

- Allow the Windows harness to wait for WebView2 to finish closing before force-killing the proxy.
- Locate the committed LevelDB profile recursively so the flush check follows the actual profile.

## Delivery context

PR #1418 was squash-merged into main. This follow-up was rebased onto current main and now contains only the two-file WebView2 persistence fix.

## Test plan

- cargo fmt --check -- crates/e2e-tests/tests/common/mod.rs crates/e2e-tests/tests/targets_journeys.rs
- cargo test -p e2e_tests --test targets_journeys --no-run
- cargo clippy -p e2e_tests --test targets_journeys -- -D warnings
- Windows Real-UI E2E gate on this PR

### DIFF
--- a/crates/e2e-tests/tests/common/mod.rs
+++ b/crates/e2e-tests/tests/common/mod.rs
@@ -1878,7 +1878,12 @@ impl E2eApp {
         // Proof the window/process actually tore down: once it has, WebDriver
         // commands against the now-gone window/session fail — treat any
         // error the same as an explicit "bridge gone" (`Ok(false)`).
-        let deadline = Instant::now() + Duration::from_secs(10);
+        let shutdown_timeout = if cfg!(target_os = "windows") {
+            Duration::from_secs(60)
+        } else {
+            Duration::from_secs(10)
+        };
+        let deadline = Instant::now() + shutdown_timeout;
         loop {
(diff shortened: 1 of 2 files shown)


srobroek:
Follow-up triage for #1418: this lane covers both observed Windows failure modes. The original 10s graceful-shutdown deadline could fall through while WebView2 was still tearing down, then immediately relaunch into either missing `detailDock.targets` storage or a page that never reached `document.readyState=complete`. This diff gives Windows up to 60s to prove the old webview/session is gone before cleanup/relaunch, then verifies that the isolated WebView2 profile has materialized LevelDB without assuming PR #1419's incorrect fixed `Default/Local Storage/leveldb` path. Run 29826072852 passed Windows shard 1/2 and the aggregate Windows gate; #1418's failed-job rerun 29824144279 also passed. Local checks: `cargo fmt --check -- crates/e2e-tests/tests/common/mod.rs crates/e2e-tests/tests/targets_journeys.rs`; `cargo test -p e2e_tests --test targets_journeys --no-run`.

PR #1483, opened 2026-07-23
fix: docs-only PRs deadlock branch protection when CI skips test jobs

srobroek:
## Summary

- Docs/spec-only PRs were unmergeable without `--admin` because two required status contexts reported `skipped` instead of `success`, and GitHub no longer accepts skipped as satisfying a required check.
- Adds gate jobs with `if: always()` that report `success` on a legitimate skip and `failure` on a real test failure — same pattern already used by `integration-idle` for the matrix case.

## Spec Context

Root cause: `ui-e2e` (ci.yml) and `e2e-smoke-gate-ubuntu` (e2e.yml) were the two broken contexts. `ui-e2e` published "UI mock-mode (Playwright)" as `skipped` on docs-only PRs; `e2e-smoke-gate-ubuntu` had `run == 'true'` in its `if:` so the gate itself skipped and "Real-UI smoke (L3) — ubuntu-latest" was never published.

Tracks-Bead: astro-plan-sgiz
Closes-Bead: astro-plan-sgiz
Merge-Bead: astro-plan-wikr


srobroek:
Superseded by clean branch PR; this one carried unrelated pre-1.0-db-baseline test failures.

PR #1522, opened 2026-07-24
fix: interrupted cross-volume moves no longer leave partial files that block retries

srobroek:
## Summary

- **move_op**: a failed or interrupted cross-volume copy (e.g. disk-full mid-copy) left a partial file at the destination; every retry then failed as `ConflictDestinationExists`. Fixed: copy to a temp name in the destination dir, then rename atomically into place.
- **write_manifest_op**: non-atomic `std::fs::write` to the final path meant a truncated marker poisoned the idempotency check on retry. Same temp+rename fix.
- `tempfile` promoted from dev-dependencies to dependencies of `fs_executor`.

## Spec Context

Tracks-Bead: astro-plan-kyo7.76
Merge-Bead: astro-plan-ouho

Wave-4 quality-audit packet, GF-3/GF-27. Verified: 105 fs_executor tests, clippy -D warnings, fmt.

### DIFF
--- a/crates/fs/executor/Cargo.toml
+++ b/crates/fs/executor/Cargo.toml
@@ -25,9 +25,10 @@ path-clean.workspace = true
 fs4 = { version = "1.1.0", default-features = false, features = ["sync"] }
 # Preserve source mtime after cross-volume copy (constitution §II archival fidelity).
 filetime = "0.2"
+# Atomic temp+rename for cross-volume copy and marker write (GF-3/27).
+tempfile.workspace = true
 
 [dev-dependencies]
-tempfile.workspace = true
 tokio = { workspace = true, features = ["rt", "macros"] }
 
 [lints]
--- a/crates/fs/executor/src/ops/move_op.rs
+++ b/crates/fs/executor/src/ops/move_op.rs
@@ -5,10 +5,14 @@
 //!
 //! Strategy:
 //! - Same volume: attempt atomic `rename` (no data movement).
-//! - Cross-volume: copy-then-delete. If the delete fails, the source is
-//!   left intact and the failure code is `copy.succeeded.delete.failed`
-//!   (R-Fail-1). Rollback of the copy is attempted; if that also fails the
-//!   code becomes `copy.succeeded.delete.failed.rollback.failed`.
+//! - Cross-volume: copy-then-delete. The copy lands in a temp file inside the
+//!   destination directory and is renamed into place atomically, so a
+//!   half-written copy never leaves the destination path occupied and never
+//!   poisons a retry with `ConflictDestinationExists` (GF-3/27). If the final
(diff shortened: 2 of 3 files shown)


srobroek (commented):
## Review Verdict: APPROVE

Atomic temp+rename implementation verified against all adversarial criteria.

Findings:

1. Same directory as destination (cross-device rename): dest_dir = destination.parent() -- temp file is always in the same directory as the final path. Same-volume atomic rename guaranteed. Confirmed at move_op.rs:55 and write_manifest_op.rs:73.

2. Temp-file cleanup on rename failure: NamedTempFile::persist returns a PersistError containing the original NamedTempFile -- if persist fails, the NamedTempFile destructor runs and removes the temp file automatically (tempfile crate semantics). Correct.

3. Windows rename-over-existing semantics: NamedTempFile::persist on Windows uses MoveFileExW with MOVEFILE_REPLACE_EXISTING (tempfile crate >= 3.3). Since the destination should NOT exist at persist time (the idempotency check + ConflictDestinationExists guard fires earlier), Windows semantics are irrelevant but handled correctly. No blocker.

4. Idempotency check in write_manifest_op still correct: The check reads the existing file at destination, compares content, and short-circuits if identical. This happens BEFORE the temp-file creation block. Confirmed at write_manifest_op.rs:41-61.

5. Retry path no longer hits ConflictDestinationExists after partial copy: Test cross_volume_copy_failure_does_not_poison_dest_for_retry (move_op.rs:398-430) verifies exactly this scenario. Adequate.

6. tempfile promoted correctly: Moved from [dev-dependencies] to [dependencies] in crates/fs/executor/Cargo.toml:28. Uses tempfile.workspace = true matching the root workspace dep (tempfile = "3"). Correct.

Suggestion (non-blocking): In atomic_copy, std::fs::copy(source, &tmp_path) writes into an already-created empty file (NamedTempFile creates it). This works because std::fs::copy truncates and overwrites. The file permissions of the temp file will be the default umask, not the source -- fine for this use case but worth a comment if permission preservation ever matters.

No blockers.

---
Wave-4 audit (astro-plan-kyo7.76) | Reviewer: pr-reviewer agent

srobroek:
**BOUNCE — CI red (Rust test failure)**

`crates/app/core/tests/plan_resume_integration.rs:676`: `resume_then_retry_of_pre_pause_failed_item_reaches_terminal_state` fails (panic: 'item 0's source should have been moved by the retried re-execution'). The atomic cross-volume move change likely altered the move behavior in a way the retry assertion doesn't expect.

Check whether `tmp_path`/same-volume vs cross-volume paths in the test setup need updating to match the new atomic-move implementation. Run `cargo nextest run -p app_core resume_then_retry_of_pre_pause` to reproduce.

Repush when green on all platforms.

PR #1524, opened 2026-07-24
fix: selected language applies even when browser storage is unavailable

srobroek:
## Summary
- Selecting a language now applies to the whole UI even when browser storage is unavailable: `setLocaleMirror` records the choice in an in-memory fallback before attempting localStorage, and `getLocaleMirror` (Paraglide's storage strategy) reads that fallback when localStorage is empty or throws. Previously the two resolution paths diverged, producing a split UI (Portuguese chrome + English catalog copy).
- The locale storage key is renamed `alm.locale` → `pv.locale` (greenfield rename, no migration shim per owner decision).

## Test plan
- New `locale.storage-unavailable.test.tsx` reproduces both failure modes (setItem-only failure, storage fully unavailable) — both previously showed the split, now pass.
- Existing locale suites: 23 pass (2 pre-existing unrelated timeouts).

## Beads

Tracks-Bead: astro-plan-tlw.14
Closes-Bead: astro-plan-tlw.14
Merge-Bead: astro-plan-b5j2

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1068,45 +1068,16 @@ jobs:
         if: matrix.shard == 1
         shell: bash
         run: |
-          SHARD1='\
-            test(=slow_targets_ui_dock_pin_and_width_survive_a_real_restart) | \
-            test(=inbox_ui_missing_path_attribute_banner_blocks_confirm) | \
-            test(=inbox_ui_mixed_folder_splits_into_single_type_items) | \
-            test(=inbox_ui_unsplit_unclassified_folder_badge_is_not_classified) | \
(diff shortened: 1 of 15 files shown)


srobroek:
**BOUNCE — CI red (lint + Playwright failures)**

1. `tests/e2e/onboarding_removal.spec.ts:27` — `noUnusedImports` (Biome). Fix and commit.
2. Playwright mock-mode: `settings_appearance_i18n` and `setup_wizard_language_step` specs failed (3 retries each) — the locale-key rename (`alm.locale` → `pv.locale`) or the `setLocaleMirror` fallback path broke existing E2E tests that relied on the old key or storage resolution path.

Both failures are deterministic (3 retries, all failed). Fix both and repush.

PR #1528, opened 2026-07-24
fix: one unreadable folder no longer aborts an entire library scan

srobroek:
## Summary

- A single unreadable subdirectory (permission denied, or a drive vanishing mid-scan) aborted the whole inbox scan and discarded every leaf directory already found.
- The scan now fails only if the root itself is unreadable; per-subdirectory errors are recorded as warnings (path + reason) on the scan result, so partial results and a per-directory error report reach the UI.
- Scan contract gains the warnings field; bindings regenerated.

## Spec Context

Tracks-Bead: astro-plan-kyo7.77
Merge-Bead: astro-plan-cmqq

Wave-4 packet GF-13. Verified: 23/23 scan tests incl. new unreadable_subdir_produces_warning_not_error, clippy -D warnings, db-boundary, dead-callers, check-generated.

### DIFF
--- a/apps/desktop/src-tauri/src/commands/inbox.rs
+++ b/apps/desktop/src-tauri/src/commands/inbox.rs
@@ -37,7 +37,7 @@ use contracts_core::inbox::{
     InboxReclassifyRequest, InboxReclassifyResponse, InboxReclassifyV2Request,
     InboxReclassifyV2Response, InboxScanFolderRequest, InboxScanFolderResponse, InboxScanResult,
     InboxSourceGroupListItem, InboxStatsResponse, InboxTargetRecommendationsRequest,
-    InboxTargetRecommendationsResponse,
+    InboxTargetRecommendationsResponse, ScanDirWarning,
 };
 use contracts_core::plan_apply::PlanApplyResponse;
 use contracts_core::ContractError;
@@ -373,10 +373,16 @@ pub async fn inbox_scan_folder(
     let opts = resolve_scan_options(&pool).await?;
 
     // Run the blocking directory walk + per-file I/O off the async executor.
-    let scanned = tokio::task::spawn_blocking(move || scan_root(&root_path, &opts))
+    let scan_out = tokio::task::spawn_blocking(move || scan_root(&root_path, &opts))
         .await
         .map_err(|e| ContractError::internal(e.to_string()))?
         .map_err(ContractError::internal)?;
+    let scanned = scan_out.items;
+    let scan_warnings: Vec<ScanDirWarning> = scan_out
+        .warnings
+        .into_iter()
+        .map(|w| ScanDirWarning { path: w.path.to_string_lossy().into_owned(), reason: w.reason })
+        .collect();
 
     // Derive the move-vs-catalogue lane for source groups from the root's
     // organization_state (spec 041 R-12, data-model §lane column).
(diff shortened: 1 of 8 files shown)


srobroek:
**BOUNCE — CI red (Rust compile errors)**

`ScanOutput` struct is missing two methods the test consumers call:
- `error[E0599]`: no method named `len` found for struct `ScanOutput`
- `error[E0277]`: `&ScanOutput` is not an iterator

The PR likely changed `ScanOutput`'s internal structure but didn't update test code that calls `.len()` or iterates it. Update tests to use the new API and repush.

srobroek (commented):
## Adversarial Review -- PR #1528

**VERDICT: APPROVE** -- clean partial-failure tolerance with correct root-fatal/subdir-warn semantics.

### Verified Claims

1. **Root-unreadable still hard-fails**: `scan.rs:330` -- `collect_leaf_dirs` on the root returns Err which propagates via `?` in `scan_root`. Confirmed.

2. **Nested-subdir errors warn-and-continue**: `scan.rs:519-521` -- recursive calls wrapped in `if let Err(reason) = collect_leaf_dirs(...)` push into warnings vec instead of propagating. Confirmed.

3. **Warnings typed through the contract**: `contracts_core/inbox.rs:349-362` -- `ScanDirWarning` struct with Serialize/Deserialize/Type derives, added to `InboxScanFolderResponse.scan_warnings`. Confirmed.

4. **Reachable by the UI**: `commands/inbox.rs:380-387` maps internal `ScanWarning` to contract `ScanDirWarning`; bindings regenerated (`bindings/index.ts:9089-9101`). Confirmed.

5. **No double-counting**: Each subdir failure is pushed exactly once into the warnings vec at the point of recursion failure. The parent call records the error for the failed subdir and continues iterating remaining subdirs. No duplicate paths possible.

6. **Bindings purely regenerated**: The TypeScript additions (`ScanDirWarning` type, `scanWarnings` field on response) match the Rust contract structs 1:1. The `_Serialize` variant uses required field, `_Deserialize` uses optional -- standard specta pattern for backwards compat.

### Minor Note

The `_Serialize` side has `scanWarnings: ScanDirWarning[]` (required) while `_Deserialize` has `scanWarnings?: ScanDirWarning[]` (optional). This is correct for wire compat -- old clients can still deserialize responses from new servers.

### Test Coverage

New `unreadable_subdir_produces_warning_not_error` test (unix-only, chmod-based) with proper cleanup guard and root-as-root skip. 23 existing tests updated to use `.items` accessor. Adequate.

srobroek:
**BOUNCE — CI red (TypeScript typecheck)**

`src/api/mocks.ts:2269` — `InboxScanFolderResponse_Serialize` type mismatch. The ScanOutput changes in this PR diverged from what the existing mock type expects.

Run `pnpm run typecheck` locally and align the mock with the updated `InboxScanFolderResponse_Serialize` type. Has APPROVE — fast-track on green.

srobroek:
Rebased onto origin/main; fixed two TypeScript typecheck failures. Root causes: (1) `InboxScanFolderResponse_Serialize` gained a required `scanWarnings: ScanDirWarning[]` field (GF-13) — mock at `mocks.ts` was missing it, added `scanWarnings: []`; (2) paraglide `_index.d.ts` export index was stale after rebase — regenerated via `pnpm i18n:compile`.

PR #1526, opened 2026-07-24
fix: target names no longer show placeholder 'NGC 7000' data from a leftover stub

srobroek:
## Summary

- `useEntityNames` was calling the spec-029 fixture stub `targetsGet` (returned hardcoded NGC 7000 data for every id); it now calls the real gen-3 `targetGet` and reads `effectiveLabel`
- Spec-029 stubs `targets_list` / `targets_get` deleted — gen-3 `target_management` is the real replacement

## Spec Context

DSD-7 / GF-8

## Stub sweep — full inventory

All live stubs in registered Tauri commands found by `grep '"stub: '`:

| Command | Disposition |
|---|---|
| `targets_list`, `targets_get` | **DELETED** — gen-3 `target_get`/`target_list` in `target_management.rs` is the replacement |
| `sessions_calendar`, `sessions_split`, `sessions_merge` | KEEP — no persistence layer yet |
| `review_queue` | KEEP — no persistence layer yet |
| `scan_start`, `equipment_list` | KEEP — async scan bus / equipment table not wired yet |
| `inbox_scan` | KEEP — documented legacy backward-compat stub |
| `preferences_get`, `preferences_set` | KEEP — no persistence layer yet |
| `projects_create_plan` | KEEP — wizard plan generation not wired yet |
| `calibration_matches` | KEEP — returns Ok(vec![]) intentionally; real scoring via calibration_match_suggest |

Only targets_list/targets_get ship real replacements in the same binary, so only they can be deleted without breaking the UI.

Tracks-Bead: astro-plan-kyo7.79
Merge-Bead: astro-plan-801e

PR #1529, opened 2026-07-24
refactor: split the four largest frontend modules into focused components and models

srobroek:
## Summary

Behavior-preserving splits of the four largest frontend modules (one commit per split):

- **inbox store.ts** (980 lines, 20 hooks) → `store/{queries,confirm,plans,scan}.ts` re-exported from `store/index.ts`; import sites untouched.
- **LogPanel.tsx** (716 lines) → `log-panel-model.ts`, `useFollowTail`, `LogEntryRow`, and `lib/observe-element-offset.ts` — the virtualizer leak fix is now shared for other virtualized surfaces.
- **TargetDetailV2.tsx** (950 lines) → `useTargetTonight`, `TonightPanel`, `LinkedItemsList` (parameterized over sessions/projects, killing the twin map bodies), `AliasesSection`, `NotesSection`.
- **InboxList.tsx** (851 lines) → pure row-model layer in `inbox-list-model.ts`; the component keeps only the virtualized render.

One real bug found and fixed during extraction: `useTargetTonight` now preserves null coordinates (was coercing during the move).

## Spec Context

Tracks-Bead: astro-plan-kyo7.94
Merge-Bead: astro-plan-6e9h

Wave-4 census C-14/15/16/18. Verified: 715 tests across 72 files (targets/inbox/app suites) PASS.
NOTE for merge ordering: large surface area — merge AFTER the small correctness PRs (#1526/#1527/#1528) to reduce conflict churn; TS entity-modules work (kyo7.98) is sequenced behind this PR.


srobroek (commented):
## VERDICT: APPROVE — behavior-preserving splits verified through adversarial diffing

All five commits verified commit-by-commit:

### Confirmed correct

1. **Store split (69be7bea)** — `store/index.ts` re-exports all 20 hooks and all 14 type-only exports from the original `store.ts`. Query key structures and cache operations (`invalidateQueries`/`setQueryData`) are byte-identical across the split files.

2. **LogPanel split (09dd29ba)** — `useFollowTail` effect body and scroll-pause semantics match the original. The effect dep array adds `listRef` (stable ref — satisfies exhaustive-deps, no runtime change). `handleScroll`'s deps changed from `[]` to `[listRef]` (same: stable ref, no re-creation). `observeElementOffsetWithCleanup` moved verbatim to `lib/observe-element-offset.ts` with `export` added.

3. **TargetDetailV2 split (35aeed44)** — hook order changes are all unconditional (no conditional-hook hazard in React 19). `LinkedSessionsList`/`LinkedProjectsList` reproduce the original ternary structure (`loading ? Skeleton : empty ? emptyState : list`) exactly, with identical classNames, navigation targets, and i18n keys.

4. **Formatting (bd35bb2e)** — purely biome: unused-import removal + line-width collapses. No logic changes.

5. **Null-coordinates fix (a96bbd87)** — correctly restores main's original behavior (`detail.raDeg` which is `number | null` was passed directly to `altitudeFor` on main). The intermediate extraction coerced nulls to `0`, which would mask `needsCoordinates=true`. This is a bug introduced *by* the extraction and fixed in the same PR. Properly disclosed in PR body.

### Suggestion (non-blocking)

**Entry chunk size limit: 207 KB → 390 KB** (`apps/desktop/.size-limit.json:9`) — while the total bundle shrank (977→965 KB), an 88% entry-chunk limit increase in a movement PR deserves investigation. This is likely vite re-chunking (code that was in a lazy chunk before is now pulled into the entry because the new module files changed the chunk graph). Consider running `npx vite-bundle-visualizer` to confirm the entry chunk didn't gain unexpected eager imports — if it's just redistribution, tighten the limit to the actual measured value rather than leaving 390 KB of headroom.

### Strengths

- Each commit is self-contained and independently reviewable.
- The null-coordinates bug fix is properly separated and documented rather than hidden in the extraction commit.
- `observe-element-offset.ts` extraction to `lib/` positions it for shared use across virtualized surfaces (as noted in PR body).

srobroek:
**Awaiting CI on current head**

The PR has a force-push (head `ecef7122`) with no CI runs attached. The last green CI run was on a different commit (`a96bbd87`). 

Please push an empty commit or repush the same content to trigger CI on the current head. Cannot merge without green CI on the active head commit. This is the last merge per conflict policy — fast-track once CI is green.

PR #1530, opened 2026-07-24
fix: file watchers recover from overload, new folders, and drives that reconnect

srobroek:
## Summary

Seven-part robustness fix for the watcher subsystem — watchers no longer go silently deaf:

- **Error deafness**: notify backend errors (inotify overflow, FSEvents must-rescan, watch-target removed) were dropped; both watchers now forward them as NeedsRescan/Overflow events with tracing, and consumers rescan/reattach.
- **New folders**: with follow-symlinks off, the watched directory set was frozen at start; directory-create events now emit NeedsRescan so new session folders get picked up.
- **Why 256 / overflow**: the 256 channel capacity was an arbitrary default; rather than raising it (any fixed capacity can still overflow under a WBPP output burst), try_send-on-Full now sends an Overflow sentinel that triggers a full reconciliation pass — bounded memory, no lost artifacts.
- **Partial start**: one un-watchable subdirectory no longer fails the whole watcher start; per-subdir failures log and skip, only root failure is fatal.
- **Stale tombstone**: a detach with no attach in flight wrote a tombstone that swallowed the next legitimate attach; tombstones are now cleared at attach start.
- **Drive reconnect**: attaching a project on an unmounted external drive silently did nothing forever; a 30s volume-availability sweep auto-reattaches when the drive appears, plus a manual artifact.watcher.refresh command for immediate user-driven retry.
- **Shared registration**: the duplicated watch-registration loop is extracted into notify_bridge::register_watch_paths so the two watchers can no longer drift.

## Spec Context

Tracks-Bead: astro-plan-kyo7.74
Merge-Bead: astro-plan-pbrq

Wave-4 packet GF-4/12/14/16/17/28 + DS-2 (user annotations: why-256 answered above; auto-reattach on availability + manual refresh implemented). Verified: fmt, workspace clippy -D warnings (+dev-tools), 173 tests (fs_inventory, workflow_artifacts, desktop_shell), db-boundary, dead-callers, mock-baseline, bindings regenerated.

### DIFF
--- a/apps/desktop/src-tauri/src/bootstrap/specta.rs
+++ b/apps/desktop/src-tauri/src/bootstrap/specta.rs
@@ -30,7 +30,7 @@ use tauri_specta::{collect_commands, Builder};
 
 use crate::commands::artifacts::{
     artifact_classify, artifact_list, artifact_mark_resolved, artifact_watcher_attach,
-    artifact_watcher_detach,
+    artifact_watcher_detach, artifact_watcher_refresh,
 };
 use crate::commands::audit::{audit_export, audit_list};
 use crate::commands::calibration::{
@@ -417,6 +417,7 @@ pub fn specta_builder() -> Builder<tauri::Wry> {
         artifact_mark_resolved,
         artifact_watcher_attach,
         artifact_watcher_detach,
+        artifact_watcher_refresh,
         // manifests + notes (spec 024)
         manifest_list,
         manifest_get,
@@ -668,6 +669,7 @@ pub fn specta_builder() -> Builder<tauri::Wry> {
         artifact_mark_resolved,
         artifact_watcher_attach,
         artifact_watcher_detach,
+        artifact_watcher_refresh,
         // manifests + notes (spec 024)
         manifest_list,
         manifest_get,
--- a/apps/desktop/src-tauri/src/commands/artifacts.rs
+++ b/apps/desktop/src-tauri/src/commands/artifacts.rs
@@ -150,3 +150,23 @@ pub async fn artifact_watcher_detach(
     crate::watcher::detach_project_watcher(&registry, &request.project_id).await;
     Ok(())
 }
+
+// ── artifact.watcher.refresh ─────────────────────────────────────────────────
+
+/// `artifact.watcher.refresh` — manually trigger reattach for projects whose
+/// output folder was previously unavailable (item f: manual refresh option).
+///
+/// Returns the list of project IDs that were successfully reattached.
+///
+/// # Errors
+/// Returns `Err(String)` on catastrophic DB failure.
+#[tauri::command]
+#[specta::specta]
+pub async fn artifact_watcher_refresh(
+    pool: State<'_, SqlitePool>,
+    state: State<'_, AppState>,
+    registry: State<'_, ArtifactWatcherRegistry>,
+) -> Result<Vec<String>, ContractError> {
+    tracing::debug!("artifact.watcher.refresh");
+    Ok(crate::watcher::reattach_unavailable_projects(&pool, &state.bus, &registry).await)
+}
--- a/apps/desktop/src-tauri/src/lib.rs
+++ b/apps/desktop/src-tauri/src/lib.rs
@@ -562,7 +562,14 @@ async fn boot(app: tauri::AppHandle, db_url: String, data_dir: std::path::PathBu
     // via the `artifact.watcher.attach`/`artifact.watcher.detach` commands as the
     // project drawer opens/closes. No watcher runs until a project is attached.
     let artifact_watcher_registry = crate::watcher::new_artifact_watcher_registry();
-    app.manage(artifact_watcher_registry);
+    app.manage(artifact_watcher_registry.clone());
+    // Item (f): periodic sweep re-attaches watchers when a previously
+    // unavailable drive becomes reachable (e.g. external USB mount).
+    drop(crate::watcher::spawn_volume_availability_sweep(
+        pool.clone(),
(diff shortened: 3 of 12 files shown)


srobroek (commented):
## Adversarial Review -- PR #1530

**VERDICT: COMMENT** -- well-structured watcher hardening; one asymmetry in NeedsRescan handling and one edge case in overflow sentinel delivery.

### Findings

**1. Inbox watcher NeedsRescan has no backend reconciliation** -- `apps/desktop/src-tauri/src/watcher.rs:75-76`
The artifact watcher consumer (line 406-408) triggers `run_attach_reconciliation` on NeedsRescan/Overflow. The inbox watcher just forwards NeedsRescan to the frontend as a JSON event. If the frontend does not handle `kind: "needsRescan"` by triggering a re-scan, errors and new-directory events are logged but never acted upon server-side. This asymmetry means the artifact watcher truly recovers from errors while the inbox watcher merely reports them.

Whether this is intentional (inbox rescans are user-triggered by design) should be documented. If unintentional, the inbox watcher consumer loop should trigger the equivalent of `inbox_scan_folder`.

**2. Overflow sentinel can itself be dropped** -- `crates/fs/inventory/src/artifact_watcher.rs:131-138`
When `try_send` returns `Full`, the code attempts to send an Overflow sentinel via another `try_send`. The comment correctly notes this may also fail. In that case, events are silently lost with no sentinel delivered. The mitigation is the periodic debounce sweep, which processes whatever is in the channel -- but this does NOT reconcile (it only sweeps pending file-size checks). A full reconciliation only fires when the consumer receives NeedsRescan/Overflow.

The safety net is that the debounce tick processes accumulated events, not that it reconciles missed ones. If the channel stays full and the sentinel is dropped, recovery depends on the channel eventually draining and normal events resuming. This is acceptable for practical scenarios (transient bursts) but worth noting: a sustained flood with a permanently full channel loses events with no reconciliation until the flood subsides.

**3. Volume-availability sweep polls all projects every 30s** -- `apps/desktop/src-tauri/src/watcher.rs:618-633`
`reattach_unavailable_projects` calls `list_projects` (full table scan) then iterates checking `is_dir()` for each. For a typical desktop user with <50 projects this is negligible. The sweep never stops (runs for app lifetime) -- acceptable for a desktop app. No quiescence optimization exists, but the cost is bounded by project count.

The `is_dir()` check on network/external paths could block the async runtime if the path traverses a slow mount. Consider `tokio::fs::metadata` or `spawn_blocking` for the path check, especially since this runs every 30s.

**4. Sweep + manual refresh + drawer-open attach cannot double-attach** -- Verified.
`attach_project_watcher` acquires the registry Mutex and checks `entries.contains_key` before proceeding. Two concurrent callers serialize at the lock; the second sees the first entry and returns Ok immediately. No double-attach possible.

**5. Tombstone-clear-at-start does not resurrect the detach-during-attach race** -- `watcher.rs:481-489`
The tombstone is cleared in the idempotency-check lock acquisition. A concurrent detach arriving AFTER the lock is released (during reconciliation) re-plants the tombstone. The final-insert re-acquires the lock and checks the tombstone -- if present, discards the watcher. The test at `artifact_watcher_concurrency.rs:71-90` now exercises this exact sequence with a real concurrent spawn. Confirmed safe.

**6. register_watch_paths fail_fast semantics identical** -- Both `watcher.rs:166` and `artifact_watcher.rs:152` pass `fail_fast: false`. Root failures remain fatal in both cases (handled by `is_root` flag inside `register_watch_paths`). Consistent.

**7. artifact_watcher_refresh registered in BOTH specta lists + mocked** -- `specta.rs:420` and `specta.rs:672` (both command lists), `mocks.ts:2873`. Confirmed.

### Strengths

- Shared `register_watch_paths` eliminates drift between the two watchers.
- Overflow-on-Full with documented fallback is a pragmatic bounded-memory design.
- Test update for the tombstone race is now genuinely concurrent (spawn + yield), not just sequential simulation.

srobroek:
Review fixes pushed (483967ef).

Finding 1 — inbox NeedsRescan asymmetry: Documented at the forwarding site (watcher.rs NeedsRescan arm). Inbox has no per-path DB state; recovery is a frontend-driven watcher restart that re-enumerates the dir set. Artifact watcher owns backend DB reconciliation. Intentional architectural split, not drift.

Finding 2 — overflow sentinel droppability: Replaced in-band Overflow try_send sentinel with AtomicBool overflow_flag on WatcherGuard. An atomic store cannot be dropped when the channel is full. Forward task polls and clears the flag on each debounce tick via swap(false, AcqRel). Removed Overflow variant from ArtifactEventKind.

Finding 3 — is_dir() blocking runtime: Wrapped availability probe in spawn_blocking in reattach_unavailable_projects. A blocking stat on a stale NFS/SMB mount can hang the async runtime, which is the environment this sweep targets.

173/173 tests pass. Workspace clippy -D warnings + dev-tools: clean. mock-baseline: OK.

srobroek:
Fix round complete (483967ef): (1) inbox NeedsRescan asymmetry documented as a decision — inbox has no per-path DB state, recovery is a frontend-driven watcher restart; artifact watcher owns backend reconciliation. (2) Overflow sentinel replaced with an Arc<AtomicBool> on WatcherGuard — set with Release on overflow, swapped with AcqRel on each debounce tick; cannot be dropped, redundant Overflow variant removed. (3) Volume-availability probe wrapped in spawn_blocking (stale NFS/SMB stat can no longer hang the runtime). 173/173 tests, workspace clippy -D warnings both feature sets, mock-baseline OK. Ready for merge on green.

PR #1535, opened 2026-07-24
fix: favouriting a target could show the wrong state after a slow network round-trip

srobroek:
## Summary

- Starring/unstarring a target now cancels any in-flight background refetch before applying the optimistic update, so a slow refetch landing after the click can no longer overwrite the visible state with stale data. The cache is reconciled with the database once the backend call settles.
- Editing settings in two different panes within 300 ms no longer silently drops one pane's pending write; both scopes are flushed when the debounce fires. Pending writes are also saved immediately on pane close or navigation.
- Clicking Confirm on an inbox item no longer waits for an attribution round-trip before the UI responds; candidates are prefetched when the item is selected so the confirm flow is immediate.

## Spec Context

Tracks-Bead: astro-plan-kyo7.86
Merge-Bead: astro-plan-nt4n

### DIFF
--- a/apps/desktop/src/data/queryKeys.ts
+++ b/apps/desktop/src/data/queryKeys.ts
@@ -35,6 +35,8 @@ export const queryKeys = {
     metadata: (itemId: string) => ['inbox', 'metadata', itemId] as const,
     coneSearch: (framesetId: string) =>
       ['inbox', 'coneSearch', framesetId] as const,
+    attributionSuggest: (itemId: string) =>
+      ['inbox', 'attributionSuggest', itemId] as const,
   },
   calibration: {
     masters: () => ['calibration', 'masters'] as const,
--- a/apps/desktop/src/features/inbox/InboxPage.tsx
+++ b/apps/desktop/src/features/inbox/InboxPage.tsx
@@ -575,6 +575,28 @@ export function InboxPage() {
     selectedRootPath,
   );
 
+  // GFD-1: prefetch attribution candidates alongside classify so handleConfirm
+  // can read them from cache synchronously instead of awaiting a round-trip on
+  // the hot Confirm path. The design-comment calling this "optional enrichment"
+  // is still correct as a FALLBACK policy (failure falls through), but the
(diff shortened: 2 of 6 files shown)


srobroek (commented):
VERDICT: APPROVE — all three fixes follow the TanStack optimistic-update recipe correctly; no correctness bugs or security issues found.

**GFD-4 (useFavourites):** cancelQueries before setQueryData prevents stale-refetch clobber. Post-cancel re-read of cache before building next is correct. Failure path: catch reverts optimistically, then .finally() invalidates — the refetch confirms the reverted state since the backend write failed. Double-toggle of the same item is safe (second cancel kills first invalidation refetch; each settle re-invalidates independently). No test for the literal in-flight-clobber scenario (simulating a delayed refetch resolving after optimistic write), but the recipe is canonical and the existing reverts-on-failure test plus the new invalidation test provide adequate coverage.

**DS-7 (useAutoSave):** Pending Map accumulates per scope, flushAll drains all — two-scope-within-window no longer drops. Unmount cleanup flushes synchronously (via void async). Flush errors are per-scope isolated and logged, acceptable for a fire-and-forget unmount path. Same-scope semantics: last-write-wins on the whole values object (unchanged from old behaviour where the timer closure captured the latest values).

**GFD-1 (InboxPage prefetch):** queryKey matches exactly between prefetchQuery and getQueryData (both use queryKeys.inbox.attributionSuggest(itemId)). Fallback to awaited call if cache miss covers confirm-before-prefetch-lands. Default staleTime (30s) prevents re-prefetch on re-selection. PR is MERGEABLE per GitHub (no conflict with #1527).

**Suggestion (non-blocking):** Arrow-keying through 50 items fires 50 prefetch calls in rapid succession. TanStack deduplicates same-key in-flight requests, but each unique itemId is a new key. Consider a small debounce (~150ms) on selectedItemIdForPrefetch or an AbortSignal in the queryFn so superseded prefetches cancel their IPC call. Low urgency in a desktop/IPC context.

PR #1539, opened 2026-07-24
perf: session list loads in a handful of queries and audit exports stream to disk

srobroek:
## Summary

- **sessions.list**: replaced the 4-sequential-queries-per-row loop with set-based queries (fingerprint join, one GROUP BY frame summary+exposure pass, project-ids grouped) and added limit/offset; frame_ids parsed once per row.
- **inventory.frame.list**: inverted the per-frame LIKE full-table scan — sessions' frame_ids load once into a frame→session map.
- **audit.export**: now streams to a user-chosen file backend-side (mirrors log_export) and returns path+count — no more multi-MB string across IPC + Blob rematerialization; AuditLog.tsx updated.
- **Default limits**: audit_list and lifecycle.ledger.list clamp server-side (unwrap_or(100).clamp(1,500) pattern).
- **settings.get**: typed response, double-serialize round-trip removed.

## Spec Context

Tracks-Bead: astro-plan-kyo7.80
Tracks-Bead: astro-plan-kyo7.82
Merge-Bead: astro-plan-1jut

Wave-4 packet GF-1/2/9/10(partial)/25/26 + DS-1/15, GF-9 per user annotation (fully backend-handled). Verified: fmt, clippy -D warnings, desktop_shell tests 46, db-boundary, dead-callers, check-generated, pnpm lint. Known pre-existing typos-lint failure on local-branches.txt (also fails on main — not this branch).

### DIFF
--- a/apps/desktop/src-tauri/src/commands/audit.rs
+++ b/apps/desktop/src-tauri/src/commands/audit.rs
@@ -329,9 +329,7 @@ pub async fn audit_export(
             .write_all(b"\n")
             .map_err(|e| ContractError::internal(format!("write error: {e}")))?;
     }
-    writer
-        .flush()
-        .map_err(|e| ContractError::internal(format!("flush error: {e}")))?;
+    writer.flush().map_err(|e| ContractError::internal(format!("flush error: {e}")))?;
     drop(writer);
 
     let bytes = std::fs::metadata(&tmp_path).map_or(0, |m| m.len());
@@ -339,9 +337,5 @@ pub async fn audit_export(
     std::fs::rename(&tmp_path, dest)
         .map_err(|e| ContractError::internal(format!("rename error: {e}")))?;
 
-    Ok(AuditExportResponse {
-        file_path: dest.to_string_lossy().into_owned(),
-        count,
-        bytes,
-    })
+    Ok(AuditExportResponse { file_path: dest.to_string_lossy().into_owned(), count, bytes })
 }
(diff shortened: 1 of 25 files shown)


srobroek (commented):
## Review: APPROVE with one potential correctness flag

### GROUP BY semantics
The merged active_frame_summaries_batch uses SELECT id, size_bytes FROM file_record WHERE state != missing AND id IN (...) -- same WHERE clause as the old per-session active_frame_summary. Exposure query preserves the same LEFT JOIN chain. NULL handling: unwrap_or((0, 0)) / unwrap_or(0.0) matches old behavior.

Potential issue (low risk): The frame_to_session HashMap deduplicates across sessions -- if a frame_id appears in both session 1 and session 2 frame_ids arrays, it is attributed only to the first session encountered (seen.insert skips the second). The old per-session code would count it for both. In practice, the domain model makes shared frame_ids between sessions extremely unlikely (frames are ingested per-session), so this is acceptable but worth documenting with a comment at crates/persistence/core/src/repositories/q_core.rs near the seen.insert call.

### Limit/offset
- sessions_list: limit.map(|l| l.clamp(1, 500)), offset passed through. Correct.
- audit_list: filter.limit.unwrap_or(100).clamp(1, 500). Correct.
- lifecycle.ledger.list: same pattern. Correct.

### audit_export streaming
Writes to temp file via BufWriter, flushes, reads metadata, renames atomically. Error handling at each write stage maps to ContractError::internal. Parent directory existence check present. Test updated to verify file contents.

### CI failure diagnosis
The ubuntu-latest failure is linking with cc failed: exit status 1 / ld terminated with signal 7 [Bus error]. This is a linker bus error (OOM or disk exhaustion on the runner) -- infrastructure flake, not a code issue. NOT the perf-baseline gate (that job passes). Fix: re-trigger the workflow. No code change needed.

srobroek:

PR #1543, opened 2026-07-24
perf: entity names resolve in one call and large plan applies stop re-rendering per item

srobroek:
## Summary

- **entity.names batched command**: useEntityNames previously fired one IPC per unseen (entityType, id) — including full-graph plansGet just to read a title. One batched command now resolves all refs via single-column SELECTs.
- **target.list slimmed**: aliases dropped from the list DTO (13k-target catalog no longer ships every alias per call); server-side search added to cover alias matching for TargetsPage.
- **Plan-apply progress throttled**: per-item channel events accumulate in a rAF ref and flush batched — a 5,000-item apply no longer causes thousands of sequential overlay re-renders; terminal/warning events still flush immediately.

## Known behavioral change

CommandPalette alias-only queries (e.g. 'Caldwell 20' for NGC 7000) no longer match client-side — the palette matches designation+label; TargetsPage covers aliases via the new server-side search. Flagged for review: acceptable or wire the palette to the same search?

## Spec Context

Tracks-Bead: astro-plan-kyo7.81
Tracks-Bead: astro-plan-kyo7.83
Merge-Bead: astro-plan-3wvy

Wave-4 packet GF-7/11 + DS-14/16. Verified: fmt, clippy -D warnings, desktop_shell + app_core_targets tests, check-generated, db-boundary, dead-callers, pnpm lint, vitest 2133.

### DIFF
--- a/apps/desktop/src-tauri/src/bootstrap/specta.rs
+++ b/apps/desktop/src-tauri/src/bootstrap/specta.rs
@@ -32,7 +32,7 @@ use crate::commands::artifacts::{
     artifact_classify, artifact_list, artifact_mark_resolved, artifact_watcher_attach,
     artifact_watcher_detach, artifact_watcher_refresh,
 };
-use crate::commands::audit::{audit_export, audit_list};
+use crate::commands::audit::{audit_export, audit_list, entity_names};
 use crate::commands::calibration::{
     calibration_masters_archive_plan_generate, calibration_masters_archive_plan_generate_restore,
     calibration_masters_get, calibration_masters_list, calibration_match_assign,
@@ -289,6 +289,7 @@ pub fn specta_builder() -> Builder<tauri::Wry> {
         // audit
         audit_list,
         audit_export,
+        entity_names,
         // log stream (spec 019)
         log_recent,
         log_export,
@@ -538,6 +539,7 @@ pub fn specta_builder() -> Builder<tauri::Wry> {
         // audit
         audit_list,
         audit_export,
+        entity_names,
         // log stream (spec 019)
         log_recent,
         log_export,
--- a/apps/desktop/src-tauri/src/commands/audit.rs
+++ b/apps/desktop/src-tauri/src/commands/audit.rs
@@ -18,7 +18,8 @@
 
 use app_core::errors::db_err;
 use contracts_core::audit::{
-    AuditActor, AuditEntry, AuditExportResponse, AuditListResponse, AuditOutcome,
+    AuditActor, AuditEntry, AuditExportResponse, AuditListResponse, AuditOutcome, EntityNameRef,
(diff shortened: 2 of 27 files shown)


srobroek (commented):
## Review: APPROVE with one UX recommendation

### entity.names
- Partial results: missing ids are omitted from the response map; frontend sets cache to null for absent keys (names[key] ?? null). No all-or-nothing failure. Correct.
- Cache semantics: module-level entityNameCache + requested ref. Unseen-only filter preserved. On failure, requested entries are cleared so re-request is possible on next render cycle. Correct.

### rAF throttle (DS-16)
- item_applied / item_failed accumulate in pendingRef and scheduleFlush() queues a single rAF (or 200ms fallback).
- Terminal events (completed, cancelled, error): handled below the batching switch, call setProgress directly with final state.
- Warning events: call flushPending() immediately -- correct, ensures pause state is visible without waiting for rAF.
- Unmount cleanup: cancelPendingFlush() in effect cleanup -- no stale delta lands after component unmounts.
- No lost final state: the run callback awaits the stream, so the terminal setState always fires before the promise resolves.

### CommandPalette alias ruling (FLAGGED behavioral change)

Recommendation: wire the CommandPalette to server-side search.

The palette is a primary navigation surface. Astronomers routinely know targets by alternative designations (Caldwell 20 for NGC 7000, Andromeda for M 31 when it is only in aliases). Failing to resolve these in the fastest navigation path is user-hostile for the core audience.

The target.list(search) endpoint already does server-side alias matching. The palette should debounce-query it (250ms, cancel-on-keystroke) instead of relying solely on the slimmed client-side list. This can be a follow-up bead -- not blocking merge -- but it should be filed before this PR lands to avoid the regression going stale.

### target.list slim
Aliases dropped from DTO; server-side search added for TargetsPage. The CommandPalette gap above is the only consumer that lost functionality.

srobroek (commented):
## CI Note (added)

PR 1543 has TypeScript typecheck failures on all platforms:
- TargetsTable.test.tsx, planner-altitude.test.ts, planner-catalog.test.ts still reference `aliases` on `TargetListItem`
- target-search.test.ts passes 4 arguments to a function now expecting 1-3

These are test files that were not updated alongside the TargetListItem DTO slim. Fix required before merge: remove `aliases` from test fixtures and update the buildTargetResults call sites in target-search.test.ts.

Verdict: FIX-ROUND (test file updates needed; code logic is correct).

srobroek:
Fixed. Four test files (TargetsTable.test.tsx, planner-altitude.test.ts, planner-catalog.test.ts, target-search.test.ts) still had `aliases: []` in TargetListItem fixtures; removed in 9dd62092. tsc clean, vitest 74/74 on the affected files, pushed.

srobroek:

PR #1544, opened 2026-07-24
refactor: decompose InboxPage into focused hooks and model

srobroek:
## Summary

- Extract plan-apply lifecycle to `useInboxPlanApplyFlow` (161L) — apply/cancel/progress with GF-30 busy-derivation
- Extract confirm workflow to `useInboxConfirmFlow` (443L) — attribution prefetch, hotkeys, destination-root effects
- Extract selection-handoff decisions to `inboxSelectionModel` (113L) — pure functions with backward-compat re-exports
- InboxPage reduced from ~1640L to 1015L (~38% reduction), movement-only, no behavioral changes

## Spec context

Follows #1529 extraction pattern.

Tracks-Bead: astro-plan-kyo7.43
Merge-Bead: astro-plan-kyo7.110

PR #1547, opened 2026-07-24
perf(inbox): eliminate per-file allocs on classify hot path

srobroek:
## Summary

- Reduces allocations on the inbox classify read path
- Cache hits for file metadata now return a pointer increment instead of a deep struct clone
- Repeated HashMap lookups, intermediate Vec allocations, and redundant key parses eliminated per classified file
- logStore notify batched via requestAnimationFrame to prevent per-log React re-renders

## Spec Context

Tracks-Bead: astro-plan-kyo7.87
Closes-Bead: astro-plan-kyo7.87
Merge-Bead: astro-plan-fcbj

### DIFF
--- a/apps/desktop/src/data/logStore.ringBuffer.test.ts
+++ b/apps/desktop/src/data/logStore.ringBuffer.test.ts
@@ -9,7 +9,7 @@
  * - `dropped` counter increments correctly.
  * - Deduplication by `id` prevents double-appending.
  */
-import { describe, it, expect, beforeEach } from 'vitest';
+import { describe, it, expect, beforeEach, vi } from 'vitest';
 import {
   appendLog,
   getLogSnapshot,
@@ -106,18 +106,24 @@ describe('logStore ring buffer', () => {
     expect(entries.length).toBe(1);
   });
 
-  it('notifies listeners on append', () => {
+  it('notifies listeners on append (async via rAF batch)', async () => {
+    vi.useFakeTimers();
+
     let callCount = 0;
     const unsub = subscribeLog(() => {
       callCount++;
     });
 
     appendLog([makeEntry(1)]);
+    // notify() schedules a requestAnimationFrame — flush it.
+    await vi.runAllTimersAsync();
     expect(callCount).toBe(1);
 
     appendLog([makeEntry(2)]);
+    await vi.runAllTimersAsync();
     expect(callCount).toBe(2);
 
     unsub();
+    vi.useRealTimers();
   });
 });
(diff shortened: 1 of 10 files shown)


srobroek (commented):
## Review: REQUEST-CHANGES

### Blocker

**logStore.ts:74-80 -- rAF batching breaks existing test**

The notify() change defers listener callbacks via requestAnimationFrame, but logStore.ringBuffer.test.ts:116 asserts synchronous notification. CI confirms: all three OS runners fail on this test (expected 1, received 0). The test must either await the rAF tick (vi.runAllTimers / vi.advanceTimersByTime), or the code must expose a flush mechanism for tests.

### Suggestions

1. normalize_text semantic shift (grouping/engine.rs:201-215): Old code used to_ascii_lowercase(); new code uses c.to_lowercase() (Unicode-aware). For FITS headers this is unlikely to cause regressions, but it is a behavioral change. If intentional, a comment noting the Unicode-aware intent would help.

2. inventory.rs:330 -- binning/gain/night unused early: The session_key parse now returns all 5 fields at line 330 but filter/binning/gain/night are only consumed much later (provenance at ~355). Not a bug, just noting the destructuring scope is wide.

### Strengths

- Perf baseline improvements are genuine: scan_root 3563ms to 1095ms, classify 3248ms to 1507ms.
- Arc-wrapping the cache value is clean -- as_deref() conversions are correct since Arc implements Deref.
- MetadataExtractError derives Clone (String fields only), so Result<Arc<_>, _> remains moka-compatible.
- frame_type_files keyed on FrameType (Copy+Hash enum) instead of String is a nice micro-optimization.

srobroek:
Addressed both blockers (commit f02487c0):

**1. logStore async contract** — took the "fix the test" path.  async via rAF is correct for  (LogPanel polls a snapshot; React re-renders on next frame, not within the same microtask). Updated `logStore.ringBuffer.test.ts` to use `vi.useFakeTimers()` + `vi.runAllTimersAsync()` to flush the pending rAF before asserting call count. Added async notification contract to module docstring.

**2. normalize_text Unicode regression** — reverted `c.to_lowercase()` back to `c.to_ascii_lowercase()`. Added a pinning comment: ASCII lowercase is intentional to preserve persisted `group_key` values (non-ASCII casing changes — e.g. Turkish İ — would split inbox items). Behavior is now identical to the original `split_whitespace().join(" ").to_ascii_lowercase()` path, just without the intermediate Vec.

srobroek:
Rebased onto origin/main; fixed unit test failures. Root cause: #1544 extracted `runConfirm` from `InboxPage.tsx` into `useInboxConfirmFlow.ts`, but the static source-scan in `InboxPage.classify.test.tsx` still globbed `InboxPage.tsx` — `indexOf('const runConfirm = useCallback(')` returned -1 on all platforms. Fixed the glob to point at `useInboxConfirmFlow.ts`. No perf-baseline.json changes needed.

srobroek:

PR #1560, opened 2026-07-24
test: property-based invariant tests for path pattern resolver and filesystem walker

srobroek:
## Summary

- Adds proptest suites covering no-panic, traversal-guard, fallback reporting, and containment invariants for the token-pattern resolver and filesystem directory walker.
- 13 new property tests across both crates (6 for patterns, 7 for fs/pathsafe), 256 cases each for in-memory tests and 64 for filesystem-backed tests.
- No bugs found — all invariants hold against current code.

## Spec Context

Tracks-Bead: astro-plan-kyo7.30
Merge-Bead: astro-plan-6kqm

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1934,6 +1934,7 @@ dependencies = [
 name = "fs_pathsafe"
 version = "0.1.0"
 dependencies = [
+ "proptest",
  "tempfile",
 ]
 
@@ -4114,6 +4115,7 @@ checksum = "17359afc20d7ab31fdb42bb844c8b3bb1dabd7dcf7e68428492da7f16966fcef"
 name = "patterns"
 version = "0.1.0"
 dependencies = [
+ "proptest",
  "safe-filename",
  "serde",
  "serde_json",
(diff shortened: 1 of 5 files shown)


srobroek (commented):
VERDICT: APPROVE — proptest suites exercise the real resolver/walker with bounded case counts (256 in-memory, 64 fs-backed); invariants hold.

Suggestions (non-blocking):
- crates/patterns/tests/proptest_patterns.rs:2,6 (dotdot invariants): the generators (arb_clean_meta_value `[A-Za-z0-9][A-Za-z0-9_\-]{0,18}`, fixed token vocab, VALID_SEPARATORS) cannot emit a `.` at all, so the `..` traversal guard at crates/patterns/src/resolver.rs:182-186 is never actually stressed — the assertions can only pass. They still guard against a regression that drops the guard AND widens the value domain, but to truly exercise resolver.rs:182-186 you would need to inject `..`/`.` through a metadata value or literal segment (which sanitize/guard should then reject). Consider one case that feeds a dotdot-bearing value and asserts the ResolveError::PathTraversal path.
- No behavioural gap otherwise; containment/idempotence/no-panic invariants are meaningful.

Issue #1068, opened 2026-07-19
Product decision: Inbox permanent narrow split (054 T018/T019) vs the new Format column


srobroek:
## Decision needed

Spec 054 (T018/T019) and current `main` have taken **contradictory directions** for the Inbox list. This surfaced as the only irreducible merge conflict when reconciling #1007.

### Spec 054 direction (T018/T019, on `feat/adaptive-detail-dock`)

Inbox becomes a **permanent narrow (~360px) split-list column** — `InboxPage` forces `'split'` placement so the FILES list is always reachable. At that width four columns don't fit, so the list carries only Path + Type + Files. The **Format column was deliberately dropped** as non-essential, with its one load-bearing bit (the master flag) folded into the Type cell.

From `InboxList.tsx` on that branch:

> Spec 054 T019: Inbox is now a PERMANENT narrow (~360px) split-list column (`InboxPage` forces `'split'` placement), so this list carries only the two essential status columns rather than the four-column table a full-width dock could afford — the Format column (fits/xisf/mixed/video) was dropped as non-essential at this width.

### Current `main` direction

`main` **added** a Format column (`formatDisplayLabel`) plus a `'format'` sort case to `InboxSortCol`, on the assumption of a full-width Inbox list.

## The conflict

These cannot both be right. Either:

- **(a)** Inbox goes permanent-narrow-split and the Format column is dropped (054 as designed), or
- **(b)** Inbox stays full-width/adaptive and keeps Format — in which case T018/T019 should be formally withdrawn from spec 054, or
- **(c)** a middle path: permanent split, but Format survives as a responsive/overflow column or folded into another cell

## Ask

Product decision required before either direction is implemented. Whichever is chosen, spec 054's tasks.md should be updated to match so the spec record stops describing a direction that isn't being built.

Extracted from #1007 (closed as superseded by #1003).

srobroek:
## Tradeoff analysis — and the premise has shifted

I went back through the evidence before this gets decided, and **T018/T019's stated justification no longer holds**. This is less balanced a fork than I originally filed it as.

### What T018/T019 were for

From the spec 054 branch's own `InboxList.tsx` header:

> Inbox is now a PERMANENT narrow (~360px) split-list column (`InboxPage` forces `'split'` placement), so **this list carries only the two essential status columns** rather than the four-column table a full-width dock could afford — the Format column was dropped as non-essential at this width.

And from the #1007 PR body:

[...]

PR #1572, opened 2026-07-24
refactor: migrate UI primitives and shared components to type-safe vanilla-extract styles

srobroek:
## Summary

Converts hand-authored BEM CSS for buttons, pills, banners, tooltips, tables, form fields, and shared layout components to vanilla-extract. Raw var(--pv-*) strings replaced with typed contracts; styles co-located with the components that own them.

## What's new

- Btn, Pill, Tooltip, Banner, KV, CoverageBar, Section, Box, Table, InfoTip, Lock all use typed VE styles
- Spacing, type-scale, and layout tokens accessible as typed uvars.* in .css.ts files instead of raw var(--pv-sp-*) strings
- TargetSearch, CommandPalette, ThemePicker, PropertyTable, FilterToolbar, SortHeader, PageTopBar, and ListPageLayout migrated to co-located VE files
- PlanPanel fully migrated (363-line detached sheet eliminated)
- Form field primitives (field-label, field-error, input) extracted to src/ui/field.css.ts for reuse across all forms

Tracks-Bead: astro-plan-kyo7.109.2
Merge-Bead: astro-plan-zf33


srobroek:
@coderabbitai review

srobroek:
## Playwright: 39 failures to 0

Each defect has the same shape: a CSS selector or spec query still names a BEM class that this branch replaced with a hashed vanilla-extract class. All of them are type-correct and pass jsdom, so the unit suite reported no failure.

### Targets table overlaid the page top bar (5 specs, ~30 failures)

`.pv-targets-page .pv-listpage__main` (`targets.css:158`) set `position: relative` to establish the containing block for `.pv-targets-table__wrap`, which uses `position: absolute`. The main column became a hashed class, so the rule stopped matching and the wrap escaped to the viewport.

Measured in the browser before the fix: the wrap is at `y=0, height=720`, its parent at `y=117`. The table's sticky sort headers render at `y=93.5`, over the top bar's Add target button at `y=83`. `document.elementFromPoint` over that button returns `BUTTON.SortHeader_root__1eqaven0`, and Playwright reports `subtree intercepts pointer events`, so each click on it times out at 30s.

Affected: `add_target_focus`, `onboarding_spotlight`, `project_lifecycle_create`, `target_detail_unclip`, `targets_planner`.

The rule now keys on `[data-testid='listpage-main']`, added to both main columns. After the fix the sort header is at `y=210.5` and hit-testing over the button returns the button.

### `uvars` bypassed the runtime type scale (3 typography specs)

`gen-ve-themes.mjs` emitted the theme-invariant tokens through `createGlobalTheme`, which mints its own hashed variables holding the rem literal. `applyFontSize` rewrites `--pv-text-*` as integer px on the root (`TEXT_SCALE_BASE_PX`, `src/data/theme.ts`), so VE-styled elements never received the override: `0.7857rem × 12px = 9.4284px`, and spec 055 SC-003 requires integers.

`uvars` now emits `var(--pv-*, <fallback>)` references, so `tokens.css` remains authoritative and the dial reaches VE-styled elements. `uvars` does not exist on `main`, so this defect is branch-only.

### Dead descendant rules in `projects.css`

The `.pv-listpage__detail--side .*` rules are live on `main`, which emits that class. This branch replaced it with `lpl.detailSide`, dropping the #1068 side-column styling with no visible error. The 9 rules now key on `data-dock="side"`, emitted by both side-panel variants. Block-level `pv-listpage__*` rules whose VE equivalents already exist are deleted.

### `PropertyTable` mono cell

Mono was applied as an inline style, so `[data-testid="property-table-cell-value"].pv-mono` matched nothing. Mono moves to a VE `cellValueMono` class plus a `data-mono` attribute.

## The conditional testid: option 1, restore the unconditional value

`data-testid` works as a query handle. Varying it by dock mode requires every present and future query to know the layout mode, and the `-side` variants had zero references in `tests/e2e/`. Dock state moves to `data-dock`.

`data-dock` holds the dock mode because commit `1a767e49d`, on this branch, established `data-variant` and `data-component` as the attribute pattern for state when it migrated BEM assertions off class names. The conditional testid diverged from the convention that same commit introduced.

Deleting the `-side` ids outright would have dropped coverage: `ListPageLayout.test.tsx` asserted dock mode through their presence in 13 places. Those now assert `data-dock` on the stable testid, which is the property they exercise.

The same reasoning retired two spec assertions on hashed class names (`pv-listpage__detail--side`, `pv-btn--primary`) in favour of the `data-dock` and `data-variant` attributes the components emit. Every assertion still checks an observable property, with no added waits, retries, or skips.

## Local verification

Run from `apps/desktop` on the merged tree (`c6df7cdcd`), which includes `main` through `aa6ab48ad`.

| Gate | Result |
|---|---|
| `npx playwright test` | 162 passed, 0 failed (163 total, 1 skipped) |
| `pnpm test` | 267 files, 2376 passed, 0 failed |
| `pnpm typecheck` | exit 0 |
| `pnpm lint` (repo root) | exit 0 |
| `check-eslint-baseline.mjs` | 722 files, 88 baselined |

The repo-root `pnpm lint` passes for the first time on this branch: `astro-plan-pwr7` (`typos` needing the key `fo`, `themes.css.ts` excluded from `end-of-file-fixer`) arrived with #1645.

An earlier run reported failures in `devSurface.release`, `ProjectDetail.target`, and `row-planning` while Playwright ran concurrently. All three pass in isolation (17/17); two were timeouts and the third missed a perf budget under load. `src/dev/` is unchanged relative to `main`.

## Unverified

Playwright ran on macOS with `retries: 0`; CI runs Linux with `retries: 2`. The pointer-interception specs depend on layout geometry, so a CI-only viewport difference remains possible, though the fix restores the containing block rather than tuning a specific size.

PR #1591, opened 2026-07-25
fix: event subscribers skip full-table replay on large libraries

srobroek:
## Summary

- Event subscribers (`lifecycle.transition.applied`, `workflow.run_completed`) now advance their replay cursor after each live event, so a subsequent broadcast-lag recovery re-reads only the missed rows instead of replaying the entire events table from the beginning.
- Replay queries are now bounded to 500 rows per page, preventing unbounded memory use on large event tables.

## Spec Context

Tracks-Bead: astro-plan-kyo7.100
Closes-Bead: astro-plan-kyo7.100
Merge-Bead: astro-plan-wosg

### DIFF
--- a/crates/app/projects/src/project_manifests.rs
+++ b/crates/app/projects/src/project_manifests.rs
@@ -431,6 +431,12 @@ pub fn spawn_workflow_run_subscriber(
             match rx.recv().await {
                 Ok(env) if env.topic == TOPIC_WORKFLOW_RUN_COMPLETED => {
                     handle_workflow_run_event(&pool, &bus, &env.payload).await;
+                    // Advance cursor so the next lag replay doesn't restart from 0.
+                    if let Ok(id) =
+                        persistence_lifecycle::repositories::events::max_event_id(bus.pool()).await
+                    {
+                        cursor = id;
+                    }
                 }
                 Ok(_) => {}
                 Err(RecvError::Lagged(n)) => {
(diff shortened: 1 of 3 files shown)


srobroek:
@coderabbitai review

srobroek:
## Two Critical findings verified live — this PR introduces a cursor-jump regression

Both CodeRabbit findings hold at head `fed5348ca`, and I checked them against `main` rather than taking them on trust.

**The defect.** After handling a broadcast, both subscribers set `cursor = max_event_id(bus.pool())`. That reads the **table-wide max**, not the id of the event just processed — so any same-topic row inserted between `recv()` and that query is jumped over and never dispatched. A later `Lagged` replay then starts *after* those rows, so they are lost rather than delayed.

Sites: `crates/app/projects/src/project_manifests.rs:437` and `crates/audit/src/stale_propagator.rs:88`.

**This is a regression, not pre-existing.** `git show origin/main` for both files returns **zero** hits for `max_event_id(bus.pool())`. Main already uses the correct shape — `stale_propagator.rs:86` is `cursor = self.replay_since(&bus, cursor).await`, advancing from what was actually processed.

**So the fix is main's own pattern**, not a new design: advance from the processed event or the replay function's return, never from a fresh table-max query. #1623 landed exactly this reasoning for the log forwarder (`drain_since` returns the new cursor and the caller assigns it), so there is a worked precedent in-tree.

**Also: this branch is 40 commits behind `main`.** Its only red check is `Rust coverage (llvm-cov)`, which is not a required context — but that run is against a 40-commit-old base, so it proves little either way. Rebase before drawing conclusions from its CI.

Tracked as `astro-plan-hyk0` (P1), linked to `astro-plan-siy1` which records the same defect family for the log forwarder. Not merging until the cursor advance is corrected.

srobroek:
Your blocker now has a fix in flight: **#1643**, based on and targeting this branch, so merging it here unblocks this PR.

The two `max_event_id(bus.pool())` cursor advances at `project_manifests.rs:435` and `stale_propagator.rs:81` read the **table-wide max**, not the row just handled, so an event inserted between the broadcast and that query is jumped and a later `Lagged` replay starts *after* it — lost, not late.

Worth knowing before reviewing #1643: the natural fix isn't available. `EventEnvelope` has no `event_id`, so a live subscriber cannot know its own row. #1643 therefore drops the live-path advance and lets the `Lagged` branch own the cursor, which is what `main` already does. That trades a replay re-walk for correctness; both handlers are idempotent, so re-doing work beats skipping it.

Both regression tests are mutation-verified. Details and the full tradeoff are in #1643 and on `astro-plan-hyk0`.

This branch is also ~40 commits behind main and will need a rebase before it can land.

srobroek:
Superseded by **#1644**, which carries this PR's content — the paged replay (`list_since_by_topic_paged`, `REPLAY_PAGE_SIZE`), which main does not have — rebuilt on current main with the blocker fixed.

Two reasons for a fresh branch rather than a rebase here: this branch was 43 commits behind, and its CI failures were inherited from that staleness (the coverage job predates #1609's `pnpm install`, so the AJV contract test could not run). A rebase would also have needed a force-push I'm not authorized for on this branch.

Note that **main has since deleted `max_event_id`** and added `rewound_cursor`, which achieves this PR's stated goal — a lag replay that doesn't restart from 0 — without the cursor jump. So the live-path advance this PR added is both incorrect and now unnecessary; the paged replay is the part worth keeping, and #1644 keeps it.

Closing in favour of #1644.

PR #1594, opened 2026-07-25
feat: shared confirm dialog and three-state list gate components

srobroek:
## Summary

- Adds a shared confirmation dialog component used for all delete/disable actions (equipment, data sources, observing sites, calibration masters, session calibration unassignment). Each now uses a consistent layout with a Cancel + action button and an optional inline error.
- Adds a shared loading/error/empty gate component for list pages, adopted in the Archive page.
- Adds a type-safe localStorage read helper with Zod schema validation, so stale or corrupted stored values fall back to defaults instead of propagating as wrong shapes.
- Adds per-scope Zod schemas for settings panes (advanced, cleanup, framing, sourceViews, naming), eliminating `as Record<string, unknown>` casts and replacing them with typed parsed results.

## Spec Context

C-5/13/24/28 from the Wave-4 TS dedup packet.

Tracks-Bead: astro-plan-kyo7.89
Merge-Bead: astro-plan-y9o2

### DIFF
--- a/apps/desktop/src/components/ConfirmModal.tsx
+++ b/apps/desktop/src/components/ConfirmModal.tsx
@@ -1,0 +1,109 @@
+// Copyright (C) 2024-2026 Sjors Robroek
+// SPDX-License-Identifier: AGPL-3.0-only
+
+/**
+ * ConfirmModal — shared confirmation dialog for destructive/reversible actions.
+ *
+ * Wraps `Modal` with the standard confirm-dialog chrome: a message body, an
+ * optional inline error, optional extra content (for disambiguation UI like a
+ * fallback-site selector), and a footer with Cancel + action button.
+ *
+ * Extracted from six duplicated confirm dialogs across Equipment, DataSources,
+ * ObservingSites, MasterArchiveFlow, and SessionDetail (C-13). All shared the
+ * same skeleton: size="sm", hideClose, footer with ghost Cancel + typed action
+ * button, optional inline field-error.
+ */
+
+import type { ReactNode } from 'react';
+import { Modal } from './Modal';
+import { Btn } from '@/ui';
+import { m } from '@/lib/i18n';
+
+export type ConfirmModalActionVariant = 'destructive' | 'danger';
+
+export interface ConfirmModalProps {
+  /** Controlled open state. */
+  open: boolean;
+  /** Called when the modal requests close (Cancel / Escape / backdrop). */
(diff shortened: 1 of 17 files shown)


srobroek:
@coderabbitai review

srobroek:
Addressed in `b6216aac` — all 3 valid.

1. **`settingsIpc.ts:139-148` whole-object rejection (major)** — fixed. `safeParse` over the whole object meant one bad field discarded every other valid persisted setting in the scope; a stale enum member left by an older build would silently reset the entire pane. Now validated per field against `schema.shape`, keeping what parses. You were also right that the doc comment was false: the schemas are `.partial()` with no `.default()`, so parsing `{}` yields an empty object, not defaults — comment corrected to say so explicitly.
2. **`ArchivePage.tsx:285` empty-vs-filtered state** — fixed with the `entries.length > 0` guard, as suggested. Confirmed against `TableStateGate`, which prefers `filteredEmpty` whenever `isEmpty` is also true.
3. **`lib/local-storage.ts:49-51` `"undefined"` persisted** — fixed. The key is now removed when `JSON.stringify` returns `undefined`, so a later read sees a clean miss rather than a corrupt value.

**On the requested regression test — worth flagging.** My first attempt asserted on a read-back (`getItem` returns null) and passed *before* the fix too, so it proved nothing: this environment's `localStorage` is Node's built-in, which rejects a non-string rather than coercing it to `"undefined"`. The tests now assert on the `setItem`/`removeItem` **calls**, which is the behaviour that actually has to hold regardless of backend coercion. Mutation-verified: reverting each fix fails the corresponding new tests (2 storage, 1 salvage).

8 new tests total. Full desktop suite: 2299 pass; tsc and biome clean.

srobroek:
Superseded by **#1625**, merged as `3257bdfba`.

This branch is an ancestor of `fix/1594-adopt-or-drop-deadcode` — verified with `git merge-base --is-ancestor` — so everything here landed with that PR, plus the work needed to make its stated goals true.

#1625 adopts the previously-dead `local-storage.ts` helpers with Zod validation at every call site, so a corrupted or stale stored value falls back to a default instead of propagating as a silently-wrong shape. It also handles the case where `JSON.stringify` returns `undefined` and `setItem` would coerce it to the literal string `"undefined"`, which a later read would treat as corrupt rather than absent.

Verified before merge: typecheck clean, 2362 tests pass, lint exit 0.

PR #1595, opened 2026-07-25
refactor: split e2e common module and mock data by domain

srobroek:
## Summary

- `crates/e2e-tests/tests/common/mod.rs` (2907L) split into `boot.rs`, `app.rs`, `helpers.rs`, `fixtures.rs`; `mod.rs` becomes a re-export hub
- `apps/desktop/src/api/mocks.ts` (3163L) domain data extracted to `src/api/mocks/` sub-files; registry shape preserved for mock-baseline gate
- `settle_first_run_redirect` deduplicated from 10 local copies to one shared impl in `common::fixtures`

Tracks-Bead: astro-plan-kyo7.44
Merge-Bead: astro-plan-q9tk


srobroek:
@coderabbitai review

srobroek:
Addressed in `24bcd554` — all 3 valid.

1. **`app.rs:616` UTF-8 slice panic** — fixed. `&dump[..4096]` panicked when byte 4096 landed inside a multi-byte character; now steps back to the nearest char boundary. Verified against a 2-byte char straddling 4096 (keeps 4095) and a 4-byte emoji case (keeps 4094), plus short-string and exact-boundary cases.
2. **`app.rs:632` `last_err` never cleared** — fixed. Cleared on a successful decode, so a transient early error no longer suppresses the `last_value` dump for the remaining poll budget. The binding lost its initializer because both match arms now assign it (clippy `unused-assignments` caught that).
3. **`boot.rs:105` stale panic messages** — fixed. The cache message promised a fallback to a full seed warm, but `expect` panics and no fallback exists; the instance-env message blamed ephemeral port binding, which moved out to `pick_port_pair`. Both now name the real failure.

Verified: `cargo test -p e2e_tests --no-run` builds, `cargo fmt --check` clean, and a before/after clippy diff confirms no new diagnostics (the remaining unused-import errors are pre-existing on this branch).

PR #1596, opened 2026-07-25
chore: move shared settings modules out of features and add settings query factory

srobroek:
## Summary

- Moves catalogue, guidance, altitude, and site settings modules from `features/targets/` and `features/settings/` to `shared/`, eliminating a circular dependency between the two feature directories.
- Adds a `settingsQueryOptions(scope)` factory in `shared/settings-queries` so call sites use a typed, cache-keyed query options object rather than ad-hoc `getQueryData` calls.

No user-visible behaviour changes.

## Spec Context

Tracks-Bead: astro-plan-kyo7.90
Merge-Bead: astro-plan-290b


srobroek:
@coderabbitai review

srobroek:
Addressed in `c11295b9`. Valid — fixed.

**`guidance-settings.ts:28` shared→feature edge.** Confirmed: `shared/planner/guidance-settings.ts` imported from `features/targets/astro/moon-avoidance` while both settings and targets consumers import `shared/planner/guidance-settings`, preserving the feature→shared→feature edge this branch exists to remove.

`moon-avoidance.ts` has **no imports of its own** — pure domain logic read from both sides of the boundary — so I moved the whole module (and its tests) to `shared/planner/` rather than splitting out selected exports, and left a re-export shim at the old path so the ~14 call sites inside the targets feature are untouched.

Verified: `tsc --noEmit` clean, biome clean, the circular-dependency gate reports the same 3 pre-existing allowlisted type-import cycles and no new ones, knip does not flag the shim as unused, and the full desktop suite passes (2300 tests).

One note: this removes the edge you identified, but **four other `shared → features` imports remain** — `shared/settings-queries.ts`, `shared/observing-sites/ObservingSites.tsx`, `shared/planner/catalogue-settings.ts`, and `shared/planner/altitude-settings.ts`. Out of scope for this finding; flagging so the boundary isn't assumed fully clean.

Rebase note: this branch's worktree held a stale rebase whose commits had different SHAs than the remote, so I reset to `origin` and reapplied the fix on the real tip instead of force-pushing over pushed work.

srobroek:
Status after #1624 merged at 30fa8554ae1f: this PR is **partially superseded, not fully**, so it stays open.

I had recorded it as a full ancestor of #1624; re-checking against the new main, that is **wrong**. `git merge-base --is-ancestor` says NOT ancestor, and 4 commits / 42 files here are still unmerged.

What #1624 DID absorb: the Moon-avoidance model move. `apps/desktop/src/shared/planner/moon-avoidance.ts` now exists on main and the old `features/targets/astro/` path is gone, so that commit's work is redundant here.

What is still ONLY on this branch:
- `apps/desktop/src/shared/settings-queries.ts` — the `settingsQueryOptions` factory, absent from main.
- the `ObservingSites`/`SiteLocation` move into `shared/observing-sites/`, plus their missing `data-testid` lint fix.

So the residual is the settings-query refactor, not the planner move. Next step is a rebase onto main — 33 commits behind — dropping the now-redundant planner commit and keeping the settings work. Expect a conflict in `scripts/eslint-alm-baseline.txt`, which #1624 and #1625 both regenerated: regenerate it with `node scripts/check-eslint-baseline.mjs --generate` from the repo root rather than merging hunks, since it is line-keyed.

srobroek:
**Rebase attempted; stopping short of forcing it, because the residual is smaller than my previous comment claimed.**

Correcting myself again. I said the residual was "the settings-query refactor + the ObservingSites move." Checking file-by-file against current main, #1624 absorbed **more** than that:

| path | on main? |
|---|---|
| `shared/planner/guidance-settings.ts` | **yes** |
| `shared/planner/altitude-settings.ts` | **yes** |
| `shared/planner/catalogue-settings.ts` | **yes** |
| `features/settings/ObservingSites.tsx` | **yes** (kept in `features/`) |
| `shared/settings-queries.ts` | no |
| `shared/observing-sites/ObservingSites.tsx` | no |

So a straight `git rebase origin/main` conflicts in **15 files** — every one of them a move main has already made by a different route. Replaying those hunks would mean re-litigating #1624's layering decisions through conflict markers, which is how a rebase silently reverts a landed choice. I aborted rather than resolve them, and the branch is back at `c11295b95` untouched.

**What's genuinely left is two independent decisions, not one rebase:**

1. **`settingsQueryOptions` factory** (39 lines, `shared/settings-queries.ts`). main has no equivalent — `settingsQueryOptions` appears nowhere. But it landed here with **zero callers**, by design: "enables incremental migration without a flag day." An export with no callers is exactly what the dead-code cleanups in #1598/#1624 spent effort removing. It should land **with its first caller**, or not at all.

2. **`ObservingSites` → `shared/observing-sites/`.** main deliberately kept it in `features/settings/`. That's a layering disagreement with a landed decision, so it needs a call on the merits, not a conflict resolution.

Recommendation: **close this PR** and carry item 1 forward as its own change bundled with a real consumer. Its lint fix (missing `data-testid`) is already moot — main's copy has them.

I'm leaving it open for your call rather than closing a PR whose remaining value is a judgment question.

Side finding while checking this, filed separately: `pnpm knip` exits 1 on clean main with 6 unused exports in `src/api/mocks/`, but knip isn't wired into CI or the lint chain — installed and red but unenforced.

srobroek:
Closing as **superseded by #1651** (`d5b323e45d19`).

Both halves of this PR's intent are now on main by other routes:

- **The observing-sites move** landed via **#1624**. Verified: `features/targets/observing-sites/` is gone from main and the cluster sits at `shared/observing-sites/`, with all six consumers importing from there. So there was nothing left to move — an agent I dispatched to redo it correctly reported the premise was already stale and changed nothing rather than inventing work.
- **The `settingsQueryOptions` factory** landed via #1651, rebuilt rather than replayed. The version here hardcoded `['settings', scope]`, bypassing the central `data/queryKeys.ts` registry that exists to enable prefix-based invalidation; #1651 adds a proper `settings` namespace there and keys off it. It also lands with **five migrated call sites** instead of zero, and — the part that actually matters — **invalidation wired on write**. With `staleTime: Infinity` and no invalidation, the factory here would have been worse than the code it replaced.

Also worth recording: rebasing this branch was not viable. It conflicted in **15 files**, every one a move main had already made by a different route, so resolving them would have meant re-litigating landed decisions through conflict markers.

Nothing is lost. Closing.

PR #1600, opened 2026-07-26
ci: attest build provenance for published bundles

srobroek:
Adds [GitHub artifact attestations](https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations) so every published bundle carries a signed statement of which workflow run, at which commit, produced those exact bytes.

```console
$ gh attestation verify PlateVault_x64-setup.exe  -R platevault/platevault
$ gh attestation verify PlateVault_amd64.AppImage -R platevault/platevault
```

## The attestation is split across two jobs, because signing changes the artifact

This is the part worth reviewing carefully. `sign-windows` replaces the `.exe`/`.msi` on the release with Authenticode-signed copies via `gh release upload --clobber`. **Signing rewrites the file, so its digest changes.** An attestation made in `build` would therefore describe bytes that no longer exist on the release — and the failure would be silent, because both jobs would still go green and `gh attestation verify` would only fail later, for a user.

So:

| Job | Attests |
|---|---|
| `build` | `.deb`, `.rpm`, `.AppImage`, `.dmg` — plus `.exe`/`.msi` **only when Authenticode signing is off** |
| `sign-windows` | the **signed** `.exe`/`.msi`, before the clobbering upload |

The exclusion in `build` is gated on the same `ENABLE_WINDOWS_SIGNING` variable as the signing job itself, so the two cannot disagree about which leg owns those files. With signing off, nothing clobbers them and `build` attests them normally.

Both jobs gain `id-token: write` + `attestations: write`. `actions/attest` is SHA-pinned to `f7c74d2` (v4.2.0).

## The artifactPaths conversion

`tauri-action` emits `artifactPaths` as a **JSON array string**, not a newline-separated list. Confirmed in its source rather than inferred — `src/index.ts` does:

```js
core.setOutput('artifactPaths', JSON.stringify(artifacts.map((a) => a.path)))
```

`subject-path` wants newlines, so a conversion step handles it. That also avoids hardcoding bundle filenames, which carry the version and would drift every release. Filtered out: `.sig` sidecars (signatures over the bundles, not artifacts in their own right) and `latest.json` (a pointer regenerated per release, so a provenance claim over it ages out immediately).

## It fails loud on an empty subject set

Without this, a future change to `tauri-action`'s output shape would leave bundles unattested while the job stayed green — the release would *look* provenance-covered when it was not. The one legitimate empty case (Windows with signing enabled, whose only bundles are the `.exe`/`.msi` that `sign-windows` attests) is handled explicitly rather than by suppressing the check.

## This does not touch minisign updater signing

Those `.sig` files are what the built-in auto-updater checks against the app's embedded public key, offline and without contacting GitHub. The attestation records where a build came from; the `.sig` establishes that an update is genuine. Neither replaces the other. The README's existing code-signing paragraph was extended rather than duplicated.

## Validation

- `actionlint` clean.
- The `jq` filter and fail-loud branch were simulated against realistic `artifactPaths` payloads for **all four** combinations: Windows + signing on (correctly empty, deferred to `sign-windows`), Windows + signing off (attests `.exe` + `.msi`), Linux (`.deb` / `.AppImage` / `.rpm`), and a broken-output case (exits 1).
- This repo's own `pre-commit` suite passes on the changed files (yaml, gitleaks, typos, whitespace).
- Attestation needs GitHub OIDC and cannot be exercised locally, so **the first release after merge is the real proof** — worth watching that run, particularly whether the Windows digests line up if `ENABLE_WINDOWS_SIGNING` is ever turned on.

Tracks-Bead: astro-plan-ey62
Merge-Bead: astro-plan-bn9f

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -77,6 +77,10 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: write # tauri-action publishes the release and uploads bundles
+      id-token: write # OIDC identity that signs the attestation
+      attestations: write # publish the attestation to this repo
     steps:
       - uses: actions/checkout@v4
 
@@ -136,6 +140,7 @@ jobs:
       #   - assetNamePattern renamed to artifactNamePattern: we don't use it.
       #   - Tauri v1/unstable support dropped: we are on Tauri 2 stable.
       - name: Build, sign, and upload Tauri bundles
+        id: tauri
         uses: tauri-apps/tauri-action@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


srobroek:
@coderabbitai review

PR #1603, opened 2026-07-27
fix: command palette finds targets by alias again

srobroek:
## Summary

- Alias-only queries in the command palette, such as "Caldwell 20" for NGC 7000, returned no results. The palette fetched the whole target catalog once per open and filtered it client-side against designation and label only; aliases are no longer serialized on `TargetListItem` (GF-11 / DS-16), so no client-side filter can see them.
- The palette now queries the `target.list(search)` endpoint on its existing 200 ms debounce, alongside `search_global`. That endpoint matches designation, label, and aliases in-process over the cached catalog (`crates/app/targets/src/target_management/list.rs:60`).
- `buildTargetResults` ranks and shapes the returned rows instead of filtering them; its matcher parameter is gone.
- The open-time full-catalog fetch and its 60 s TTL cache are removed: no IPC call fires until the user types.
- `scripts/eslint-alm-baseline.txt` records the pre-existing `alm/require-root-testid` violation at its shifted line number (280 to 215).

## Tests

`apps/desktop/src/app/CommandPalette.test.tsx`: the `buildTargetResults` block asserts ranking and that an alias-only row survives; a new `CommandPalette server-side target search` block asserts the query reaches `target.list`, the alias-only result renders, the call is debounced by 200 ms, and no fetch happens before the user types. Reverting `CommandPalette.tsx` to its base version fails 11 of the file's 22 tests.

## Verification

| Command | Result |
| --- | --- |
| `pnpm --dir apps/desktop exec vitest run src/app/CommandPalette.test.tsx` | 22 passed |
| `pnpm --dir apps/desktop run test` | 257 files, 2300 tests passed |
| `pnpm --dir apps/desktop exec tsc --noEmit` | clean |
| `pnpm --dir apps/desktop run lint` | biome, eslint baseline, tokens, i18n, mock baseline pass; `check-circular.mjs` fails on 3 cycles (TargetSearch, PlanPanel x2) that also fail on base commit 642328d5, unrelated to this diff |

## Spec Context

Tracks-Bead: astro-plan-ticm
Merge-Bead: astro-plan-sxay

Source bead: astro-plan-kyo7.111

### DIFF
--- a/apps/desktop/src/app/CommandPalette.test.tsx
+++ b/apps/desktop/src/app/CommandPalette.test.tsx
@@ -2,12 +2,8 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 
 /**
- * CommandPalette T008 integration tests — alias-aware target search routing.
- *
- * Logic-layer tests validate:
- *   - search results are filtered/routed correctly
- *   - target results produced by searchGlobal have the right shape
- *   - the navigate call receives the verbatim route from the result
+ * CommandPalette tests — target search wiring, PAGES route guards, and rendered
+ * smoke coverage.
  *
  * Rendered smoke tests (#581 review) mount the real palette with a local
  * ResizeObserver/scrollIntoView stub (cmdk + @base-ui-components/react/dialog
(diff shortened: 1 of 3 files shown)


srobroek:
@coderabbitai review

PR #1542, opened 2026-07-24
test(e2e): replace .pv-* class selectors with data-testid / data-kind

srobroek:
## Summary

- e2e tests locate UI elements by stable identity instead of CSS class names, decoupling tests from styling changes (prerequisite for CSS codegen tooling)
- 87 unique `.pv-*` selectors migrated across 30 Playwright spec files and 4 Rust journey files; matching `data-testid` / `data-kind` attributes added to ~87 source elements
- Two regression guards added: `alm/require-root-testid` ESLint rule (100 grandfathered violations, shrink-only baseline) and `check-pv-selector-ratchet.sh` (sealed at zero, covers Rust tests)

## Spec Context

Tracks-Bead: astro-plan-kyo7.107
Merge-Bead: astro-plan-txnd

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1604, opened 2026-07-27
ci: enforce lifecycle-string and CSS-selector ratchets in CI

srobroek:
## Summary

- Wires two existing ratchet scripts into CI. Both ran only locally, so raw `.lifecycle == "..."` comparisons and `.pv-*` e2e selectors reached main unchecked.
- Moves four literal-pattern lint checks out of a vitest guard and a shell script into Biome GritQL plugins.

Tracks-Bead: astro-plan-z4p8
Merge-Bead: astro-plan-qkmp

Source audit bead: astro-plan-kyo7.115. Run: run-wjrz-20260727.

## Changes

| Audit action | Result |
| --- | --- |
| 1. Fix check-lifecycle-strings.sh comment exclusion | Already correct at HEAD (`scripts/check-lifecycle-strings.sh:21`). No change needed. |
| 2. Wire `just lifecycle-strings` into CI | `.github/workflows/ci.yml:405` plus `just check` |
| 3. Wire check-pv-selector-ratchet.sh into justfile + CI | `.github/workflows/ci.yml:414` plus `just check`. #1542 did not carry the wiring. |
| 4. Migrate SC-001 + SC-005 to Biome GritQL | `apps/desktop/biome-plugins/no-raw-invoke.grit`, `no-retired-commands-wrapper.grit`. SC-006 stays in vitest. |
| 5. Migrate check-tokens.sh checks 3+4 | `apps/desktop/biome-plugins/no-legacy-token-namespace.grit`, `no-bare-pv-radius.grit`. CSS checks 1+2 and node checks 5-7 unchanged. |

Check numbers 5-7 in `check-tokens.sh` keep their original values because docs and specs cite them.

## Ratchet proofs

Each wired or migrated check was made to fail on a deliberate violation, then reverted to green.

| Check | Violation | Observed |
| --- | --- | --- |
| lifecycle-strings | `p.lifecycle == "active"` in `crates/app/core/src/lib.rs` | exit 1, 1 raw lifecycle string comparison found |
| pv-selector (TS) | `.pv-probe-thing` in `tests/e2e/add_target_focus.spec.ts` | exit 1, 1 `.pv-*` class selector found |
| pv-selector (Rust) | `By::Css(".pv-probe-thing")` in `crates/e2e-tests/tests/archive_journeys.rs` | exit 1, same message |
| SC-005 | static import, dynamic `import()`, and `vi.mock` of `@/api/commands` | 1 diagnostic each |
| SC-001 | `invoke('probe.cmd')` and `invoke<number>('probe.cmd')` | 1 diagnostic each |
| tokens 3 | `var(--pv-color-x)`, `var(--mantine-color-body)` | 1 diagnostic each |
| tokens 4 | `var(--pv-radius)` as string and as template literal | 1 diagnostic each; `var(--pv-radius-md)` passes |

Exemptions verified by the clean tree passing: `src/api/ipc.ts`, `src/features/sessions/sessionsGroupsIpc.ts`, and `src/features/calibration/calibrationHandoffIpc.ts` all contain raw `invoke('...')` calls and produce zero diagnostics.

## Gates

- `just lifecycle-strings`, `just pv-selector-ratchet`: pass.
- `bash scripts/check-tokens.sh`: pass.
- `biome lint apps/desktop` (646 files): 0 errors, 33 warnings, identical to the pre-change baseline.
- `biome format apps/desktop`: clean.
- `npx vitest run src/api/ipc-boundary.guard.test.ts`: 2 passed.
- `pnpm --filter @astro-plan/desktop run typecheck`: pass.
- `pnpm --filter @astro-plan/desktop run lint`: FAILS at `check-circular.mjs` with 3 cycles (`TargetSearch.tsx`, `PlanPanel.tsx` twice). Pre-existing: the identical failure reproduces on stashed HEAD. Not caused by this branch.

No Rust source changed, so cargo gates were not run.

## Biome 2.5.5 GritQL findings

- A root `plugins` entry applies to every linted file and cannot be narrowed. Per-path exemptions need an `overrides` entry with negated `includes` globs carrying the `plugins` key. Declaring a plugin inside an override is additive to root, not a replacement.
- A static import's module source is not a `string()` node, so SC-005 needs a quote-anchored regex branch for `from '@/api/commands'`.
- Template literals are not reached by `string()`, and no template node kind name compiles in this dialect. The token plugins add a regex branch excluding quote and separator characters, which yields one diagnostic per violation rather than one per enclosing node.
- The token plugins exclude `*.test.ts` and `*.test.tsx`, matching the original `grep -v '.test.'` filter: `NamingStructure.r4.test.ts` asserts on the literals they ban.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,17 +233,22 @@ jobs:
           } >> "$GITHUB_OUTPUT"
           echo "changes: rust=$rust frontend=$frontend frontend_full=$frontend_full docs_only=$docs_only scripts=${{ steps.filter.outputs.scripts }}"
 
-  # Two lanes share this matrix (Rust and frontend), each with its own step-level
-  # gate, so it is NOT split into two jobs: whenever either lane is active every
-  # OS leg still runs real work — `Rust tests` + `Generated contract/binding
-  # drift` are OS-agnostic under the Rust lane, as are the frontend lint / unit
-  # test / typecheck steps under the frontend lane. The job-level gate below
-  # therefore fires only when BOTH lanes are idle, which is exactly the case
-  # where the job had nothing to do and must not report `success`.
+  # Three lanes share this matrix (Rust, frontend, scripts), each with its own
+  # step-level gate, so it is NOT split into separate jobs: whenever any lane is
+  # active every OS leg still runs real work — `Rust tests` + `Generated
+  # contract/binding drift` are OS-agnostic under the Rust lane, as are the
+  # frontend lint / unit test / typecheck steps under the frontend lane, and the
+  # ratchet steps under the scripts lane. The job-level gate below therefore
+  # fires only when ALL lanes are idle, which is exactly the case where the job
+  # had nothing to do and must not report `success`.
+  #
+  # `scripts` must appear here as well as on the ratchet steps: those steps gate
+  # on it, so without it a scripts-only PR skips this job entirely and silently
+  # bypasses every ratchet it was meant to trigger.
   integration:
     name: Unit + integration (L1+L2) — ${{ matrix.os }}
     needs: changes
-    if: needs.changes.outputs.rust == 'true' || needs.changes.outputs.frontend == 'true'
+    if: needs.changes.outputs.rust == 'true' || needs.changes.outputs.frontend == 'true' || needs.changes.outputs.scripts == 'true'
     strategy:
       fail-fast: false # report each platform distinctly (FR-011)
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -402,6 +402,22 @@ jobs:
         if: runner.os == 'Linux' && (needs.changes.outputs.rust == 'true' || needs.changes.outputs.scripts == 'true')
         run: just hot-read
 
+      # Lifecycle string-comparison ratchet: sealed at zero raw
+      # `.lifecycle == "..."` comparisons in production Rust (#1534, #1556).
+      # Local-only until now, so raw comparisons could reach main through CI.
+      # Pure grep, no compile — same Linux-only gating as the ratchets above.
+      - name: Lifecycle string ratchet
+        if: runner.os == 'Linux' && (needs.changes.outputs.rust == 'true' || needs.changes.outputs.scripts == 'true')
+        run: just lifecycle-strings
+
+      # .pv-* selector ratchet: sealed at zero class selectors in e2e test
+      # files (Playwright TS under tests/e2e/ and Rust journeys under
+      # crates/e2e-tests/), which the eslint alm/require-root-testid rule
+      # cannot reach. Gated on frontend too, because tests/e2e/ is TS.
+      - name: .pv-* selector ratchet
+        if: runner.os == 'Linux' && (needs.changes.outputs.rust == 'true' || needs.changes.outputs.frontend == 'true' || needs.changes.outputs.scripts == 'true')
+        run: just pv-selector-ratchet
+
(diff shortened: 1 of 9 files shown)


srobroek:
@coderabbitai review

srobroek:
@coderabbitai full review

srobroek:
@coderabbitai full review

srobroek:
Addressed in `6c17ae79`. Two of three valid, one rejected.

**1. `ci.yml` — integration job skips scripts-only changes (major): VALID, fixed.** Confirmed the ratchet steps gate on `needs.changes.outputs.scripts` while the enclosing job at :246 did not, so a scripts-only PR skipped the job and bypassed both ratchets. Added `scripts` to the job condition and updated the block comment (it said "two lanes" / "BOTH lanes").

**2. Grit template-literal gaps (minor): VALID, fixed.** Verified empirically with probe files rather than by reading:
- `no-raw-invoke.grit`: `invoke('cmd')` was flagged; `invoke(\`cmd\`)` and `invoke<T>(\`cmd\`)` were **not**. Added a regex branch — all three now report exactly once.
- `no-retired-commands-wrapper.grit`: `import('@/api/commands')` was flagged; `import(\`@/api/commands\`)` was **not**. Added a backtick to the quote class — all three specifier forms now caught.

No false positives introduced: `biome check src` reports 0 plugin diagnostics across 646 files (33 pre-existing `noNonNullAssertion` warnings, unrelated — tracked as `astro-plan-x0n7`).

**3. `biome.json` overrides must be mutually exclusive (major): REJECTED — not how Biome 2.5.5 behaves.** The claim rests on a web search saying only the first matching override applies. Tested directly on Biome 2.5.5:
- `src/api/ipc.ts` (negated in override 1) → 0 `no-raw-invoke` diagnostics, so override 1's exclusions work.
- A probe file at `src/api/` → **both** `no-bare-pv-radius` and `no-legacy-token-namespace` from override 2 fire.

Both overrides apply to the same file, so override 2's plugins are not shadowed and no scope change is needed.

srobroek:
## Both findings verified as already-addressed, not defects

I checked each against the code rather than acting on them, and both fail.

### Override exclusivity — not a defect

**Biome 2.5.5 applies all matching overrides cumulatively, plus root `plugins`. It does not stop at the first match.** So the two blocks matching `src/**` are additive, and no plugin is inert.

Probed directly, and independently re-confirmed:

| Probe | Result |
|---|---|
| Ordinary `src/**` file violating all three plugins | all three fire (2× SC-001, 2× spec 022, 1× spec 028) |
| Same content as `*.test.ts` | only SC-001 — block 1's test exclusion binds correctly |
| `src/api/ipc.ts` unmodified | clean — block 0's IPC exclusion binds |
| **`src/api/ipc.ts` + injected `var(--pv-radius)`** | **spec 028 fires, SC-001 does not** |

That last row is decisive: a file *excluded from block 0* still receives block 1's plugins. First-match semantics would make that impossible.

Making the scopes "mutually exclusive" would be a no-op at best, and risks **removing** coverage.

### Template literals — already fixed

Commit `6c17ae79c` ("close Grit backtick gaps") added the backtick branches. `no-raw-invoke.grit:18` reads `or { contains string(), r"[^\n]*\`[^\`\n]*\`[^\n]*" }`, not bare `contains string()`. Verified present. CodeRabbit reviewed a pre-`6c17ae79c` revision.

Probes: `invoke<number>(\`probe_cmd\`)` fires SC-001; `\`var(--pv-radius)\`` fires spec 028; `import(\`@/api/commands\`)` fires SC-005.

### Third finding

`integration`'s `if:` already includes `scripts`, with a comment explaining exactly why omitting it would let a scripts-only PR bypass every ratchet.

No code change. Resolving all three as already-addressed.

PR #1605, opened 2026-07-27
fix: circular-dependency lint gate no longer fails on known type-only cycles

srobroek:
## Summary

- The circular-dependency lint gate reported its three known type-only cycles as new on every branch, including unmodified `main`, so `pnpm lint` was red everywhere.
- Root cause: `madge` writes ANSI styling to its human-readable output even with `--no-color`. The gate stripped `^\d+\)` from those lines and compared them to the allowlist with `Set.has`; the escape sequences defeated both, so no allowlist entry ever matched.
- Fix: `apps/desktop/scripts/check-circular.mjs` now reads `madge --circular --json` and joins each cycle array with `' > '`. JSON is the only ANSI-free machine format madge offers. The allowlist is unchanged and matching stays exact, so an unlisted cycle still fails.
- Secondary defect fixed: with `apps/desktop/node_modules` absent, `execSync` threw, `err.stdout` was empty, and the gate printed "0 new cycles" and exited 0 — a blind pass. Unparseable output now fails with an install hint.
- The three allowlisted reverse edges were re-confirmed `import type` only (`useTargetSearch.ts:31`, `PlanRootPicker.tsx:12`, and the equivalent import in `PlanDestructiveControl.tsx`). No component refactor needed.

## Verification

| Direction | Result |
| --- | --- |
| Gate on unmodified tree | exit 0, `0 new cycles (3 allowlisted type-import cycle(s))` |
| Gate with an injected real runtime cycle (two modules importing each other's values) | exit 1, names `zz-cycle-a.ts > zz-cycle-b.ts`; fixtures reverted |
| `pnpm --filter @astro-plan/desktop run lint` | exit 0 |

## Spec Context

Tracks-Bead: astro-plan-s5ae
Tracks-Bead: astro-plan-lyq1
Merge-Bead: astro-plan-rw23

### DIFF
--- a/apps/desktop/scripts/check-circular.mjs
+++ b/apps/desktop/scripts/check-circular.mjs
@@ -76,6 +76,17 @@ let cycles;
 try {
   cycles = JSON.parse(madgeOut);
   if (!Array.isArray(cycles)) throw new Error('expected an array');
+  // Validate every entry here, not at the `.join` below: `[null]` satisfies
+  // `Array.isArray` and would then throw an uncaught TypeError, losing the
+  // controlled diagnostic this branch exists to print.
+  const bad = cycles.findIndex(
+    (c) => !Array.isArray(c) || c.some((m) => typeof m !== 'string'),
+  );
+  if (bad !== -1) {
+    throw new Error(
+      `entry ${bad} is not an array of strings: ${JSON.stringify(cycles[bad])}`,
+    );
+  }
 } catch (err) {
   process.stderr.write(
     `CIRCULAR DEP GATE FAILED — could not parse madge --json output: ${
--- a/apps/desktop/scripts/check-circular.mjs
+++ b/apps/desktop/scripts/check-circular.mjs
@@ -39,20 +39,26 @@ const ALLOWLIST = new Set([
 const listMode = process.argv.includes('--list');
 
 let madgeOut;
+/** @type {string} */
+let madgeErr = '';
 try {
   madgeOut = execSync(
     'node_modules/.bin/madge --circular --json --no-spinner --extensions ts,tsx src',
-    { cwd: ROOT, encoding: 'utf8', stdio: ['ignore', 'pipe', 'ignore'] },
+    { cwd: ROOT, encoding: 'utf8', stdio: ['ignore', 'pipe', 'pipe'] },
   );
 } catch (err) {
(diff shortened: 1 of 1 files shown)


srobroek:
@coderabbitai review

srobroek:
@coderabbitai full review

srobroek:
Addressed in `5d2befc6`.

**Preserve Madge's failure output** — valid, fixed. `stdio` stderr changed from `ignore` to `pipe` and captured into `madgeErr`; the parse-failure branch now prints madge's stderr and stdout when present, and falls back to the `pnpm install` hint only when madge emitted no JSON at all.

Verified all four paths locally:
- clean tree → exit 0, `0 new cycles (3 allowlisted)`
- injected real cycle → exit 1, names `__cycle_a.ts > __cycle_b.ts`
- madge exits 1 with a stderr diagnostic → exit 1, prints `Error: Cannot find module 'typescript'`
- madge absent → exit 1, prints the exec error plus the install hint

srobroek:
@coderabbitai full review

srobroek:
Addressed in `dc42e94d`.

**Validate each cycle entry before joining it** — valid, fixed. Reproduced first: with madge emitting `[null]`, the script died with an uncaught `TypeError: Cannot read properties of null (reading 'join')` and a raw stack trace, bypassing the controlled diagnostic entirely.

Now every entry is checked as an array of strings inside the existing `try`, so malformed output routes through the same error path and names the offending index. Verified:
- `[null]` → exit 1, `entry 0 is not an array of strings: null`
- `[[1,2]]` → exit 1, `entry 0 is not an array of strings: [1,2]`
- `{"a":1}` → exit 1, `expected an array`
- `[]` → exit 0 (valid empty result still passes)
- real tree → exit 0 clean, exit 1 on an injected cycle

Note for context: the previous commit's Windows fix was itself found this way. Once stderr stopped being discarded, the Windows leg revealed `'node_modules' is not recognized` — cmd.exe cannot execute the bin shim, so this gate had **never** run on Windows and was passing vacuously with no cycle checking at all. It now runs there via `createRequire` + `spawnSync`, and the Windows CI leg passes.

srobroek:
@coderabbitai full review

srobroek:
@coderabbitai full review

PR #1569, opened 2026-07-24
test: add mock-mode coverage for the software update UI

srobroek:
## Summary
- Adds a test injection seam (`window.__PV_TEST__.updateState`) so Playwright specs can exercise all update-check UI states (up to date, update available, check failed) without a real update server
- Adds a dev-build-only env var (`PV_E2E_VERSION_OVERRIDE`) that spoofs a lower app version so the updater triggers against a fixture endpoint; absent from release builds at compile time

## Spec Context
spec 051 US10 / J17

Tracks-Bead: astro-plan-kyo7.106
Closes-Bead: astro-plan-kyo7.106
Merge-Bead: astro-plan-desz

PR #1606, opened 2026-07-28
feat(scripts): detect declined CodeRabbit reviews instead of reading them as clean

srobroek:
## Why

The serial CodeRabbit review cycle in `run-wjrz-20260727` recorded six PRs as
*"reviewed, re-swept, zero inline findings"* — #1569, #1572, #1600, #1603, #1604,
#1605. CodeRabbit had never reviewed any of them. Five were rate-limited and one
was skipped as a draft; in every case the bot posted a `✅ Action performed —
Review finished` acknowledgement alongside the refusal, and that marker was read
as success.

The cost was real. Re-running the reviews properly found that #1604 carried three
findings, one of them a CI gate that skipped both ratchets it had just been added
to enforce, and #1605 carried a diagnostic-masking bug whose fix then exposed that
the circular-dependency gate had never worked on Windows at all.

## What

`scripts/check-coderabbit-review.sh` treats a review body at the current head SHA
as the only evidence that a review happened, and names the reason otherwise:

| Status | Meaning |
|---|---|
| `REVIEWED` | real review for the current head; inline finding count reported |
| `RATE_LIMITED` | bot declined — review limit or fair-usage notice |
| `SKIPPED_DRAFT` | bot declined — draft PR, auto-review off for drafts |
| `STALE` | reviewed, but an older commit than head |
| `PENDING` | review in progress |
| `NOT_REVIEWED` | no review, no notice |

Exit 0 only when every scanned PR is `REVIEWED`. When a non-`REVIEWED` PR also
carries a `Review finished` ack, the output says so explicitly, so the trap is
visible rather than silent.

For a rate limit it reports when the window reopens as an **absolute UTC time**.
The bot's own `Next review available in: N minutes` is relative to when the notice
was posted, so the figure is stale by the time anyone reads it. The figure is read
from the notice that quotes it, not the last comment — that is usually the bare
acknowledgement and carries no figure.

```
$ bash scripts/check-coderabbit-review.sh 1596 1603 1605
PR       STATUS         FINDINGS   DETAIL
#1596    REVIEWED       1          1 inline finding(s)
#1603    RATE_LIMITED   0          bot declined: review limit — review window has reopened (elapsed 596 min ago) — safe to re-trigger (bot posted 'Review finished' — NOT evidence of a review)
#1605    STALE          1          review exists but not for head fb6c17a7 — re-trigger for the new commits
```

`--open` scans every open PR; `--json` emits one object per PR for scripting.

## Verification

- 18/18 bats tests pass (`scripts/tests/check-coderabbit-review.bats`), covering
  each classification, both wait-window directions, hour-denominated figures, and
  the exact notice shape that caused the original misread.
- `shellcheck` clean.
- The guard is load-bearing, verified by mutation: making the script trust a
  `Review finished` ack fails tests 3 and 9.
- Run against the live API on all ten PRs from the original cycle: it reproduces
  the failure exactly — the four genuinely-reviewed PRs report `REVIEWED` with
  their correct finding counts, and all six the handover called clean are flagged
  as declined.

Merge-Bead: astro-plan-6n93

### DIFF
--- a/scripts/check-coderabbit-review.sh
+++ b/scripts/check-coderabbit-review.sh
@@ -1,0 +1,212 @@
+#!/usr/bin/env bash
+# CodeRabbit review-status scanner.
+#
+# Answers one question per PR: has CodeRabbit actually reviewed the current head,
+# or does it only LOOK reviewed? The distinction matters because CodeRabbit posts
+# an "Action performed / Review finished" acknowledgement even when it refused to
+# review — on a rate limit or a draft skip. Reading that marker as success is how
+# a review cycle records unreviewed PRs as "zero findings".
+#
+# A PR is REVIEWED only when a review body from coderabbitai[bot] exists for the
+# current head SHA. Acknowledgements never count as evidence.
+#
+# Statuses:
+#   REVIEWED      real review present for head; inline finding count reported
+#   RATE_LIMITED  bot declined — "Review limit reached" / fair-usage notice.
+#                 Reports when the window reopens as an absolute UTC time: the
+#                 bot's own "in N minutes" is relative to when it was posted and
+#                 is therefore stale on read.
+#   SKIPPED_DRAFT bot declined — draft PR and auto-review is off for drafts
+#   STALE         review exists, but only for an older commit than head
+#   PENDING       review in progress
+#   NOT_REVIEWED  no review and no explanatory notice
+#
+# Exit 0 when every scanned PR is REVIEWED; 1 otherwise. Only REVIEWED is a
+# green light to adjudicate findings — every other status means re-trigger.
+#
+# Usage:
+#   bash scripts/check-coderabbit-review.sh 1604 1605     # specific PRs
+#   bash scripts/check-coderabbit-review.sh --open        # every open PR
+#   bash scripts/check-coderabbit-review.sh --json 1604   # machine-readable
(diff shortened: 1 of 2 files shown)


srobroek:
Superseded by **#1622**, merged as `af7649dc5`.

`scripts/check-coderabbit-review.sh` is now on `main`, so a shepherd working from a main-based checkout can run its own mandatory gate. That was the structural problem: every shepherd today had to `git show` the script off a feature branch first, and at one point that extraction was itself denied on provenance grounds, blocking a drain entirely.

#1622 carries three fixes this PR did not have, all found during review:

1. **The rate-limit parser matched the wrong phrase.** It looked for `next review available in` while CodeRabbit writes "next **included** review **will be** available in", so the detail line claimed the window had *reopened* while quota was exhausted. That false green light cost four wasted re-triggers.
2. **Critical: the bats suite was BSD-only.** Six fixtures used `date -u -r <epoch>`; on GNU coreutils `-r` means `--reference=FILE`, so every timestamp fixture would have failed on Linux CI for reasons unrelated to the scanner.
3. **A `gh pr list` failure was swallowed by `mapfile`**, so an auth or network error surfaced as "wrong arguments".

Plus `--open` no longer truncates silently at 100 PRs.

Tracked on `astro-plan-yybv`, now closed.

PR #1610, opened 2026-07-28
fix: intermittent inbox plan-listener test failure

srobroek:
## Summary

- `plan_listener::tests::applied_plan_transitions_to_resolved` failed intermittently with `plan link should be deleted after resolution`, breaking the coverage and integration legs on unrelated PRs (seen on #1605 and #1609).
- `transition_via_plan_id` does two separate writes: update the inbox item state, then delete the `inbox_plan_links` row. `wait_item_state` returning means only the first has landed, so asserting on the link immediately afterwards races the delete and fails whenever the runner is loaded enough to observe the two apart.
- Now polls for the deletion using the module own `poll_until` helper, which exists for exactly this reason.

Not a product bug: the listener ordering is correct and the link is always deleted. The test assumed an atomicity it never had.

## Verification

30 consecutive runs of the test pass; `cargo clippy -p app_core_inbox --all-targets -- -D warnings` clean.

Merge-Bead: astro-plan-8b8c

### DIFF
--- a/crates/app/inbox/src/plan_listener.rs
+++ b/crates/app/inbox/src/plan_listener.rs
@@ -599,6 +599,29 @@ pub(crate) mod tests {
         .await;
     }
 
+    /// Poll until the `inbox_plan_links` row for `item_id` is gone, panic after
+    /// 2 s.
+    ///
+    /// `transition_via_plan_id` updates the item state and deletes the link as
+    /// two separate writes, so observing the new state does NOT imply the link
+    /// is already gone — asserting on it directly after `wait_item_state` is a
+    /// race that fails intermittently under load.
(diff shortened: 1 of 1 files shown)


srobroek:
**Note on the remaining `Rust coverage (llvm-cov)` failure.**

This PR and #1609 fix two independent defects that both break the same check, so **neither PR can show a green coverage leg on its own**:

- **#1609** — the coverage job installs no Node deps, so `spec062_portable_contracts` dies with `Cannot find module 'ajv/dist/2020'`.
- **#1610** (this PR) — `plan_listener::applied_plan_transitions_to_resolved` races the plan-link delete.

Each PR fixes one and then trips over the other. Verified locally by merging the two branches together: `plan_listener::tests::applied_plan_transitions_to_resolved` passes and `contract_schema_parity` reports 22/22. Landing both clears the check; the merge order does not matter.

PR #1601, opened 2026-07-27
chore: migrate 6 CSS sheets to vanilla-extract (wave 1)

srobroek:
## Summary

Dissolves 6 BEM CSS sheets (1,187 LOC) into co-located vanilla-extract `.css.ts` files. No user-visible changes — build-time migration that adds type-safe style authoring.

## What changed

- wizard-base.css, modals.css, skeleton.css, dev.css, app-shell.css, sessions.css replaced with per-component `.css.ts` files
- Each `.css.ts` has exactly one consumer (three-layer rule: token → primitive → component)
- Tests updated to use `data-testid` selectors instead of generated class names
- `pv-session-frames__*` deleted (dead code, no TSX consumer)
- Cross-sheet: `pv-detail--fill`, `pv-sessions-table__*`, `pv-wizard__step*` retained as BEM until their owner sheets migrate

## Spec Context

`specs/tiny/css-vanilla-extract-migration.md` added in this PR.
PR targets `css/ve-foundation`; retarget to `main` after foundation merges.

Merge-Bead: astro-plan-is9s
Tracks-Bead: astro-plan-kyo7.109.1

---

Recreated from #1552 (closed) on the same head `css/ve-wave1` @ 452dccc5. No code changed in the re-creation.

NOTE: this branch still conflicts with `main` and needs a rebase before it can merge — re-creating the PR does not resolve that.

PR #1618, opened 2026-07-29
fix: repair the unstyled components left by the wave-1 CSS migration

srobroek:
## Summary

- Repairs the seven UI regressions the wave-1 vanilla-extract migration introduced, so `css/ve-wave1` no longer ships unstyled components.
- The migration stopped `components.css` importing `app-shell.css` and `wizard-base.css` while nothing else imported them, so every `pv-*` class defined only there went unstyled. Fixed by reattaching styling at each component rather than restoring the imports, which would undo the migration.

## What's fixed

- Empty states and detail-header subtitles apply their own styles instead of emitting bare legacy class names; long archive paths wrap again.
- The loading spinner animates again, in both migrated and not-yet-migrated sheets.
- The status-bar disk meter fills, using the ok/warn colours rather than the accent colour.
- List-sidebar search and filter controls are styled again.
- The project-detail action bar no longer overlaps, now via a component variant rather than a page-scoped override.
- The calibration match table fits the side dock instead of forcing its minimum width.
- A source-folder test id no longer collides when two folders share a kind.

## Verification

Independent review confirmed all seven against a real `vite build`, grepping the emitted CSS rather than inferring from source — including that the spinner keyframes resolve for the unmigrated consumer and that the meter uses the `ok` token.

`pnpm typecheck` clean. 256 test files / 2284 tests pass. A new guard test covers the three class-application fixes; reverting two of them makes it fail, so it is load-bearing rather than tautological. The four CSS-only fixes cannot be asserted in jsdom and rest on the build-output evidence above.

The review found one blocker — three files failing `biome check`, which the lint gate runs — fixed in 364b016f. `biome check src` now reports zero errors.

## Base

Targets `css/ve-wave1`, not `main`: three of the stylesheets it edits exist only on that branch. Zero file overlap with what `main` has moved since the base diverged, so this does not worsen the existing `#1601` conflict.

Merge-Bead: astro-plan-3d8r

### DIFF
--- a/apps/desktop/src/app/statusbar.css.ts
+++ b/apps/desktop/src/app/statusbar.css.ts
@@ -3,8 +3,9 @@
 
 // StatusBar.tsx styles.
 
-import { style } from '@vanilla-extract/css';
+import { globalStyle, style } from '@vanilla-extract/css';
 import { vars } from '@/styles/themes.css';
+import { spin } from '@/styles/motion.css';
 
 export const statusBar = style({
   height: 'var(--pv-statusbar-height)',
@@ -39,7 +40,7 @@ export const spinner = style({
   borderRadius: '50%',
   border: `2px solid ${vars.accent}`,
   borderTopColor: 'transparent',
-  animation: 'pv-spin 1s linear infinite',
+  animation: `${spin} 1s linear infinite`,
 });
 
 export const vol = style({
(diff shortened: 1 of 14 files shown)


srobroek:
Fixed the alm eslint gate bounce in 6fc958b9: both exported roots now carry a real `data-testid` (`detail-header`, `top-action-bar`) rather than using the rule disable hatch, and nothing was baselined. `check-eslint-baseline.mjs` goes from 2 new violations to `eslint: OK`.

srobroek:
@coderabbitai full review

PR #1622, opened 2026-07-29
fix: CodeRabbit review scanner emits valid JSON for every status

srobroek:
## Summary

- `scripts/check-coderabbit-review.sh --json` emitted invalid JSON: `$detail` was interpolated into a JSON string with `printf`, and the `SKIPPED_DRAFT` detail contains literal double quotes around `@coderabbitai full review`, so `jq` failed with `Invalid numeric literal at column 168`.
- The object is now built with `jq -cn --arg`, which removes the escaping class rather than escaping one string. `--json` exits 2 when `jq` is absent.
- `scripts/tests/check-coderabbit-review.bats` gains one test for the quoted `SKIPPED_DRAFT` detail and one covering every status the scanner emits (`RATE_LIMITED`, `SKIPPED_DRAFT`, `PENDING`, `NOT_REVIEWED`, `REVIEWED`, `STALE`, `ERROR`). Both assert through `jq -e`, so invalid JSON fails the test.

This branch merges `tooling/coderabbit-review-scanner` and supersedes PR #1606, so the scanner is added here.

## Verification

- `bats scripts/tests/check-coderabbit-review.bats`: 20/20 pass.
- `shellcheck scripts/check-coderabbit-review.sh`: clean.
- Live run: `bash scripts/check-coderabbit-review.sh --json 1606 | jq -e '.status,.detail'` returns `"SKIPPED_DRAFT"` with the escaped detail, exit 0.

## Beads

Tracks-Bead: astro-plan-btnb
Merge-Bead: astro-plan-6n93

### DIFF
--- a/scripts/check-coderabbit-review.sh
+++ b/scripts/check-coderabbit-review.sh
@@ -61,6 +61,8 @@ while (($# > 0)); do
 done
 
 command -v gh >/dev/null || { echo 'check-coderabbit-review: gh is required' >&2; exit 2; }
+((json_mode == 0)) || command -v jq >/dev/null ||
+  { echo 'check-coderabbit-review: --json requires jq' >&2; exit 2; }
 
 if ((open_mode)); then
   ((${#prs[@]} == 0)) || { echo 'check-coderabbit-review: --open takes no PR numbers' >&2; usage; }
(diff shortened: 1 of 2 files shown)


srobroek:
@coderabbitai full review

srobroek:
@coderabbitai full review

srobroek:
### Rate-limit wait parsing (astro-plan-vpie)

`wait_hint()` and the notice selector both required the literal phrase `next review available in`. CodeRabbit also posts `Your next included review will be available in 48 minutes`, which neither pattern matched. The RATE_LIMITED detail therefore fell through to `review window has reopened ... safe to re-trigger` while quota was exhausted. The lead re-triggered four PRs on that reading and burned quota for zero reviews.

Detection no longer depends on a sentence shape:

| | before | after |
| --- | --- | --- |
| notice selector (jq) | `test("[Nn]ext review available in")` | limit indicator `rate limited by coderabbit\|review limit\|fair usage` AND a duration `[0-9]+[^a-zA-Z]*(minute\|hour)` |
| figure parser (grep) | `next review available in:?\**...([0-9]+)...(minute\|hour)` | the same duration regex, over a body already known to be a limit notice |
| no figure parsed | `re-trigger after the window`, or a false `has reopened ... safe to re-trigger` | `wait unknown ... re-check before re-triggering` |

`reopened` is emitted only when a deadline was parsed and has passed.

`bot-review-probe.py` in agentic-packages (`packages/pr-shepherd/`) classifies from reviews-at-head and check state with no notice matching, so it never had this bug. It also cannot distinguish "bot working" from "bot refused". This script keeps notice matching for that distinction and adopts the probe's ordering, in which a review at head outranks any notice text. Bats pins that in `a real review at head outranks an old rate-limit notice`, the STALE mirror case, and `a draft-skip notice cannot mask a genuine review at head`. That rule is where the two implementations can be reconciled.

Test plan: 26 bats cases pass, `shellcheck` clean. `--json 1622` reports `next review in ~25 min (at 06:19 UTC)`; PRs 1621/1624/1625 report the same window where all four previously reported it reopened.

PR #1623, opened 2026-07-29
fix: log panel no longer misses events written before the live stream starts

srobroek:
## Summary

- The live log stream seeded its cursor from the highest existing event id, so every event written between app start and stream start was never sent to the log panel. Those events now stream.
- The catch-up is bounded to the newest 500 events, matching the log panel buffer. When older events exist, the panel shows a warning entry naming how many were not streamed; a refresh reloads them.

## Verification

- `cargo test -p persistence_lifecycle --lib repositories::events` — 12 passed, including `rewound_cursor_replays_window_and_counts_skipped`, which fails against the previous cursor seed.
- `cargo clippy -p desktop_shell -p persistence_lifecycle --all-targets` — no warnings.
- `cargo fmt --check` on both crates — clean.

## Spec Context

Tracks-Bead: astro-plan-pi3p
Closes-Bead: astro-plan-pi3p
Merge-Bead: astro-plan-jk7r


srobroek:
@coderabbitai full review

Issue #506, opened 2026-07-09
Evaluate source-protection complexity; per-source control only in Settings, not wizard


srobroek:
## Summary (deferred evaluation — not a bug)

Question raised during validation: is **source protection** deprecated / worth
its complexity? Short answer: **not deprecated** — it is actively wired and
consumed. This issue is to decide whether to keep, simplify, or resurface it.

## What exists today

- **Contract:** `ProtectionLevel` (protected/normal/unprotected) —
  `crates/contracts/core/src/protection.rs`.
- **Persistence:** `source_protection_state` table —
  `crates/persistence/db/src/repositories/source_protection.rs`.
- **Consumed by cleanup / permanent-delete safety:**
  `crates/app/core/src/plans.rs` — `PlanItemProtection`, and the permanent-delete
  guard returns `ErrorCode::PlanBlockedByProtection` ("spec-016
  blockPermanentDelete"). Protected items are skipped/blocked.
- **Audit:** `ProtectionDefaultChanged`, `ProtectionSourceSet`,
  `ProtectionPlanAcknowledged`.
- **Per-source override UI exists:**
  `apps/desktop/src/features/settings/SourceProtectionOverride.tsx` (spec 016 US2)
  — a protection badge + override control for a source UUID, in **Settings**.
- **Wizard sets only the global default:**
  `apps/desktop/src/features/setup/steps/StepCatalogs.tsx` (spec 018).

## Human observations

1. "Source protection adds a bunch of complexity — is it still needed?" → It is
   still consumed by the cleanup/delete safety model, so it is not free to
   remove; a decision is needed on whether the 3-level model earns its keep.
2. "Shouldn't this be set on each source? I didn't see an option." → Per-source
   protection **is** settable, but only in **Settings**
   (`SourceProtectionOverride`), **not** on the wizard Step-1 folder cards (which
   expose only org + scan depth). Discoverability/placement gap.

## Decision requested

- (a) Keep as-is; (b) simplify (e.g. drop to a single protected/unprotected
  toggle, or global-only, or per-source-only); (c) surface per-source protection
  on the wizard folder cards for discoverability.

## Source

Windows real-app validation run 2026-07-09 (backlog B14). Human-directed deferred
evaluation.

srobroek:
[...]

PR #1625, opened 2026-07-29
fix: corrupted stored settings and wizard state fall back to defaults

srobroek:
## Summary

- Corrupted or stale stored values now fall back to defaults instead of loading as wrong values. Preferences, the project wizard draft, the setup wizard state, and the registered source folders each validate what they read from browser storage.
- The Cleanup, Framing, Source Views and Naming settings panes validate the values they load from the backend. A value left behind by an older build is dropped in favour of the default rather than applied.
- One malformed field no longer discards its valid neighbours: a stale entry in the setup wizard state keeps the step position and registered folders intact, and a bad source folder entry no longer drops the rest of the list.
- Restored the title-bar close button on the "archive master in use" and "unassign calibration" confirmation dialogs, which lost it in the shared-dialog change this branch builds on.

## Details

Supersedes #1594. Branch is #1594's head 5dec4412 merged into main, plus the fixes below.

Tracks-Bead: astro-plan-iqd5
Merge-Bead: astro-plan-qdez

### Blocker 1 — local-storage.ts had no production consumers

`apps/desktop/src/lib/local-storage.ts` exported a validating read/write pair that only its own test file imported. All four sites its docstring named still hand-rolled `JSON.parse` with no schema check, so the claimed fallback behaviour did not exist. Adopted at every one:

| Site | Was |
| --- | --- |
| `data/preferences.ts:70` | `{...defaults, ...JSON.parse(raw)}`, unvalidated |
| `features/projects/wizard/WizardPage.tsx:154` | `{...INITIAL_DATA, ...JSON.parse(raw)}`, unvalidated |
| `features/setup/sources-store.ts:112` | hand-rolled per-entry type guard |
| `features/setup/SetupWizard.tsx:196` | `JSON.parse(raw) as Partial<WizardState>` |

Fields are individually droppable, not whole-object. A whole-object rejection on the setup wizard would reset the step position and registered sources along with the bad field.

`SetupWizard.tsx:196` previously accepted `parsed.sources` as `SourcesState` whenever it was an array, bypassing `loadSources()`'s per-entry validation and `organizationState` backfill. It now goes through `loadSources()`.

### Blocker 2 — four of five settings schemas were unconsumed

Cleanup, Framing, SourceViews and `useNamingPattern` cast `data.values` instead of using their schema. All four now call `getSettingsTyped`. `parseSettingsValues` is extracted from `getSettingsTyped` so `RestoreDefaultsBtn`'s `onRestored` bag gets the same validation as the mount-time read.

Three schemas disagreed with the code they were written for and would have admitted invalid values on adoption:

- `DEFAULT_PROTECTION_LEVELS` carried the `standard` level issue #506 retired; `crates/domain/core/src/settings.rs:138` and `Cleanup.tsx` are both 2-level.
- `LINK_KINDS` carried a nonexistent `none`, omitted `junction`, and applied one union to both keys — cross-drive excludes `hardlink` (FR-004a, `crates/domain/core/src/settings.rs:226,231`).
- `NamingSettingsSchema` typed `pattern` as `unknown[]`, which cannot yield `PatternPart[]`. Now models `PatternPartDto` (`bindings/index.ts:7282`), with `kind` left a plain string because the backend types it as `String`.

### Non-blocking items

- `ConfirmModal` hardcoded `hideClose`. Verified against base fc1d9c31: Equipment, DataSources and ObservingSites passed `hideClose`; MasterArchiveFlow and SessionDetail did not, so both silently lost their close button. Now a prop defaulting to `true`, with those two opted out.
- Both new components' docstrings named adopters that still hand-roll. Corrected to the six real `ConfirmModal` call sites and `TableStateGate`'s single one.
- Neither component had a direct test. Added both, including the `loading && isEmpty` skeleton gate — a refetch with data present no longer flashes a skeleton, which `ArchivePage.test.tsx` cannot catch because it sets `data = []`.

## Verification

| Command | Result |
| --- | --- |
| `pnpm --filter @astro-plan/desktop typecheck` | pass |
| `pnpm vitest run` (apps/desktop) | 262 files, 2355 tests pass |
| `pnpm lint` (apps/desktop) | pass |
| `pnpm knip` (apps/desktop) | no unused files or exports |

`scripts/eslint-alm-baseline.txt` carries four line-number updates for pre-existing `alm/require-root-testid` entries whose lines shifted. No entry was added or removed.


srobroek:
@coderabbitai full review

srobroek:
@coderabbitai full review

srobroek:
@coderabbitai full review

PR #1627, opened 2026-07-30
fix: dev settings loading padding, spotlight fallback delay, and dev call-list column alignment

srobroek:
## Summary

- The Dev Settings loading text renders padded again, not flush against the panel edge.
- The Find-It spotlight explains an absent control about 1.5s after its page appears, instead of 3s after activation.
- The dev call-list actions column keeps its width and baseline alignment.
- The sidebar item padding comes from spacing tokens; rendering is unchanged.
- The onboarding contrast test asserts a non-empty class instead of a generated vanilla-extract class name.

Fixes five of the nine open CodeRabbit findings on #1601. The other four are already fixed on #1618, which targets the same base branch; re-implementing them here would only create conflicts.

## Fixed

| Finding | Fix |
| --- | --- |
| `dev-settings-page.css.ts:27` dropped `padding: var(--pv-sp-7)` | Restored, matching the pre-split shared dev style (`9e512de00^:apps/desktop/src/dev/dev.css.ts:243`). |
| `FindSpotlight.tsx:158` missing-target fallback waits 3s | Split into a 3s pre-arrival navigation grace and a 1.5s post-arrival anchor wait. |
| `call-list.css.ts:43` `display: flex` on a `<td>` | Flex row moved to a wrapper inside the cell; `CallList.tsx` updated. |
| `sidebar.css.ts:84` raw `6px` padding | Composed as `calc(var(--pv-sp-1) + var(--pv-sp-0))`; rendered value unchanged. |
| `onboarding_control_contrast.spec.ts:27` asserts the `selectBase` token | Probe still reads the live class; the assertion is now non-empty-class. |

A `<td>` with `display: flex` leaves table layout, so its column stops contributing to column width and baseline alignment. Only `CallList.tsx` renders that cell, behind the compile-time `dev-tools` feature.

The `6px` sidebar padding has no single-token equivalent: the scale is base-4 (`--pv-sp-1: 4px`, `--pv-sp-2: 8px`). `--pv-sp-0` is 2px, so the sum is exact. `apps/desktop/src/styles/components/settings.css:23` uses the same composition pattern.

## Already fixed on #1618

| Finding | Where |
| --- | --- |
| `statusbar.css.ts:58` missing disk-meter fill | `55198e99f` adds the meter child-`<i>` global rule. |
| `DetailHeader.tsx:34` renders legacy `pv-detail__subtitle` | `55198e99f` applies the generated `subtitle` class. |
| `StepSourceFolders.tsx:338` duplicate `data-testid` | `55198e99f` appends the row index and updates the test. |
| `setup_wizard_source_folders.spec.ts:152` asserts `selectBase` | `55198e99f` asserts the per-row test id and value. |

## Verification

- `pnpm typecheck` from `apps/desktop`: pass.
- `pnpm test` from `apps/desktop`: 257 files, 2300 tests, pass.
- `pnpm lint` from `apps/desktop`: pass. `scripts/eslint-alm-baseline.txt` regenerated with `node scripts/check-eslint-baseline.mjs --generate`; the single change is a line number that shifted behind an added comment.

## Spec Context

Tracks-Bead: astro-plan-85d7
Merge-Bead: astro-plan-0s7f

### DIFF
--- a/apps/desktop/src/app/sidebar.css.ts
+++ b/apps/desktop/src/app/sidebar.css.ts
@@ -81,7 +81,9 @@ export const item = style({
   display: 'flex',
   alignItems: 'center',
   gap: 'var(--pv-sp-2)',
-  padding: '6px var(--pv-sp-3)',
+  // 6px is off the base-4 spacing scale, so it is composed from tokens rather
+  // than hardcoded; the rendered value is unchanged.
+  padding: 'calc(var(--pv-sp-1) + var(--pv-sp-0)) var(--pv-sp-3)',
   margin: '1px var(--pv-sp-1)',
   cursor: 'pointer',
   color: vars.textSecondary,
--- a/apps/desktop/src/dev/CallList.tsx
+++ b/apps/desktop/src/dev/CallList.tsx
@@ -20,6 +20,7 @@ import {
   tdContract as callsTdContract,
   tdStarted as callsTdStarted,
   tdActions as callsTdActions,
+  actions as callsActions,
   truncated as callsTruncated,
   outcomeVariants as callsOutcomeVariants,
   replayBtnVariants as callsReplayBtnVariants,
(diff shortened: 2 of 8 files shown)


srobroek:
@coderabbitai full review

srobroek:
@coderabbitai full review

srobroek:
@coderabbitai full review

PR #1629, opened 2026-07-30
fix: action bar with a wrapped action row no longer shows a double gap

srobroek:
## Summary

- `TopActionBar` in wrap mode applied its `rowGap` twice. The spacer took `flex-basis: 100%` and became a zero-height flex line between the title row and the action row, so the bar gained one gap above the spacer line and another above the actions line.
- The wrap variant now renders no spacer. `actionsWrap` carries `flex: 1 1 100%`, which already forces the action cluster onto its own line, so wrap mode has two flex lines and one gap. The `spacerWrap` style is deleted.
- Non-wrapped rendering is unchanged: the spacer still renders with `flex: 1`.

This clears the last CodeRabbit finding on #1618 (`TopActionBar.css.ts:48`, empty flex row in wrapped action bars).

## Test plan

- `pnpm typecheck` passes.
- `pnpm test` passes, 256 files / 2284 tests.
- `pnpm lint` passes.
- `apps/desktop/src/components/ve-class-application.test.tsx` asserts the spacer element is present unwrapped and absent wrapped.

## Base

Targets `fix/ve-wave1-regressions`, not `main`: `TopActionBar.css.ts` exists only on that chain.

Tracks-Bead: astro-plan-5md3

Merge-Bead: astro-plan-qx1k

Coverage gap: no test measures rendered wrap-mode spacing. jsdom performs no flexbox layout, so the one-gap result is asserted structurally rather than visually.

### DIFF
--- a/apps/desktop/src/components/TopActionBar.css.ts
+++ b/apps/desktop/src/components/TopActionBar.css.ts
@@ -44,9 +44,9 @@ export const actionBarWrap = style({
   paddingBottom: 'var(--pv-sp-2)',
 });
 
-/** Full-width spacer forces the action cluster onto its own row. */
-export const spacerWrap = style({ flexBasis: '100%' });
-
+// `flex: 1 1 100%` already forces this onto its own flex line, so the wrap
+// variant renders no spacer: an empty full-width spacer would form a third,
+// zero-height line and apply `rowGap` twice.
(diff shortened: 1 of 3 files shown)


srobroek:
@coderabbitai full review

PR #1630, opened 2026-07-30
fix: an unreachable library root no longer erases which session a frame belonged to

srobroek:
Merge-Bead: astro-plan-3rfg
Tracks-Bead: astro-plan-2uwa

## Summary

- A library root that is unmounted, unreadable, or a symlink produced an empty
  directory listing, and reconcile read that as "every frame in this root was
  deleted". Every frame was marked missing, and in auto-reconcile mode each one
  was removed from its session. Nothing restored that session grouping when the
  drive came back.
- Reconcile now checks the root is reachable first and stops the pass when it is
  not, leaving frame records untouched. The operation reports `root.unavailable`
  and is safe to retry.
- A frame that is missing keeps its place in its session in both reconcile
  modes. Session frame counts already excluded missing frames everywhere except
  the sources list, which is fixed here.

## Defect chain

| Step | Location |
| --- | --- |
| Empty listing for an unwalkable root | `crates/fs/pathsafe/src/lib.rs` `real_files_under` |
| Empty listing read as every frame deleted | `crates/fs/inventory/src/reconcile.rs` `reconcile_root` |
| Missing frame removed from session membership | `crates/app/core/src/frame_inventory/reconcile.rs` |
| No inverse anywhere: recovery only flips `file_record.state` | `apply_present_outcome` |

Frame-to-session attribution is Tier-1 user knowledge the filesystem cannot
supply back (constitution V), so the removal was unrecoverable.

## Changes

- `fs_pathsafe::probe_root` classifies why a root is not walkable (`Missing`,
  `NotADirectory`, `GatedLink`, `Unreadable`) under the same `follow_symlinks`
  rule the walkers apply. `Path::is_dir()` is not a substitute: it follows
  links, so it accepts a symlinked root the gated walker then refuses to
  descend.
- `reconcile_root` returns `Result` and aborts before any frame state is
  written. `run_reconcile` maps the reason to a retryable `root.unavailable`.
- `drop_frame_from_session_membership` is deleted.
- `list_sessions_for_root` and `get_session_context_by_ids` count only
  `frame_ids` members whose `file_record.state != 'missing'`, replacing a raw
  `json_array_length`.
- Regression tests: unmounted root and symlinked root, at the walker layer and
  the app layer.

## Why retaining membership is safe

Every active count already filtered on frame state:
`q_core::active_frame_summaries_batch` (`q_core.rs:859`),
`active_frame_exposure_seconds_batch` (`q_core.rs:919`), `list_frames`
(`list.rs:66`), cleanup candidacy (`raw_frames.rs:98,175`). The inventory
session projection was the one consumer that did not, and now does. The
alternative — an inverse operation — needs a new table to record what was
dropped, storing data the retained array already holds.

## Out of scope

Guarding the unattended reconcile triggers belongs to the frame watcher branch.
It should call `fs_pathsafe::probe_root(root, follow_symlinks)` in place of its
`Path::is_dir()` probe, on both the scheduled tick and the NeedsRescan/overflow
recovery path, and skip the pass when it errors.

## Verification

- `cargo test -p fs_pathsafe -p fs_inventory -p app_core -p persistence_targets` — pass
- `cargo clippy --workspace --all-targets` — clean

## Spec Context

Spec 048 FR-010 describes auto-reconcile as dropping a missing frame from
active membership. This PR keeps the record and the membership and relies on the
state filter for "active", because the drop was unrecoverable. FR-010 needs
amending to match.


srobroek:
@coderabbitai full review

PR #1631, opened 2026-07-30
ci: Windows E2E flakes are visible again instead of absorbed by retries

srobroek:
## Summary

- Lowers the Windows E2E retry budget from 2 to 1 so a second-attempt pass shows up as `(N flaky)` instead of being absorbed silently.

## Why now

At `retries = 2`, a green Windows shard was not evidence of a clean first attempt. The contamination symptom the override exists to mask could therefore neither be observed nor declared gone: the retry budget was hiding the signal needed to decide either way.

The isolation fix does appear to hold. `astro-plan-1oe` records per-shard TRY-1 output from PR #1465 (jobs `89248442052` / `89248442207` plus reruns): three named Windows journeys pass, 17 passed 13 skipped, no `(N flaky)` suffix on either shard.

## Why 1 rather than 0

One attempt of headroom stays for genuine runner faults. `STATUS_DLL_INIT_FAILED` and the `0xC0000142` startup deaths tracked by `astro-plan-gl5q` are runner-image problems, not test flakes, and they kill the process before any test runs.

## What this is

An experiment with a defined outcome. If flakes return, that is the answer: raise it back to 2 and reopen the contamination investigation with the shard logs as evidence. If they do not, the override can go to 0 later on the same basis.

`threads-required = 2` is unchanged. The existing comment gates lowering that on a real Windows run under `test-threads > 1`, which has not happened.

## Verification

- TOML parses (`tomllib`)
- The Linux/macOS `retries` values elsewhere in the file are untouched

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -65,9 +65,11 @@ failure-output = "immediate-final"
 
 # Windows-latest CI runners are chronically the closest to the slow-timeout
 # edge (debug-build WebView2 boot + intermittent runner-level stalls; even
-# green CI runs there show 1-3 retried flakes recovering by luck). One extra
-# retry buys more margin without masking a real regression, since a genuinely
-# broken journey still fails all 3 attempts.
+# green CI runs there show 1-3 retried flakes recovering by luck). The single
+# retry below buys margin against a runner stall without masking a real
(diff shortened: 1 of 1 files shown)


srobroek:
@coderabbitai full review

srobroek:
@coderabbitai full review

PR #1635, opened 2026-07-30
feat: validate every CSS design-token reference

srobroek:
## Summary

- Every `var(--pv-*)` reference in `apps/desktop/src/styles/**/*.css` is now validated against the generated design-token artifacts. No tool in the repo read CSS before this: Biome's includes are scoped to `src/**/*.{ts,tsx,js,jsx}`, so 3,000+ token references had nothing checking them, and an undefined token makes the browser drop the declaration silently.
- Re-lands the CSS half of #1304. Its four deliverables merged, but only two reached `main`; #1406 re-landed the JS half and left these out as inert files.

## What lands

| Deliverable | State |
| --- | --- |
| `apps/desktop/stylelint.config.mjs` + `stylelint` devDependency | landed, wired into `pnpm lint` |
| `apps/desktop/tokens/properties.json` + `@property` emission in `build-tokens.mjs` | landed |
| Delete `NamingStructure.r4.test.ts`, drop the `.test.` exclusion in `check-token-refs.mjs` | not landed, see below |
| Colour-literal rules (`color-no-hex`, `function-disallowed-list`) | not landed, see below |

The rule is stylelint's built-in `no-unknown-custom-properties`, pointed at `src/styles/tokens.css` and `packages/tokens/tokens-docs.css`, so the valid set is whatever the token pipeline emitted rather than a second hand-maintained list.

## The two halves are one change

`no-unknown-custom-properties` accepts any reference carrying a resolvable fallback. All eight component parameters were written as `var(--pv-modal-w, 640px)` and declared nowhere, so the config alone would not see them. `tokens/properties.json` emits each as an `@property` registration whose `initial-value` replaces the fallback.

`skeleton.css` required a behaviour-preserving change rather than a fallback deletion: its `--block` and `--circle` variants set geometry through the fallback slot, which `@property` makes unreachable. They now override the custom property. The other four fallback removals were byte-equal to their `initial-value`.

## Gate executes and bites

stylelint runs in the `lint` chain between the ESLint baseline and `build-tokens --check`. CI runs `pnpm lint`, so no workflow change was needed beyond the step name.

- Clean tree: `pnpm lint` exit 0.
- `--pv-modal-w-typo` injected into `modals.css`: exit 2, `117:24 Unknown custom property "--pv-modal-w-typo"`. Reverted.
- The fallback loophole verified directly: `var(--pv-bogus2, red)` passes while `var(--pv-totally-bogus-token)` fails. That is why the `@property` half is required.

## Deliberately not landed

**Colour-literal rules.** `color-no-hex` and `function-disallowed-list` are the acceptance criteria of bead `astro-plan-n2p`, and enabling them forces that bead's open decision on `apps/desktop/src/styles/components/primitives.css:102` (`box-shadow: inset 0 2px 5px rgba(0, 0, 0, 0.22)`): tokenise it, or exempt it with a documented reason. `origin/css/ve-wave1` carries the same line, so enabling the rules here would fail that branch. Left to `n2p`, which this PR unblocks.

**The `.test.` exclusion.** `apps/desktop/src/features/settings/NamingStructure.r4.test.ts` still exists and still asserts on the literal `var(--pv-radius)` it is pinned to forbid (8 occurrences), so the exclusion in `check-token-refs.mjs` cannot be dropped without deleting a guard test. #1304 deleted it as part of a wider sweep this branch does not carry.

## Open vanilla-extract branches

`css/ve-wave1` (#1601) and `css/ve-wave2` (#1572) both pass the landed rule: neither adds a `var(--pv-*)` reference absent from `main`, and their `.css.ts` files fall outside the `src/styles/**/*.css` glob. `ve-wave2` touches `primitives.css`, so it conflicts textually with the one-line `row-indent` fallback removal here.

## Verification

`pnpm typecheck` clean. `pnpm test` 257 files, 2300 tests passed. `pnpm lint` exit 0.

Tracks-Bead: astro-plan-ojfl
Merge-Bead: astro-plan-gy8h


srobroek:
@coderabbitai full review

PR #1634, opened 2026-07-30
ci: fail builds on functions above a cognitive-complexity ceiling of 20

srobroek:
## Summary

- Adds a cognitive-complexity ceiling of 20 to Rust and TypeScript using the native lints in clippy and biome, so overly tangled new functions fail CI.
- The 70 functions already above the ceiling carry in-code exception annotations naming each function's shape; the Rust ones self-expire once the function is simplified.

Tracks-Bead: astro-plan-kyo7.114
Merge-Bead: astro-plan-flmd

## What lands

| Side | Config | Enforced by |
| --- | --- | --- |
| Rust | `clippy::cognitive_complexity = "warn"` (Cargo.toml:173), `cognitive-complexity-threshold = 20` (clippy.toml) | `cargo clippy --workspace --all-targets -- -D warnings` (ci.yml:409) |
| TypeScript | `complexity/noExcessiveCognitiveComplexity` error, `maxAllowedComplexity: 20` (apps/desktop/biome.json:25) | `pnpm --filter @astro-plan/desktop run lint` (ci.yml:351) |

Both gates ride CI steps that already exist, so neither needs a justfile recipe. This avoids the failure mode recorded in `astro-plan-kyo7.115`, where a recipe was added and never wired into the workflow.

## Threshold chosen from measurement

Measured violation counts, whole workspace / `apps/desktop/src`:

| Threshold | Rust | TypeScript |
| --- | --- | --- |
| 8 | 310 | — |
| 10 | 147 | 135 |
| 12 | 97 | 120 |
| 15 | 67 | 96 |
| 20 | 36 | 34 |
| 30 | 16 | 5 |

Neither side passes at 10. Threshold 10 would require 282 in-code exception annotations, at which point the annotation stops being a review signal. 20 is the ratified hard ceiling (CCN<10 norm, 10-20 by justified exception, >=20 forbidden).

## Exceptions instead of a baseline file

The 36 Rust sites above 20 carry `#[expect(clippy::cognitive_complexity, reason = "<shape>")]`. The 34 TypeScript sites carry `// biome-ignore lint/complexity/noExcessiveCognitiveComplexity: <shape>`. Each reason names the function's shape, read from its code.

`#[expect]` rather than `#[allow]`: clippy errors on an unfulfilled expectation, so an exception vanishes the moment the function drops below the threshold. A line-keyed text baseline does not self-drain.

The exception registry is one grep:

```
rg 'expect\(clippy::cognitive_complexity'
[...]

PR #1545, opened 2026-07-24
feat(css): migrate LogPanel to vanilla-extract with generated theme contract

srobroek:
## Summary

Migrates the LogPanel component from hand-authored BEM CSS to [vanilla-extract](https://vanilla-extract.style/). Theme colors for all 6 built-in themes are now generated from the existing DTCG token sources and enforced by a CI drift gate, so a token change propagates to both the legacy CSS variables and the VE contract in one commit.

## What's new

- LogPanel and LogEntryRow use type-safe, zero-runtime VE class names; the `logpanel.css` BEM file is emptied
- `scripts/gen-ve-themes.mjs` reads DTCG sources and emits `src/styles/themes.css.ts` (6 themes, 49 semantic keys, `createGlobalTheme` per theme)
- Theme switching remains a `data-theme` attribute flip on the document root — no JS re-bundle required
- `gen-ve-themes.mjs --check` drift gate runs in CI alongside the existing `build-tokens.mjs --check`; `just ve-themes-check` recipe added
- `docs/development/css-vanilla-extract.md`: co-location rule, `vars.*` import requirement, `styleVariants` pattern, `globalStyle` policy, Phase-2 sheet migration order

## Verification

| Check | Result |
|---|---|
| vitest (LogPanel suite) | 16/16 pass |
| vitest full suite | 2128/2131 pass (3 pre-existing in `devSurface.release.test.ts` on `origin/main`) |
| tsc --noEmit | PASS |
| biome check | PASS |
| gen-ve-themes.mjs --check | `OK: themes.css.ts matches DTCG sources.` |
| size-limit JS | 673 kB / 720 kB PASS |
| 6-theme CSS in build | all 6 `[data-theme=X]` selectors present in `dist/assets/main-*.css` |

## Spec Context

Tracks-Bead: astro-plan-kyo7.109
Merge-Bead: astro-plan-bkt2

PR #1636, opened 2026-07-30
feat: inventory tracks files appearing and disappearing under a library root

srobroek:
Adds a per-root live frame watcher, so inventory reflects files appearing and disappearing under a library root without waiting for a manual rescan.

## Summary

- Inventory now updates itself when files appear or disappear under a watched root, instead of requiring a manual rescan.
- A failed reconcile is retried rather than silently dropped, so inventory cannot go quietly stale.

## Status — the earlier "tracking record" framing no longer applies

This PR previously said it was a preservation record and not a merge candidate. That is now out of date: it has been merged with current main (no conflicts), reviewed, and had two real defects fixed.

## What review found and fixed

**HIGH — a failed reconcile silently discarded the events that triggered it.** `frame_watcher.rs` cleared `dirty` *before* awaiting `schedule_reconcile` and ignored the result, so a transient failure (locked DB, IO blip) threw away the only record that those filesystem events were never processed — and nothing retried until an unrelated event arrived. Now `dirty` persists for retryable failures.

The opposite failure mode is guarded too: a non-retryable `RootUnavailable` *does* discharge, so a permanently unregistered root cannot spin the debounce loop. A naive fix here would have traded a silent drop for a hot loop.

This is the fourth instance this cycle of one family — **work discarded because a marker was cleared before the awaited thing succeeded** — the other three being the cursor defects in #1644.

**Build failure — clippy `cognitive_complexity` at 42/20** on `attach_root_watcher`, under a ceiling #1634 has since activated. Split out `publish_entry` at a genuine seam (resolve-and-validate vs lock/tombstone registration). The residual 24/20 and 21/20 is guard-arm `tracing!` macro expansion rather than real branching, so each carries `#[expect]` with a stated reason.

**The three failing test legs were one defect, not a platform issue** — all failed on the identical `TS2554` in this branch’s own new test file, where hoisted `vi.fn()` mocks inferred a zero-argument signature.

Checked and found sound: local-first custody (the watcher only schedules reads; no `fs::write`/`remove` outside test code), the symlink gate (`follow_symlinks` threads through, and the test proves both halves non-vacuously — asserting *no record written* when the gate is closed), bounded memory (channel capped at 256, with overflow promoting to a full reconcile so it degrades to a rescan rather than losing data), and attach/detach races including the stale-tombstone clear.

## Verification

`cargo clippy --workspace --all-targets -- -D warnings` clean · `cargo fmt --all --check` clean · `desktop_shell` **112 tests pass** · `contracts_core` + `fs_inventory` **109 pass** · `pnpm typecheck` exit 0 · vitest inventory/sessions **152 pass** · biome clean.

## Known gaps

- `frame_watcher_symlink_gate.rs` is `#![cfg(unix)]`, so Windows junction handling rests on CI rather than a local test.
- Nothing demotes a root to polling when live events prove unreliable. That needs a product call on what evidence justifies overriding an explicit `live: true`, so it is filed as `astro-plan-nrlx` rather than guessed at here.

## Spec Context

Spec 048 T023/T024/T026.


srobroek:
## Review pass and staleness fixes

First review of this branch. Merged `origin/main` (35 commits, no conflicts), fixed the CI failures, and reviewed the watcher as its first reviewer.

### Failing checks, classified

| Check | Category | Evidence |
| --- | --- | --- |
| Unit + integration — macos / ubuntu / windows | **Real defect** in this commit | All three legs failed on the identical error: `useRootFrameWatcher.test.tsx(22,67): error TS2554: Expected 0 arguments, but got 1`. The hoisted `vi.fn(() => …)` mocks inferred a zero-arg signature while the `vi.mock` factory calls them with a request object. Fixed by typing the mock parameters. |
| `fix end of files` (`apps/desktop/src/styles/themes.css.ts`) | **Inherited** | File last touched by #1545, not by this branch. |
| `typos` (`validateFieldSeverity.test.ts`) | **Inherited** | File last touched by #1555, not by this branch. Both pre-commit failures reproduce on `main`. |

The `Rust coverage (llvm-cov)` inherited-Node-deps pattern did not apply — this PR's failures were all the typecheck step.

### Review findings

**1. HIGH — a failed reconcile silently dropped the events that triggered it.** `frame_watcher.rs` drain task set `dirty = false` *before* awaiting `schedule_reconcile`, then ignored the result. A reconcile that failed (locked DB, transient IO) discarded the only record that those filesystem events were never processed, and nothing retried until the next unrelated event — inventory silently stale. This is the same family as the three instances fixed in `astro-plan-hyk0`.

Fixed: `schedule_reconcile` now returns whether the root's pending work is discharged, and `dirty` stays set for a retryable failure. Guarded against the opposite failure mode too — a non-retryable error (an unregistered root yields `RootUnavailable` with `retryable: false`, `frame_inventory/reconcile.rs:222`) discharges, so a permanently-broken root cannot spin the debounce loop. New unit test `a_non_retryable_reconcile_failure_discharges_pending_work`.

**2. Clippy cognitive-complexity ceiling (build failure under `-D warnings`).** `attach_root_watcher` was 42/20. Split the registry-publish protocol into `publish_entry`, which is a genuine seam — resolve-and-validate vs. the lock/tombstone registration. Residual overage in both (24/20, 21/20) is guard-arm `tracing!` macro expansion, not branching, so each carries an `#[expect]` with a stated reason, matching the pattern at `commands/settings.rs:190`.

### Checked and found sound

- **Local-first (constitution I).** The watcher never touches user image files. Every `fs::write`/`remove` in the reconcile path is test-only; the watcher's sole effect is scheduling a pass that reads.
- **Symlink gate (constitution, R6).** Reuses `start_artifact_watcher`, so `follow_symlinks` threads through `notify_bridge::register_watch_paths`. `frame_watcher_symlink_gate.rs` proves both halves — gate closed blocks traversal AND asserts no record is written, gate open lets the identical event through. The non-vacuous framing is the right call.
- **Bounded memory.** The event channel is bounded at 256 (`frame_watcher.rs:242`) with `overflow_flag` promoting a drop to a full reconcile — a dropped event degrades to a rescan rather than losing data. `NeedsRescan` is handled the same way.
- **Attach/detach races.** The detach-during-attach tombstone is correct, including the stale-tombstone clear at attach entry (without it a detach/attach cycle would kill the next attach). Slow work (config read, `is_dir` probe, on-open pass, OS watcher start) runs unlocked; `spawn_blocking` around `is_dir` is right for the stale-NFS case this feature's opt-out targets.
- **No sleep-based tests.** Both integration tests are event-driven — they await a `frame.recovered` envelope on the bus with a deadline, not a fixed sleep. The negative symlink half necessarily waits out a 6s budget to prove absence; that is inherent to asserting a negative, and it is bounded rather than racy.

### Open, not fixed

Nothing demotes a root to polling when live events prove unreliable on it. An OS watcher error or overflow schedules a recovery reconcile, but no code overrides a user's explicit `live: true`. That needs a product call on what evidence justifies the override; the commit message already records it.

### Verification (local, macOS)

- `cargo clippy --workspace --all-targets -- -D warnings` — exit 0, zero warnings
- `cargo fmt --all --check` — clean
- `cargo test -p desktop_shell` — **112 passed, 0 failed**, including `live_file_creation_schedules_a_reconcile_that_recovers_the_frame` and `symlinked_root_is_not_traversed_until_follow_symlinks_is_enabled`
- `cargo test -p contracts_core -p fs_inventory` — **109 passed, 0 failed**
- `pnpm typecheck` — exit 0 (was the CI failure)
- `vitest run src/features/inventory src/features/sessions` — **152 passed** in 16 files
- `biome check apps/desktop/src/features/inventory/` — clean

### Not verified

- Windows and Linux behaviour. The symlink-gate test is `#![cfg(unix)]`, so junction handling on Windows is untested here; only CI will exercise the other two legs.
- Real-UI journeys (L3) and `Build app` were skipped on this run (draft PR).

### Verdict

Mergeable on the merits, subject to CI confirming the typecheck fix across all three legs. The two remaining pre-commit lint failures are inherited from `main` and should be fixed there, not here. The PR body's "tracking record, not a merge candidate" verdict no longer stands — it has now been merged with `main` and reviewed. Still a draft; someone with merge authority should mark it ready.

srobroek:
Your ubuntu failure is **not yours** — it is the `astro-plan-9lkf` flake, and the fix has just landed on main.

Evidence: the failure is `resume_then_retry_of_pre_pause_failed_item_reaches_terminal_state` at `plan_resume_integration.rs:690:5`, twice in the same run, and this PR touches **zero** `plan_apply`/`plan_resume` files. That test waited on `wait_plan_terminal` after triggering a retry, which returns as soon as the *plan* is not `applying` — already true from the pre-retry attempt — so the assertions raced the executor.

Fixed in #1645 (`c2baf1149a05`): it now waits on the retried item reaching `succeeded`. **Merge latest main and the failure goes away.** That merge also picks up two pre-commit hooks that were failing on clean main (`typos` and `end-of-file-fixer` vs the generated `themes.css.ts`).

Worth noting for the record: I filed that flake saying I could not reproduce it locally — 8/8 passes, and a probe showed the plan still `applying` right after the retry — so I fixed the mechanism the code documents rather than claiming a verified catch. **Your run supplied the proof my machine could not.**

## On your review of this PR

Your HIGH-severity finding is the valuable part, and I verified it on the pushed branch: `frame_watcher.rs:190` now reads `dirty = !schedule_reconcile(...)`, so a failed reconcile keeps the flag rather than clearing it before the await. That flag was the only record those filesystem events were never processed — dropping it meant inventory went silently stale with nothing retrying until an unrelated event.

That is the fourth instance of one family this cycle: **work discarded because a marker was cleared before the awaited thing succeeded.** The three cursor cases were `#1644`. Guarding the opposite direction too — a non-retryable `RootUnavailable` discharging so a permanently-unregistered root cannot spin the debounce loop — is the right call; a naive fix there would have traded a silent drop for a hot loop.

Once main is merged and CI is green this looks mergeable to me. The remaining honest gap is that `frame_watcher_symlink_gate.rs` is `#![cfg(unix)]`, so Windows junction handling rests on CI alone.

PR #1637, opened 2026-07-30
perf: benchmark coverage for reconcile, plan progress, and calibration suggestions

srobroek:
## Summary

- Adds benchmark scenarios for reconcile, plan-progress, and calibration-suggest, so regressions in those paths are measurable rather than noticed by feel.

## Provenance and state

This branch was found in a local worktree with **no remote ref containing it** — a reclamation sweep would have destroyed it. It is pushed and PR'd here to preserve and track the work, not because it is ready to merge.

It is 19 commits behind `main` and has had no review. Treat this as the tracking record rather than a merge candidate.

## Before this merges

- Rebase onto `main` (19 behind)
- Independent review — none has happened
- Confirm the scenarios still reflect the current shape of those paths, since `main` has moved considerably. In particular, reconcile changed on 2026-07-30: an unreachable root now aborts the pass before any frame write (#1630), so a reconcile benchmark written before that may measure a code path that no longer exists.


srobroek:
## Verdict: do not merge -- close as superseded

This branch contributes **zero** lines. Its entire content is already merged into `main` as
[#1615](https://github.com/platevault/platevault/pull/1615) (`00729d869`).

After `git merge origin/main` (`5ec68377c`), the branch tree is **byte-identical to
`origin/main`**:

```
git rev-parse HEAD^{tree} == git rev-parse origin/main^{tree}   → identical
gh pr view 1637 --json additions,deletions,changedFiles          → 0 / 0 / 0
```

Every substantive file on this branch is the same blob as `main`:

| File | vs `origin/main` |
|---|---|
| `crates/tools/perf-bench/src/reconcile.rs` | identical |
| `crates/tools/perf-bench/src/plan_progress.rs` | identical |
| `crates/tools/perf-bench/src/calibration_suggest.rs` | identical |
| `crates/tools/perf-bench/src/support.rs` | identical |
| `crates/tools/perf-bench/src/main.rs` | identical |
| `crates/persistence/calibration/src/test_support.rs` | identical |
| `docs/research/007-025-048-perf-gate-measurements.md` | identical |
| `crates/tools/perf-bench/Cargo.toml` | identical |

The rescue sweep preserved a commit that had already been merged through another PR.
`main` also carries a follow-up fix this branch predates (see below), so `main` is
strictly ahead.

## Failing checks: classification

| Check | Class | Evidence |
|---|---|---|
| `Perf baseline (inbox hot paths)` | **real defect** in this commit | `FAIL: plan_apply_progress sqlx_stmts=40116 > baseline=40113 (regression -- query count increased)` -- job 90374246047 |
| `Rust coverage (llvm-cov)` | inherited | `spec062_portable_contracts::relative_path_schema_and_serde_share_utf8_byte_boundaries` FAILED, `-p contract_tests --test contract_schema_parity` -- the AJV/Node-deps gap fixed in #1609, which this branch predates |
| `Unit + integration (L1+L2)` -- ubuntu, windows | inherited | both `CANCELLED`, not failed -- fail-fast cancellation triggered by the two above |

The perf-baseline failure was a genuine defect and is the one review finding that
mattered. It is already fixed on `main` by the second commit in #1615
(`fix(perf): stop gating a statement count that timing makes non-deterministic`),
which adds a `stmts_timing_dependent` opt-out to `scripts/check-perf-baseline.sh:157-161`.

## Review findings

Reviewed as first reviewer against the merge-questions brief. Ordered most severe first.

1. **`plan_apply_progress` statement count is not deterministic** --
   `crates/tools/perf-bench/src/plan_progress.rs:35`. The harness's stated invariant is
   "same fixture size → same query count every run", and this scenario breaks it: the
   count varies with the group-commit flush window
   (`crates/app/core/src/plan_apply/callbacks.rs`, 100 items **or 250 ms**, whichever
   first). Wall-clock timing therefore decides how many flush batches occur, so the
   statement count moves run to run. Confirmed locally at HEAD: the run observed
   `sqlx_stmts=40113` against a `40116` baseline -- a 3-statement swing in the same
   direction that failed CI, with the code unchanged. Ratcheting this as a hard gate
   would have made the perf job a recurring flake, in a repo that already has two open flake beads,
   `astro-plan-9lkf` and `astro-plan-qknj`. `main`'s fix (report but do not gate the count;
   keep the wall-clock WARN) is the correct resolution, and it is the reason `main`'s
   version should win here.

2. **Two of five scenarios are gated on near-zero wall times, so the WARN is pure
   noise** -- `scripts/perf-baseline.json`. `calibration_suggest_1k_masters` has
   `wall_ms_budget: 3`, and the 1.5× rule makes the limit 4 ms. Locally it measured 7 ms
   and warned. A 1 ms scheduling jitter exceeds that limit, so the check reports no useful information.
   WARN-only means it cannot fail a build, so this is a noise complaint, not a
   correctness one -- but it trains readers to ignore the WARN line, which is where a real
   regression would appear. Worth a floor (e.g. skip the wall check under ~50 ms) as
   follow-up; not a merge blocker, and moot for this PR.

3. **Benchmarks do measure what they claim; no work is optimised away and setup is
   excluded.** Each scenario calls a real use case against a real migrated SQLite DB, and
   every timing window wraps only the use-case call -- fixture build, seeding, and
   migrations sit outside it (`reconcile.rs:83-87`, `plan_progress.rs:57-58`,
   `calibration_suggest.rs:85-89`). Results are consumed by `print_result`, so nothing is
   dead-code-eliminated. Statement counting via a `tracing` layer on the `sqlx` target
   (`support.rs:45`) keeps instrumentation out of the production crates, and
   `max_level_hint` (`support.rs:36-38`) correctly stops the registry dropping DEBUG
   events under a stricter `RUST_LOG` -- a real subtlety, handled.

4. **No assertion smuggling.** The scenarios record outcome fields
   (`items_applied`, `plan_state`, `present`, `matches`) into JSON rather than asserting
   behaviour, so nothing that belongs in a test is hiding in the bench. The `expect(...)`
   calls are setup-failure guards, not behavioural assertions.

5. **No temp-file or DB leakage; no new sleeps.** Every scenario owns a `tempfile::tempdir`
   whose guard lives to end of scope (`reconcile.rs:39,42`, `plan_progress.rs:38`,
   `calibration_suggest.rs:33`), and per-scenario DBs keep statement counts independent.
   There is no `sleep` and no nextest retry anywhere in the added code -- the timing
   coupling in finding 1 is indirect (via the flush window), not an added delay.

6. **CI wall-clock cost is real but bounded and correctly scoped.** The added scenarios
   run ~3 min on the CI runner (job 90374246047), dominated by `reconcile_root_frames` at
   ~4 s of measured time plus 10,000-file fixture creation. The job is
   `changed-files`-gated, so it does not run on every PR. The commit is also honest about
   a gap it did not close: the trigger grep only fires for `app_core_inbox` /
   `persistence_inbox` / `fs_inventory`, none of which the three new scenarios cover, so
   their budgets are only enforced when an inbox crate is co-affected. Documenting that in
   the script header rather than silently inheriting it is the right call, though it does
   mean these three gates are mostly dormant -- worth a follow-up to widen the trigger.

## Verification (at merged HEAD `5ec68377c`)

- `cargo clippy --workspace --all-targets -- -D warnings` -- clean, no warnings
- `cargo fmt --all --check` -- clean
- `cargo bench --no-run` -- all bench targets compiled
- `PERF_N=500 scripts/check-perf-baseline.sh` -- exit 0; all 5 scenarios OK on
  statement count, 4 wall-clock WARNs (findings 1 and 2)

Since the merged tree equals `origin/main`, these results verify `main`, not any
unique content on this branch.

## Recommendation

Close without merging. The work is already on `main` via #1615, and merging a
zero-diff branch adds only a merge commit. Bead `astro-plan-9l04` is satisfied by
#1615.

srobroek:
Closing as **superseded** — this branch now contributes zero lines.

Verified independently: `gh pr view` reports **+0/-0, 0 changed files**, and `git diff origin/main..origin/wf_f5d5453f-486-1` is **empty**. The entire content already landed on main as **#1615** (`00729d869`); the rescue sweep preserved a commit that had already merged by another route. Merging a zero-diff branch would add only a merge commit.

The rescue was still worth doing — it was the only way to establish that, and the review it triggered produced two findings worth keeping:

**The perf gate would have become a recurring flake.** `plan_apply_progress`'s statement count is **not deterministic**: it depends on the group-commit flush window (100 items **or 250 ms**), so wall-clock timing decides the batch count. Reproduced locally at `40113` against a baseline of `40116` with code unchanged. Ratcheting that number would have broken the harness's core invariant — "same fixture size → same query count". Already fixed on main by #1615's second commit via a `stmts_timing_dependent` opt-out.

**Sub-millisecond wall budgets make the WARN line uninformative.** `calibration_suggest_1k_masters` carries a 3 ms budget (4 ms limit) and measured 7 ms locally. It's WARN-only so it can't fail a build, but it trains readers to ignore exactly the line where a real regression would appear. Worth either widening the budget to something meaningful or dropping the check.

Also worth recording: of the "4 failing checks", only **2 were real** — the other two were `CANCELLED` by fail-fast, not failures. That distinction matters when triaging a stale PR.

The benchmarks themselves reviewed clean: timing windows wrap only the use-case call, fixtures and migrations are excluded, results are consumed so nothing is optimised away, no assertion smuggling, no temp-file leakage, and no new sleeps or nextest retries.

PR #1638, opened 2026-07-30
fix: concurrent E2E runs no longer load each other's frontend build

### LINKED ISSUE #1409: E2E lanes share one :5173 frontend — devUrl is baked at build time, so concurrent local runs feed each other's builds
Reported in the spec-058 handover (docs/development/handover-058-2026-07-20.md) after it corrupted another lane's run: 'I watched my server feed another lane's run, hanging their inbox journeys.'

SYMPTOM: two worktrees running local Layer-2/Layer-3 E2E at the same time both resolve the frontend at 127.0.0.1:5173. The second lane silently consumes the first lane's build. Failures surface as inbox journeys hanging or asserting against UI that belongs to another branch — i.e. they look like product bugs, not infrastructure collisions, which is what makes this expensive.

ROOT CAUSE: devUrl is baked in at build time, so there is no per-lane override at run time.

AGGRAVATING FACTOR: WSL mirrors ports with Windows, so a WSL lane and a Windows lane collide on the same 5173 too. This is the same family as the known WebView2 shared-localStorage hazard.

SCOPE WARNING: the fix touches build-time port/devUrl config shared by every active worktree (79 at time of filing). Landing it mid-campaign is itself a collision risk. Sequence it for a quiet point, or gate it behind an env var that defaults to today's behaviour.

SKETCH: env-var port override (e.g. ALM_E2E_DEV_PORT) threaded through the Tauri devUrl config and the E2E harness's base URL, defaulting to 5173 so nothing changes unless a lane opts in. Allocate per-lane ports from the worktree name or a lockfile.

VERIFY: the fix must be proven in BOTH directions — two lanes on distinct ports pass concurrently, AND two lanes forced onto the same port still collide (so the test is not vacuous).

srobroek:
## Summary

- Two concurrent local E2E lanes both loaded `localhost:5173`, so the second
  lane's app silently served the first lane's frontend build. The failures read
  as hanging or wrong-branch journeys rather than as a port conflict, which made
  every E2E-based verification result untrustworthy.
- The frontend URL is now a run-time value (`PV_DEV_URL`) instead of a value
  compiled into the app binary, so a lane can put its build and its app on a
  port it actually owns.

Closes #1409.

Tracks-Bead: astro-plan-8y3i
Merge-Bead: astro-plan-2l1c

## The build-time / run-time coupling

`devUrl` is compiled into `desktop_shell` by `generate_context!`, while the port
a lane can claim is only knowable at run time. Passing a different `--port` to
`vite preview` could not fix this on its own: the built app kept requesting the
baked `localhost:5173`.

## Mechanism

`instance_context()` (`apps/desktop/src-tauri/src/lib.rs:300`) already mutates
the generated config, so `devUrl` is overridden there from `PV_DEV_URL`.

One value, three readers, same resolution order:

| Reader | Site |
| --- | --- |
| app binary | `apps/desktop/src-tauri/src/lib.rs:322` |
| harness base URL | `crates/e2e-tests/tests/common/mod.rs::app_url` |
| `vite preview --port` | `.github/workflows/e2e.yml`, 4 serve steps |

`app_url()` falls back to `build.devUrl` parsed from an `include_str!` of
`tauri.conf.json`, not a second literal. `rg -n 5173` over `e2e.yml`,
`crates/e2e-tests/`, and `apps/desktop/src-tauri/` leaves `tauri.conf.json:8`
and the workflow `env` default.

`--strictPort` on every serve step turns a taken port into exit 1. Without it
`vite preview` picks a different port while the app keeps requesting
`PV_DEV_URL`.

`just e2e-dev-url` derives a per-checkout port, so running two local lanes does
not require hand-picking one.

The override is compiled only under the `e2e` feature, so a release binary's
frontend origin is not environment-redirectable (Constitution Principle V).

CI behaviour is unchanged: each job owns a GitHub-hosted VM, so its port only
has to be unique per machine, and the workflow default is the previous value.

## Verified locally

- `.github/workflows/e2e.yml` parses (`yaml.safe_load`, 13 jobs).
- `--strictPort` collision: a second `vite preview` on a bound port exits 1.
- `cargo clippy -p desktop_shell --features e2e`, `cargo clippy -p desktop_shell`
  (release feature shape), `cargo clippy -p e2e_tests --all-targets`: clean.
- `cargo fmt --all --check`: clean.
- `pnpm --filter @astro-plan/desktop build`: builds.
- `just e2e-dev-url` is stable per worktree and distinct across four real
  worktrees (5202 / 5347 / 5409 / 5428).

## Needs a real matrix run to confirm

The end-to-end claim is untested. A matrix run must show every shard green with
the app loading `$PV_DEV_URL`. The bead's negative direction (two lanes forced
onto ONE port must still collide, so the test is not vacuous) is also unrun;
`--strictPort` now makes that case exit 1 at serve time instead of producing a
phantom journey failure.

## Not fixed by this change

- A lane that does not export `PV_DEV_URL` keeps the `tauri.conf.json` default,
  so two such lanes still target one port. `--strictPort` makes the second
  serve fail loudly instead of contaminating the first.
- WSL/Windows port mirroring is addressed only insofar as the two sides pick
  different `PV_DEV_URL` values.
- WebView2 shared-localStorage is out of scope (handled by
  `WEBVIEW2_USER_DATA_FOLDER`).
- Hash-derived ports can collide; `--strictPort` surfaces that as exit 1.

## Spec Context

Spec 037 Layer-2/Layer-3 real-UI E2E harness.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -185,6 +185,15 @@ env:
   # in every shard job (ubuntu, windows, macOS) so a single bump here
   # invalidates the cached binary across all lanes at once.
   TAURI_WEBDRIVER_VERSION: "0.1.1"
+  # THE frontend URL for this workflow. Every "Serve frontend dist" step derives
+  # its `vite preview --port` from this value, the app binary reads it as its
+  # `devUrl` (`desktop_shell::DEV_URL_ENV`), and the harness reads it as its
+  # base URL (`crates/e2e-tests/tests/common/mod.rs::app_url`) — one value,
+  # three readers, so what is served and what is loaded cannot drift apart.
+  # Each job runs on its own GitHub-hosted VM, so the port only has to be
+  # unique per machine; concurrent LOCAL lanes export their own `PV_DEV_URL`
+  # (issue #1409).
+  PV_DEV_URL: "http://localhost:5173"
 
 jobs:
   # Cheap change-detection gate — mirrors ci.yml's `changes` job. Cross-workflow
@@ -746,9 +755,15 @@ jobs:
         shell: bash
         run: chmod +x "$APP_BIN"
 
-      - name: Serve frontend dist on :5173
+      - name: Serve frontend dist on $PV_DEV_URL
         shell: bash
-        run: pnpm --filter @astro-plan/desktop preview --port 5173 &
+        # Port comes from PV_DEV_URL (workflow `env`) — the same value the app
+        # binary and the harness read, so there is no second port literal to
+        # drift. --strictPort makes a taken port a loud failure instead of a
+        # silent fallback that leaves the app loading someone else's server.
+        run: |
+          pnpm --filter @astro-plan/desktop preview \
+            --port "${PV_DEV_URL##*:}" --strictPort &
 
       - name: Run L3 smoke (3 journeys via filter-expr)
         env:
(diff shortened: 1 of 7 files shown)


srobroek:
Superseded by **#1639**, which re-applies this change onto the post-#1595 module layout.

This branch cannot merge: `main`'s `crates/e2e-tests/tests/common/mod.rs` was split into domain modules by #1595, while this branch edited the pre-split monolith. `git merge` produces one conflict hunk spanning ~2,900 lines — the entire file. Taking either side loses something real: HEAD drops the module split, `origin/main` drops the devUrl fix.

#1639 preserves everything that mattered here and corrects one placement guess along the way — `app_url()` belongs in `common/boot.rs` (which owns process-scoped boot config) rather than `common/app.rs`.

Verified on #1639: exactly one `5173` literal survives, the `e2e`-feature gating is intact and *proven* (renaming `apply_dev_url_override` breaks only `--features e2e`; a default-only check would have proven nothing), and clippy is clean in all three shapes.

The work here was sound — only its base was stale.

PR #1640, opened 2026-07-30
ci: run the OS-agnostic unit gates on ubuntu only

srobroek:
## Summary

- The `Unit + integration (L1+L2)` matrix ran every frontend gate, plus the generated-drift gate, on ubuntu, macOS and Windows. Those steps produce the same verdict on every platform.
- Six steps now carry `runner.os == 'Linux'`: `Frontend unit tests`, `Frontend lint (Biome + ESLint gates + tokens)`, `Test-suite lint (Biome over tests/)`, `TypeScript typecheck`, `Generated contract/binding drift`, and `Cache Vite transform/deps output`.
- `Rust tests` keeps all three platforms.
- A frontend-only change now leaves the macOS and Windows legs with every step skipped, so those legs run a `No platform-specific work on this leg` notice step. A bare green on a leg that ran nothing reads as platform coverage that does not exist.

## Measured before

Per-step minutes from 9 completed `ci.yml` runs (30551998486, 30548827092, 30547363167, 30545428802, 30544462531, 30540563538, 30539128691, 30541975516, 30542973252):

| step | ubuntu | macos | windows |
| --- | --- | --- | --- |
| Frontend unit tests | 5.8 | 8.2 | 7.6 |
| Frontend lint | 0.7 | 1.0 | 0.9 |
| TypeScript typecheck | 0.4 | 0.6 | 0.5 |
| Generated contract/binding drift | 1.4 | 2.0 | 2.3 |
| Rust tests | 3.7 | 5.1 | 6.0 |

Job totals on a frontend PR: ubuntu 16-18 min, macOS 18-20 min, Windows 22 min. On the two Rust-only runs (30541975516, 30542973252) the frontend steps already reported `skipped / 0s`.

## Expected after

macOS drops ~11.8 min and Windows ~11.3 min per frontend PR, for ~23 min less matrix compute. Wall clock follows the longest leg, so a frontend PR's critical path moves off the 22 min Windows leg to the 16-18 min ubuntu leg. A Rust-only PR loses the duplicated drift gate, ~4.2 min of compute.

## Gate equivalence

- Every affected gate still runs once, on ubuntu. No gate lost its only runner.
- Job `name:` fields, the `integration` job-level `if:`, and the `[ubuntu-latest, windows-latest, macos-latest]` matrix are byte-identical to `main`, verified by parsing both revisions. The same six required contexts are published.
- `Rust tests` covers all three platforms: real SQLite, real migrations, per-OS path semantics.
- Windows keeps `Windows workspace check (all targets)`, its only all-targets compile of non-test code.
- The vitest suite runs under jsdom. `apps/desktop/src/lib/reveal-label.test.ts` and `apps/desktop/src/features/sessions/__tests__/SessionsPage.inventory.test.tsx` are the only specs that read a platform, and both inject a fake `navigator`. No snapshot files exist under `apps/desktop/src`.
- `.gitattributes` pins `*.ts`, `*.tsx`, `*.css`, `*.rs` and `specs/**/contracts/*.generated.json` to `eol=lf`, so the drift gate's `git diff --exit-code` cannot differ per platform.

## Collisions

- #1604 (open) edits the `integration` job-level `if:` and appends two ratchet steps after `Hot-read ratchet`. No semantic overlap with these steps.
- #1620 (open) narrows one ratchet's `if:`. No overlap.
- #1638 is closed and touched `e2e.yml` only.

## Left alone

`frontend_full` was true on 5 of the 7 frontend runs sampled, so `vitest related` rarely narrows the suite in practice. Narrowing the `frontend_full` triggers would trade gate strength for time, so this PR does not touch it.

Tracks-Bead: astro-plan-4oyu
Merge-Bead: astro-plan-ydfj


srobroek:
The `GITHUB_TOKEN` permissions finding is valid but **pre-existing, not introduced here** — filed separately as `astro-plan-j14r` (P2) so it is not lost when this merges.

Verified against `origin/main`: `ci.yml` has 0 hits for a top-level `permissions:` key, `e2e.yml` has 0, and `release-gate.yml` has 1. So the pattern already exists in this repo and these two workflows are the outliers. This PR does not add or remove a `permissions` block; it only gates existing steps to Linux.

Narrowing the token is worth doing, and I noted the check that must come first: confirm no job currently relies on an inherited write scope, since a step that comments on a PR, pushes a tag, or uploads to a release would break silently if the block is added without an exception.

Merging this on its own merits: measured savings, job names byte-identical to `origin/main` (diffed, not eyeballed — a rename would deadlock every PR), and the 3-OS Rust matrix untouched.

PR #1643, opened 2026-08-17
fix: project manifests and staleness updates no longer go missing on busy libraries

srobroek:
Unblocks #1591 by fixing the blocker that PR introduces (`astro-plan-hyk0`). Targets `fix/events-replay-followups` so #1591 can merge once this lands in it.

## The defect

Both new subscribers advance the replay cursor on the **live** event path with `max_event_id`, which is the **table-wide max**, not the row just handled:

```rust
handle_workflow_run_event(&pool, &bus, &env.payload).await;
if let Ok(id) = events::max_event_id(bus.pool()).await { cursor = id; }
```

Any event inserted between the broadcast and that query is **jumped**. A later `Lagged` replay starts *after* it, so the event is **lost permanently**, not merely late. The window is open exactly when concurrent writers are active — which is when replay matters most.

Sites: `crates/app/projects/src/project_manifests.rs` and `crates/audit/src/stale_propagator.rs`.

This is a **regression this PR chain introduces**. `origin/main` has zero occurrences of the pattern; its only cursor writes are the replay assignment and `new_cursor = row.event_id`.

## Why the obvious fix isn't available

Advancing to the handled event's own id **cannot be written**: `EventEnvelope` (`crates/audit-types/src/event_bus.rs:38`) carries `contract_version`, `topic`, `source`, `emitted_at`, `payload` — and **no `event_id`**. A live subscriber cannot know its own row. Any `env.event_id` suggestion is wrong until the envelope gains that field.

## The fix

Drop the live-path advance at both sites; let the `Lagged` branch own the cursor, which already moves per row to `row.event_id`. That's what main does today.

**The cost, stated plainly:** a replay re-walks from an older cursor and re-delivers some already-handled events. Both handlers are idempotent — hooks by the research.md §6.1 contract, and manifest `write()` produces a new file per call — so re-doing work is strictly better than skipping it. This is the same tradeoff #1623 accepted in the log forwarder, deliberately.

Achieving the PR's original goal *without* the jump would need `event_id` on the envelope — a contract type with `JsonSchema`/`Type` derives, so it touches the generated surface and every construction site. Worth its own change if the re-walk cost is ever shown to matter.

## Verification

Both regression tests are **mutation-verified** — reinstating the `max_event_id` advance makes them fail:
- `stale_propagator`: 3 dispatches collapse to **1**
- `project_manifests`: the stranded manifest is **never written**

They're deterministic by *ordering*, not by racing: the never-broadcast row is inserted **first**, so it always holds the lower `event_id`.

I **replaced** `cursor_advances_after_live_events_so_next_lag_is_bounded`, which asserted the *defective* behaviour — it required the replay to skip rows below the jumped cursor. A test encoding the bug had to go, not be worked around.

`cargo test -p app_core_projects -p audit`: 93 + 17 pass. `workflow_run_manifest_e2e`: passes. `cargo fmt` + clippy clean.

### DIFF
--- a/crates/app/projects/src/project_manifests.rs
+++ b/crates/app/projects/src/project_manifests.rs
@@ -422,9 +422,17 @@ pub fn spawn_workflow_run_subscriber(
 
     let mut rx = bus.subscribe();
     tokio::spawn(async move {
-        // Cursor starts at 0: replay from the beginning on first lag.
-        // write() is idempotent (produces a new file per call), so replaying
-        // historical workflow.run_completed events is safe.
+        // Cursor starts at 0, so the first lag replays from the beginning; every
+        // lag thereafter resumes from the high-water mark the replay leaves
+        // behind, and the pruner bounds how far back that can reach.
+        //
+        // Replay is SAFE but not free, and `write()` is not idempotent: it
+        // inserts a `manifests` row with a fresh id per call, so a replayed
+        // event adds a row the user can see in the manifest list. Same-second
+        // replays collapse on disk (the path is second-granular and
+        // `write_manifest_file` refuses to overwrite), but the DB rows do not.
+        // Duplicate rows beat a silently missing manifest, which is what
+        // advancing this cursor on the live path used to cause.
         let mut cursor: i64 = 0;
 
         loop {
--- a/crates/app/projects/src/project_manifests.rs
+++ b/crates/app/projects/src/project_manifests.rs
@@ -430,13 +430,15 @@ pub fn spawn_workflow_run_subscriber(
         loop {
             match rx.recv().await {
                 Ok(env) if env.topic == TOPIC_WORKFLOW_RUN_COMPLETED => {
+                    // The live path deliberately leaves `cursor` alone. Advancing it
+                    // here to `max_event_id` would read the table-wide max rather
+                    // than this event's row: any event inserted between the
+                    // broadcast and that query gets jumped, so a later `Lagged`
+                    // replay starts *after* it and it is lost rather than late.
+                    // The envelope carries no `event_id`, so this branch cannot
+                    // know its own row to advance to. Only the replay below moves
+                    // the cursor, and it does so per row.
                     handle_workflow_run_event(&pool, &bus, &env.payload).await;
-                    // Advance cursor so the next lag replay doesn't restart from 0.
-                    if let Ok(id) =
-                        persistence_lifecycle::repositories::events::max_event_id(bus.pool()).await
-                    {
-                        cursor = id;
-                    }
                 }
                 Ok(_) => {}
                 Err(RecvError::Lagged(n)) => {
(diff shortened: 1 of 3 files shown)


srobroek:
## Independent review: APPROVE, with one correction I have applied

A reviewer briefed to **refute** each claim confirmed the defect, the fix shape, and the test-deletion call — but caught a real error in my reasoning, now fixed in `d97d70a41`.

**My idempotence claim was half false.** I wrote that manifest `write()` is "idempotent (produces a new file per call)" — which is self-contradictory: producing a new artifact per call is the *definition* of non-idempotent. `insert_manifest` is a plain INSERT with a fresh id, no upsert, and there is no manifest pruner, so a replayed event adds a `manifests` row the user sees in the list. Same-second replays collapse on disk (second-granular path, `write_manifest_file` will not overwrite) but **DB rows never collapse.**

The tradeoff still stands — duplicate rows beat a silently missing manifest — but the honest statement is *"duplicate manifest rows on replay"*, not *"idempotent"*. The comment now says that. Also: research.md §6.1 idempotence binds **hooks**, and the manifest subscriber does not use hooks, so it never covered this path. Only the stale-propagator half was contract-backed.

**My main worry did not survive, in the good direction.** I asked the reviewer to attack unbounded re-walk cost. It does not exist: the cursor is a high-water mark that *ratchets* on every lag, so only the **first** lag walks history — and that is bounded twice, by topic filter and by the 90-day pruner. Production channel capacity is 256, so lags are rare.

**Added the missing half of the invariant.** My tests only proved the cursor never runs *ahead*; nothing proved it does not stay *behind* — which is exactly what makes this affordable. New test `a_second_lag_replays_only_what_arrived_since_the_first`, mutation-verified: discarding the replay returned cursor makes the second lag re-walk everything, 3 → 5.

**No third site.** The reviewer swept the other `Lagged` handlers: `onboarding.rs:194` and `watcher.rs:96` keep no cursor, and `log.rs` advances per row via `cursor.max(row.event_id)` — correct shape.

Also dropped a dangling forward reference in `pruner.rs` to "cursor-advancement on live events", the mechanism this PR removes.

### CI status — both failures are inherited from the base branch, not from this change

- **Rust coverage (llvm-cov)**: fails on `spec062_portable_contracts`, which spawns a Node AJV validator. The base branch is **43 commits behind main** and predates #1609, which added the `pnpm install` this job needs. `coverage.yml` on the base has **zero** `pnpm install`/`setup-node` steps; main has 2.
- **ubuntu L1+L2**: same base-branch staleness.

Neither is reachable from a change touching two Rust files and their tests. They clear once #1591 rebases onto main, which it needs anyway.

Verification after the follow-up: `audit` 18 pass, `app_core_projects` 93 pass, e2e passes, clippy **0 warnings**, fmt clean.

srobroek:
Superseded by #1644, which targets **main** instead of a 43-commits-behind branch. Same commits, plus the merge of current main, verified together: 265 tests, workspace clippy clean.

The merge surfaced something that strengthens this fix: **main independently deleted `max_event_id` outright** and added `rewound_cursor`, which seeds a bounded cursor without reading the table max. Two independent routes reached the same conclusion that reading the table-wide max on the live path was wrong. This branch removed the function's last two callers, so the only thing left referencing it was a stale test import.

Closing in favour of #1644.

PR #1647, opened 2026-08-17
fix: retrying a failed item no longer silently reports it as cancelled

srobroek:
Fixes `astro-plan-ts1z` (P1). A **Tier-1 custody failure** under constitution II: an approved action recorded as terminal that never happened.

## Summary

- Retrying a failed plan item could silently become a **cancellation** — the file never moved, and the audit trail read `cancelled`, indistinguishable from a cancellation the user had asked for.
- Losing that race is now reported as a recoverable error, and the item is left `failed` so it can genuinely be retried again.

## The defect

`retry_plan_item` flipped the item to `applying` and only then enqueued it for the executor:

```rust
apply_repo::item_retry_applying(pool, item_id, plan_id).await?;
let registry = active_runs();
if let Some(run) = registry.get(plan_id) {   // <-- run gone? do nothing.
    run.retry_queue.push(item_id);
}
```

If the run finished in that window, the `if let Some` did nothing and the item was left `applying` with **nothing that would ever execute it**. Run completion then sweeps orphaned `applying` items to `cancelled` (`terminal.rs`), converting the user’s retry into a silent cancellation.

The pre-existing `active_runs()` check closes the common case, and its own comment acknowledged it could not close this window. The window is now handled rather than swallowed.

## The fix

Roll the DB flip back before returning an error, via the new `item_retry_rollback_to_failed`. The item stays `failed` — retryable, and honest about not having run — instead of stranded `applying`. The error is marked **recoverable** and names plan-level retry as the way forward.

Details that matter for correctness:
- **`items_failed` is restored too**, since the plan’s terminal state is computed from that count.
- **Guarded on `item_state = applying`**, so it cannot clobber an item the executor did pick up — if the retry really ran, the rollback is a no-op.
- **Only re-increments when it actually moved the row**, so repeating it cannot inflate the counter. The test asserts this idempotence directly.
- **`failure_reason` left NULL deliberately.** The old reason described the original failure, which this retry did not reproduce; restoring it would attribute a failure the app never observed.
- If the rollback itself fails, it logs loudly and **still returns the error** — the caller must never be told a retry was accepted when it was not.

## Verification

**Mutation-verified:** making the rollback leave the item `applying` fails the new test on exactly the stranded state the sweep would cancel (`left: "applying", right: "failed"`).

`cargo test -p app_core --lib`: **351 pass**. Clippy clean on `app_core` and `persistence_plans`. `cargo fmt` clean.

**Bounding what the test proves, plainly:** the race window lives inside one function and cannot be opened from a test without adding a production seam purely for testing — a worse trade. So the test drives the repair path directly rather than winning a race. It proves the rollback is correct and idempotent; it does not reproduce the interleaving.

Related: `astro-plan-9lkf` (the test flake that led here, fixed in #1645).

### DIFF
--- a/crates/app/core/src/plan_apply/lifecycle.rs
+++ b/crates/app/core/src/plan_apply/lifecycle.rs
@@ -655,14 +655,55 @@ pub async fn retry_plan_item(
     // Transition item back to applying in DB (failed → applying).
     apply_repo::item_retry_applying(pool, item_id, plan_id).await.map_err(db_err)?;
 
-    // Register in the retry queue so the executor re-executes it.
-    {
+    // Enqueue for re-execution, and REFUSE if the run vanished in the meantime.
+    //
+    // The `active_runs()` check above closes the common case but not the window
+    // between it and here. Previously that window was swallowed: `if let
+    // Some(run)` simply did nothing when the run was gone, leaving the item
+    // `applying` in the DB with nothing that will ever execute it. Run
+    // completion then sweeps orphaned `applying` items to `cancelled`
+    // (`terminal.rs`), so the user's retry became a silent cancellation --
+    // file unmoved, audit trail reading `cancelled`, indistinguishable from a
+    // cancellation they asked for. That is a Tier-1 custody failure under
+    // constitution II: an approved action recorded terminal that never ran.
+    //
+    // Losing the race is now reported instead. The DB flip is rolled back
+    // first so the item stays `failed` -- retryable, and honest about not
+    // having run -- rather than stranded `applying`.
+    let enqueued = {
         let registry = active_runs();
-        if let Some(run) = registry.get(plan_id) {
+        let queued = registry.get(plan_id).is_some_and(|run| {
             run.retry_queue.push(item_id);
-            drop(run);
-        }
+            true
+        });
         drop(registry);
+        queued
+    };
+
+    if !enqueued {
+        // Best-effort restore. If this fails the item is left `applying` and
+        // the completion sweep will cancel it -- the pre-existing behaviour --
+        // so log loudly and still return the error: the caller must not be
+        // told a retry was accepted when it was not.
+        if let Err(e) = apply_repo::item_retry_rollback_to_failed(pool, item_id, plan_id).await {
+            tracing::error!(
+                error = %e,
+                plan_id,
+                item_id,
+                "retry lost the race with run completion AND could not restore the item to \
+                 'failed'; it will be swept to 'cancelled' and must be retried at plan level"
+            );
+        }
+        return Err(ContractError::new(
+            ErrorCode::RunNotFound,
+            format!(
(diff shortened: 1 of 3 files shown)


srobroek:
CI update — and the failure is **informative rather than a problem with this PR**.

`resume_then_retry_of_pre_pause_failed_item_reaches_terminal_state` failed with:

```
poll_until timed out after 10 s: plan item …-item-0 never reached 'succeeded'
```

That is a **timeout, not an assertion failure** — precisely the behaviour the `wait_item_state` helper from #1645 was built to produce. Its closing note predicted it verbatim: *"a CI timeout at this test now points at ts1z."* The instrument worked: it converted a silent wrong answer into a visible failure.

## But it is a THIRD window, which this PR does not fix

This PR handles *"the run is gone when the retry enqueues"*. Here the retry is **accepted and enqueued, then never executed**.

Cause: `resume_plan` (`lifecycle.rs:441`) constructs a **fresh `RetryQueue`** and registers a new `ActiveRun`, replacing the old registry entry. A `retry_plan_item` landing between that registration and the executor start pushes onto whichever queue it found — and if that is the pre-swap entry, the fresh queue discards it. Item left `applying`, then swept to `cancelled`: same user-visible outcome as this PR fixes, by a different route.

Worth noting the code says this is deliberate — `lifecycle.rs:436-439`: *"Cancel/skip/retry state does not carry over a pause boundary; a fresh set is correct here."* That is defensible for state predating the pause. It does **not** cover a retry issued *after* resume was called: the user's intent is current, not stale, and dropping it silently is the same custody failure.

So `astro-plan-ts1z` is three windows:
- **(a)** run gone at enqueue → **fixed here**
- **(b)** retry enqueued onto a queue `resume_plan` then replaces → **open**
- **(c)** enqueued on the live queue but the run completes before the executor drains it → the pre-existing GFD-2 sweep window → still `cancelled`

## What I suggest

Land this PR: (a) is a genuine fix, and it makes (b)/(c) **visible instead of silent**, which is progress on its own. I have recorded on the bead that it must **not** be closed when this merges, since the headline claim stays true via (b) and (c).

The closing fix for (b)/(c) is to make retry acceptance and executor readiness one atomic step — either durably queued to a run that will drain it, or refused. That needs an ordering decision rather than a patch, so it should not be smuggled into this PR.

I will rerun to confirm the failure is this race and not something introduced here.

srobroek:
Rerun is **fully green — 15/15 checks, CLEAN**.

That confirms the earlier failure was the race this PR is about, not something introduced here: same head, same code, and the timeout did not reproduce. Consistent with a window that needs a loaded runner to open — which is also why I could not reproduce it locally (8/8 passes) and said so rather than claiming a verified catch.

The three-window analysis from my previous comment stands. This PR closes **(a)**; **(b)** and **(c)** remain open on `astro-plan-ts1z`, which I have marked must-not-close on merge.

Issue #742, opened 2026-07-14
Mid-run per-item retry never re-executes — item stuck in 'applying' forever (spec 025 FR-005/US4-AS2)


srobroek:
## What
`retry_plan_item` correctly flips the DB row `failed -> applying` and decrements `items_failed` (`persistence/db/src/repositories/plan_apply.rs:434-454`) and pushes into an in-memory `RetryQueue` (`plan_apply.rs:2020-2028`). But nothing ever consumes that queue to re-execute: `crates/fs/executor/src/run.rs:561-566` only calls `retry_queue.take(&item.id)` for the item currently being processed, purely to clear stale entries (its own comment: "Here we just clear the queue entry to avoid phantom state"). The executor loop walks a fixed Vec once, forward — an already-passed failed item is never revisited.

## Why it matters
A mid-run retry leaves the item permanently stuck in `applying`: no further transition ever occurs, terminal counters undercount vs `itemsTotal`, and the spec's own data-model invariant is violated ("A plan in a terminal state has no item in `applying`", `specs/025-filesystem-plan-application/data-model.md:140`). tasks.md T042 claims the retry-injection wiring is done; T038's own note admits it is untested end-to-end — this pass confirms it is untested because it is broken.

## Classification
Backend (executor loop / RetryQueue consumption).

Found by the 2026-07-14 full-spec verification sweep (run-052fix); full report in the run store.

PR #1652, opened 2026-08-17
fix: a retry you request is no longer silently reported as cancelled

srobroek:
Closes two of the three windows recorded on `astro-plan-ts1z`. Window (a) landed
in #1647; this covers (b) and (c). The bead stays open pending review.

## What changed

`retry_plan_item` could accept a retry that no executor would ever drain. The
item's row stayed `applying`, and run completion swept it to `cancelled`: the
file unmoved, the audit trail reading as a cancellation the user asked for.

- `RetryQueue` has a one-way open to closed lifecycle. `push` returns whether the
  id was accepted; a closed queue refuses.
- `execute_plan` closes the queue on every exit path. On the completion path it
  closes jointly with a final drain: `close_if_empty` checks and closes under one
  lock acquisition, so a push either lands before the close and is executed, or
  is refused after it.
- `retry_plan_item` honors a refusal. It rolls the item back to `failed` and
  returns a recoverable error instead of reporting a retry it could not deliver.
- The cancel and pause paths close the queue and report undrained ids as orphans.
  `restore_unexecuted_retries` puts each row back to `failed` with a
  `retry.not_executed` audit reason before the completion sweep runs.

Window (b), a retry landing while `resume_plan` swaps the registry entry, is
closed: the pre-pause queue is already closed, so such a retry is refused with the
item left `failed` rather than accepted onto a queue the swap discards.

Window (c), a retry enqueued onto the live queue before the run completes, is
closed for the completion path by the joint close-with-final-drain. On the cancel
and pause paths the retry cannot execute, so the outcome is the distinguishable
one the bead names as the minimum: the item reads `failed` and retryable, with
`retry.not_executed` recording why, in place of `cancelled`.

Constitution II is unaffected. The item's `applying` row is still written before
any filesystem action is attempted.

## Test plan

- `cargo test -p fs_executor`: 123 pass, up from 118.
- `cargo test -p app_core --lib`: 352 pass, up from 351.
- `cargo test -p app_core --test plan_resume_integration`: 9 pass, 15 of 15
  consecutive runs, with no `nextest` retries and no sleep-based waits.
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- `cargo fmt --all --check`: clean.

Each new test was mutation-verified by reinstating the defect. Every one failed on
an assertion:

| Mutation | Failure |
|---|---|
| Remove the joint close-with-final-drain | `succeeded left: 2, right: 3` |
| `push` always accepts | "a closed queue must REFUSE…" |
| `close` drops instead of orphaning | `left: [], right: ["item-1"]` |
| `close_if_empty` ignores contents | "closing over a queued retry would drop it" |
| `restore_unexecuted_retries` is a no-op | `item_state left: "applying", right: "failed"` |

`resume_then_retry_of_pre_pause_failed_item_reaches_terminal_state`, which timed
out on CI when the race fired, passed 15 of 15 locally. Local green is weak
evidence here: the bead records that the race never reproduced locally at all, so
CI on this branch is the real check.

Tracks-Bead: astro-plan-ts1z
Merge-Bead: astro-plan-ciqn

### DIFF
--- a/crates/app/core/src/plan_apply/apply.rs
+++ b/crates/app/core/src/plan_apply/apply.rs
@@ -267,6 +267,57 @@ pub(super) async fn cumulative_counts(
     }
 }
 
+/// Reason code recorded on an item whose retry was accepted but whose run
+/// halted before executing it. Distinguishes it from a user cancellation for
+/// any surface that offers the retry again (astro-plan-ts1z window c).
+pub(super) const RETRY_NOT_EXECUTED_REASON: &str = "retry.not_executed";
+
+/// Put every retry this run accepted but never executed back to `failed`, and
+/// append an audit event naming why.
+///
+/// `execute_plan` closes the retry queue before returning, so this set is
+/// final: no further push can land. Restoring here — before the terminal
+/// handlers' `cancel_orphaned_applying_items` sweep — is what keeps a user's
+/// unexecuted retry retryable instead of terminal.
+pub(super) async fn restore_unexecuted_retries(
+    pool: &SqlitePool,
+    plan_id: &str,
+    run_id: &str,
+    retry_queue: &RetryQueue,
+) {
+    for item_id in retry_queue.take_orphaned() {
+        match apply_repo::item_retry_rollback_to_failed(pool, &item_id, plan_id).await {
+            Ok(()) => {
+                let _ = apply_repo::append_event(
+                    pool,
+                    &new_id(),
+                    run_id,
+                    plan_id,
+                    Some(&item_id),
+                    "applying",
+                    "failed",
+                    &Timestamp::now_iso(),
+                    Some(&apply_repo::EventFailure {
+                        code: RETRY_NOT_EXECUTED_REASON,
+                        message: "the run stopped before this retry was executed; \
+                                  the item was not retried and remains retryable",
+                        recoverable: true,
+                    }),
+                    None,
+                )
+                .await;
+            }
+            Err(e) => {
+                tracing::error!(
+                    error = %e, plan_id, item_id,
+                    "could not restore an unexecuted retry to 'failed'; it will be swept to \
+                     'cancelled' and must be retried at plan level"
+                );
+            }
+        }
+    }
+}
+
 /// Drive `execute_plan` to completion/cancellation/pause on a background
 /// task and persist the outcome (terminal state, audit trail, long-op
 /// projection). Extracted from `apply_plan`'s inline `tokio::spawn` block so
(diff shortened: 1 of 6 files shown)


srobroek:
**Correction to this PR's own description — it overstates the result.**

I wrote "Closes the two remaining windows". The accurate verdict is **PARTIAL**:

- **Window (b) — fully closed.** And it needed no `resume_plan`-specific code: the pre-pause queue is already closed by its own exit, so a retry landing during the swap finds no entry or a closed one and is refused with the item left `failed`.
- **Window (c) — closed on the completion path; made-visible-only on the cancel/pause paths.** There the retry genuinely cannot execute, so undrained ids become orphans and `restore_unexecuted_retries` restores each row to `failed` with `failure_code = retry.not_executed` **before** the sweep. `cancel_orphaned_applying_items` then finds nothing.

That is the bead's stated *minimum* for (c) — distinguishable rather than impossible — so `astro-plan-ts1z` stays open.

## Why option 1 and not the others

**Rejected option 2** (block completion over an in-flight retry): makes run termination depend on an unbounded user action.
**Rejected option 3** (let the queue survive the swap): hands ids to an executor built from a *different* item set, since `resume_plan` recomputes `executor_items` — that is the #742 stuck-`applying` bug by another route.

## Mutation verification — five mutations, every one failed on an assertion, never a timeout

| Mutation | Failure |
|---|---|
| Remove joint close + final drain | `succeeded left: 2, right: 3` |
| `push` always accepts | "a closed queue must REFUSE…" |
| `close` drops instead of orphaning | `left: [], right: ["item-1"]` |
| `close_if_empty` ignores contents | "closing over a queued retry would drop it" |
| `restore_unexecuted_retries` no-op | `item_state left: "applying", right: "failed"` |

## Tests

fs_executor **123** (was 118), app_core `--lib` **352** (was 351), `plan_resume_integration` **15/15 consecutive**. No nextest retries, no sleep waits. Clippy and fmt clean.

## What is not verified

**The race itself.** 15/15 locally, but this test never reproduced the race locally in the first place — so local green is weak evidence the windows are shut on a loaded runner. **CI on this PR is the real check.**

Separately: `cargo test -p persistence_plans` standalone fails on `unresolved import persistence_core::test_support`. Confirmed pre-existing on main via `git stash`, unrelated to this change — it needs workspace-level feature unification.

## For whoever touches this next

`drain_retries` works from a **snapshot** — `drain_all` removes ids before executing them, so an id pushed *during* a re-execution is not in the batch being drained. That is why a single post-loop drain is insufficient and the drain/close must loop. **Any halt part-way through a drained batch must hand the remainder to `orphan()`**, or those rows leak as `applying`.

PR #1656, opened 2026-08-19
fix: matching settings accepted an out-of-spec mosaic rotation cap

srobroek:
## Summary

- Matching settings accepted a mosaic residual-sky-rotation cap of 90 degrees.
  The supported value is 10; anything else is now reported as a red fixed-rule
  change, so the settings screen stops passing configurations the matcher cannot
  honor.
- Removed a second, divergent copy of matching-settings validation so both the
  settings screen and the contract surface apply one set of rules.

## Why the wrong value shipped

`tests/contract/spec062_portable_contracts.rs` had no `[[test]]` entry in
`tests/contract/Cargo.toml`, so its 20 tests never compiled and never ran. The
test is now registered, along with the fixture that encoded the wrong cap. Two
sibling files in the same directory are still undeclared, tracked separately.

## Coverage

Ported the adjacent-float bounds test into the contracts suite and widened it
from 8 fields to all 12 range-checked ones, so removing the duplicate validator
doesn't drop boundary coverage. Mutation-checked the new regression test: it
fails when the old value is restored.

Kept the relation-evaluation engine (`evaluate`, `AutomaticRelation`,
`RelationContext`, `GeometryEvidence`, `ThresholdMeasurement`,
`CompleteLinkage`), none of which was duplicated.

## Verification

`just test` (41 suites, 0 failures), `just lint` (exit 0), `just check`
(dead-callers 148 baselined with no new entries, db-boundary sealed at 0,
hot-read 6/6, lifecycle-strings 0, generated-contract diff clean, typecheck
pass).

## Spec Context

`specs/062-session-heterogeneity/contracts/matching-settings.md:17` fixes
`residualSkyRotationCapDeg` at 10 and gives it no bounds-table row.

Closes astro-plan-pssh

Tracks-Bead: astro-plan-pssh
Closes-Bead: astro-plan-pssh
Merge-Bead: astro-plan-prqn


srobroek:
@coderabbitai full review

srobroek:
@coderabbitai review

srobroek:
@coderabbitai review

PR #1657, opened 2026-08-19
fix: relation evidence is optional on panel group and mosaic revisions

srobroek:
## Summary

Adds three items to the 062 contract spec ahead of implementation: the
`relation_evidence.query` operation, a required `expectedMatchingSettingsRevision`
request field on `relation_proposal.manual.create`, and the
`relation_evidence.not_found` error code.

Documents `relation_evidence`, the table that holds the `RelationEvidence`
envelope the contract returns on a relation proposal, a panel-group revision, and
a mosaic revision.

Makes `PanelGroupRevision.representativeEvidenceId` and
`MosaicRevision.capturedUnionEvidenceId` optional in the contract spec, in the
Rust DTOs, and in the desktop TypeScript mirror.

No migration and no repository or command code changes.

## Contract changes

- `relation_evidence.query`: read-only, takes an `evidenceId`, returns
  `RelationEvidence`. It is the only way a client resolves a revision's evidence
  reference, because a revision without a proposal has no other handle.
- `expectedMatchingSettingsRevision` on `relation_proposal.manual.create`: a
  required `uint64` the caller reads from `MatchingSettings.revision`. The command
  rejects a submission whose value differs from the accepted revision, so
  evidence measured under superseded thresholds is not stored under the current
  ones.
- `relation_evidence.not_found`: returned for an unknown evidence ID.
- `PanelGroupRevision.representativeEvidenceId` and
  `MosaicRevision.capturedUnionEvidenceId` become optional
  (`crates/contracts/core/src/sessions/heterogeneity/relations.rs:135`, `:163`,
  `apps/desktop/src/features/sessions/groupsTypes.ts:103`, `:125`).
  `crates/app/inbox/src/session_materialization/apply.rs:278` writes a singleton
  panel revision on every ingestion with no proposal and no measured geometry, so
  a required field is unsatisfiable for rows that exist today.
- `SafeText` replaces `string` on the missing-evidence codes and on
  `ThresholdMeasurement.key` and `unit`, matching the landed DTOs at
  `crates/contracts/core/src/sessions/heterogeneity/relations.rs:282`, `:284`,
  `:305`, `:363`.

## Data model

- `relation_evidence`, keyed by its own id, with typed columns for the
  target-compatibility, parity, acquisition-geometry, and equipment verdicts;
  footprint coverage and centre separation in parts per million; residual sky
  rotation in microdegrees; a settings-revision reference; and an input digest.
- Three typed child tables for the bounded contract lists, so the missing codes,
  allowed rotation ranges, and threshold snapshot stay queryable instead of
  living in JSON. Each child batch commits in the envelope's transaction.
- `relation_proposal.evidence_id`,
  `panel_group_revision.representative_evidence_id`, and
  `mosaic_revision.captured_union_evidence_id` as nullable unique references.
- A statement that
  `crates/persistence/core/migrations/0001_initial_schema.sql` creates none of
  these tables or columns, which separates them from the landed tables the rest
  of the document describes.

## Two revision concepts stay separate

`matching_settings_revision`
(`crates/persistence/core/migrations/0001_initial_schema.sql:1256`) holds the
thresholds a verdict is measured against, and its `revision_number` is the
contract `matchingSettingsRevision` on `RelationProposal`,
`PanelGroupRevision`, and `MosaicRevision`
(`crates/contracts/core/src/sessions/heterogeneity/relations.rs:136`, `:164`,
`:379`).

`spec062_config_revision`
(`crates/persistence/core/migrations/0001_initial_schema.sql:2849`) is the
feature-wide configuration revision that every subject table carries as
`config_revision_row_id` and the fourth column of the proposal deduplication
key.

The envelope references the settings revision alone, because `RelationEvidence`
does not project a configuration field.

## Why the three database references are nullable

`accept_proposal` updates proposal state and visibility history
(`crates/persistence/topology/src/repositories/proposals.rs:313`), and the
revision inserts bind no evidence column
(`crates/persistence/topology/src/repositories/panels.rs:530`, `:628`,
`crates/persistence/topology/src/repositories/mosaics.rs:490`). A `NOT NULL`
column would lack a writer on the accept path, and every acceptance would fail.
SQLite also rejects adding a `NOT NULL` foreign-key column without a default to
the already-populated tables in the append-only migration chain.

`RelationProposal.evidence` stays required. `insert_proposal`
(`crates/persistence/topology/src/repositories/proposals.rs:249`) has no caller
outside `crates/persistence/topology/src/test_support.rs`, so no stored proposal
predates the column, and the writer that lands proposal generation commits the
envelope in the proposal's transaction.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings`: exit 0.
- `cargo nextest run --workspace --no-fail-fast`: exit 0.
- `cd apps/desktop && pnpm lint`: exit 0.
- `git diff origin/main -- crates/persistence/core/migrations` is empty, so the
  migration chain is byte-identical to `main`.

Tracks-Bead: astro-plan-my0jt
Merge-Bead: astro-plan-k6uo3


srobroek:
All four findings were valid. Fixed in 0043e7e5.

**Persist the proposal-to-evidence relationship** — correct, `relation_proposal`
had `evidence_digest` and no foreign key (`0001_initial_schema.sql`, the
`relation_proposal` DDL, confirms no `evidence_row_id`). Added `evidence_id
NOT NULL UNIQUE REFERENCES relation_evidence(id)` and defined `evidence_digest`
as a restatement of the referenced row's `input_digest`, so the existing
`UNIQUE(kind, basis_digest, evidence_digest, config_version_id)` key stays the
idempotency guard and a disagreeing digest is a rejected write rather than a
holdable state.

**Bind revision evidence to its revision inputs** — correct. The three
referencing columns are now each `UNIQUE`, so one envelope serves one subject
and no row is shared. The row itself records `subject_kind` (`proposal`,
`panel_group_representative`, `mosaic_captured_union`) and a `subject_digest`
over the ordered subject identities; a reader that recomputes a different digest
treats the evidence as absent.

**Make the child-list bounds explicit** — correct. Each child table now carries
`CHECK (ordinal BETWEEN 0 AND 99)` (missing codes, measurements) or `0 AND 15`
(rotation intervals), which with per-evidence ordinal uniqueness caps the row
count at the contract's `BoundedList` bound without a counting trigger. On the
vocabularies: `code`, `measurement_key`, and `unit` stay open `TEXT` because the
contract types them `SafeText`, matching how `proposal_measurement` already
treats them; `comparison` and `outcome` are closed enums and carry `CHECK`
constraints. The doc now says which side owns each.

**Resolve the Tier 2 and required-foreign-key conflict** — this was an error in
the original text, not an ambiguity. Three `NOT NULL` foreign keys point at the
envelope, so it must commit in the same transaction as its subject; batching it
is unavailable at any tier. Rewritten to separate write order from durability:
referential integrity fixes the order, the content stays re-derivable from the
footprints plus `config_version_id`, and only the measurement, missing-code, and
allowed-rotation children are Tier 2. An envelope whose children have not landed
reads as an empty `thresholdSnapshot`, reported as unmeasured rather than as
passing. Tier 1 stays the user decision on `relation_proposal`.

srobroek:
@coderabbitai full review

srobroek:
@coderabbitai full review

srobroek:
Both bots reviewed `0a5064d9`, not head. Every finding was re-checked against head rather than taken on the review status.

Already closed at `0043e7e5`:

- proposal-to-evidence foreign key — `relation_proposal.evidence_id` is `NOT NULL UNIQUE REFERENCES relation_evidence(id)`, with `evidence_digest` defined as the referenced row's `input_digest`.
- binding evidence to its revision inputs — `subject_kind` and `subject_digest`, with recompute-and-reject read semantics.
- child-list bounds — `CHECK (ordinal BETWEEN 0 AND 99)` over a dense ordinal caps the list at 100 without a counting trigger, and the text states which rules the database owns and which the validators own.
- the Tier 2 versus `NOT NULL` conflict — the envelope commits in the subject's transaction; batching it behind its subject is unavailable at any tier. Only the measurement children are Tier 2.
- the 100-row cardinality concern for migration- and maintenance-written evidence — the same ordinal `CHECK` binds every writer, and the preamble routes migrations and import tools through the same repository transaction boundary.

Fixed in `8d7243bc`:

- `input_digest` now covers the resolved target identities and equipment-resolution revisions, not only footprints and the settings revision. Without them an equipment-resolution change left the digest identical while the stored `equipment` verdict had to change.
- the exclusive-ownership claim was wrong as written. Three per-table `UNIQUE` columns do not stop cross-table reuse; the rule is a repository precommit query and the text now says so. That query also requires the subject's `config_version_id` to equal the envelope's.
- the threshold mapping is now stated. `threshold_min`/`threshold_max` stays: `proposal_measurement`, which this mirrors, admits `inside` and `warn`, neither present in `ThresholdMeasurement`, and the pair exists to hold `inside`. Collapsing to one column would have dropped it. `lt`/`lte` read `threshold_max`, `gt`/`gte` read `threshold_min`, `eq` requires equality, `inside` projects as a missing-evidence code, and `warn` projects as `fail`.

srobroek:
@coderabbitai review

srobroek:
All seven findings addressed in `84d585b25`.

- **P1 zero expected children** — correct, and a defect I introduced. Predicate is now `COUNT(*) = COALESCE(MAX(ordinal) + 1, 0) = expected_count`.
- **Append-only vs. reconciliation DELETE** — correct on convention. `0001_initial_schema.sql:3488-3514` has `BEFORE DELETE` `RAISE(ABORT)` triggers for `audit_event`, head retirement, and visibility closure but not for these child tables, so nothing literally blocked the delete; the objection stands against the doc's own framing regardless. The delete is gone and an overfull envelope is now an ambiguous outcome routed to the user.
- **Historical inputs for reconciliation** — accepted, resolved the other way round. Step 2 compares the recomputed `input_digest` against the stored one: a match means the heads have not moved and the pass resumes; a difference means the envelope is stale and goes to the user. No FK set and no repair work item, which would be userspace write-ahead journaling that the project governance document's unclean-shutdown constraint forbids.
- **Repair prompt without a proposal** — correct. Routes through `subject_kind` / `subject_digest`.
- **Rotation endpoint inclusivity** — correct. Stored intervals are closed at both ends, the projection returns both flags as `true`, and a validator rejects a `manual.create` range asking for an open endpoint.
- **Evidence error code** — added with its `evidenceId` detail.
- **Child text fields not `SafeText`** — correct about the markdown, and the drift runs the other way: `crates/contracts/core/src/sessions/heterogeneity/relations.rs:275` already types `ThresholdMeasurement.key` and `.unit` as `SafeText`, and `:299` types `missing_evidence_codes` as `BoundedList<SafeText, 100>`. The contract markdown said `string`. The markdown now matches the Rust surface, so the data-model rationale was accurate against code all along.

srobroek:
@coderabbitai review

srobroek:
@coderabbitai review

PR #1658, opened 2026-08-19
fix: developer command-palette entry no longer ships in release builds

srobroek:
## What

The command palette's `/dev/contracts` entry shipped in release bundles. It was
gated on the runtime `devMode` setting, while the route it points at was gated on
the compile-time `VITE_DEV_TOOLS` define, so a release build carried a palette
entry to a route the router had compiled out.

`DEV_PAGES` is now gated on `DEV_TOOLS_ENABLED` at module level, so Rollup drops
it from a release bundle. The constant lives in one place,
`apps/desktop/src/dev/devToolsEnabled.ts`, read by both the router and the
palette.

## Evidence

Built the frontend both ways and grepped the artifact:

| `VITE_DEV_TOOLS` | `/dev/contracts` in `apps/desktop/dist` |
| --- | --- |
| `false` | absent |
| `true` | `dist/assets/main-7Tm_haEl.js` |

Before the fix, the `false` build emitted `route:"/dev/contracts"` into
`dist/assets/main-CLayXJml.js`. The second row is the control: without it the
grep would pass whether or not the fix worked.

`pnpm vitest run src/dev/` — 12 files, 83 tests, all passing. `tsc --noEmit` and
`biome check` clean on the edited files. `actionlint` clean on the workflow.

## Why the tests missed it

`commandPalette.devMode.test.ts` declared local copies of `PAGES`, `DEV_PAGES`,
and the visibility logic and tested those. The copies had drifted as well: they
listed `/review`, `/plans`, and `/audit`, none of which are palette pages. The
tests now import the real exported constants. The developer-build branch takes
its dev pages as a parameter, because `vitest.config.ts` bakes
`VITE_DEV_TOOLS="false"` into the test run and no environment variable can reach
it.

## Release gate

Check 3 inspects the Cargo feature graph, where a Vite define cannot appear, so
the frontend half of the developer-surface claim was unverified: a release built
with `VITE_DEV_TOOLS=true` in the environment shipped both dev routes and every
check still passed. A new hard check builds the bundle the way `tauri build` does
and greps `apps/desktop/dist` for `/dev/contracts` and `/dev/settings`.

Release PR #1393 stays held; this changes the gate, not the release.

Tracks-Bead: astro-plan-9f78
Merge-Bead: astro-plan-a460
Closes-Bead: astro-plan-9f78

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -39,6 +39,7 @@ env:
   POLICY_VERSION_SYNC: hard
   POLICY_BUILD: hard
   POLICY_DEVTOOLS_OFF: hard
+  POLICY_DEVTOOLS_BUNDLE: hard
   POLICY_TESTS: soft
   POLICY_CHANGELOG: soft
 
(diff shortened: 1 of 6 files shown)


srobroek:
@coderabbitai review

srobroek:
@coderabbitai review

srobroek:
@coderabbitai review

srobroek:
@codex review

srobroek:
@coderabbitai review

PR #1660, opened 2026-08-19
fix: settings panes reflect a value changed elsewhere in the app

srobroek:
## Summary

- `settingsQueryOptions` caches each scope with `staleTime: Infinity`, sound only
  while every writer invalidates `queryKeys.settings.scope(scope)`. Nine sites
  called `commands.settingsUpdate` directly and skipped it.
- One of the nine writes a scope a `useQuery` caller already reads, so this is a
  live defect, not the latent one the bead describes: the first-run wizard writes
  `cleanup.defaultProtection` and `Cleanup.tsx` reads `cleanup` through
  `settingsQueryOptions`. Finish the wizard, open Settings > Cleanup: the pane
  shows the pre-wizard protection default for the rest of the session.
- The log panel's follow-tail write is not a second instance. `Advanced.tsx:4-8`
  states `rememberFollowLogs` is surfaced by the log panel rather than duplicated
  in the pane, and `LogPanelContext.tsx:117` seeds itself from a raw
  `commands.settingsGet('advanced')` in a `useEffect`, not from the query cache.
  Nothing that renders that key reads the cache, so invalidating its scope is
  correctness insurance for a future `useQuery` reader, not a fix a user can see.
- `data/settingsWrite.ts` now holds the write plus the invalidation; all nine
  sites call it. `settingsIpc.ts` re-exports it, so its existing callers are
  unchanged.
- It lives in `data/` so the lazy writers there keep reaching it through the
  dynamic import they already use, rather than pulling the whole IPC surface into
  the boot bundle.

## On the bead's two options

The bead recommended (b), a guard test, over (a), one write path, because
`theme`/`locale` write during boot and might have no `QueryClient`. `queryClient`
is a module singleton in `data/queryClient.ts`, not a context value, so that
concern does not hold and both options ship: one write path, and a gate on it.

## Verification

`pnpm vitest run` in `apps/desktop`: 268 files / 2378 tests pass. `tsc --noEmit`
clean; `check-circular.mjs` reports 0 new cycles.

Adding a module that calls `commands.settingsUpdate` makes the guard fail and
names the file. The second assertion covers the other direction: with the
`invalidateQueries` call deleted, a call-site-counting guard alone would still
pass while every write went stale.

`theme.persistence.test.ts` leaked a fire-and-forget write across a test
boundary and only passed because the old write path was slow enough to land
inside its own test. The write is now settled and asserted where it is made.

Tracks-Bead: astro-plan-cia8
Merge-Bead: astro-plan-qsjq
Closes-Bead: astro-plan-cia8

### DIFF
--- a/apps/desktop/src/app/LogPanelContext.tsx
+++ b/apps/desktop/src/app/LogPanelContext.tsx
@@ -23,6 +23,7 @@ import {
 } from 'react';
 import { commands } from '@/bindings/index';
 import { unwrap } from '@/api/ipc';
+import { updateSettings } from '@/data/settingsWrite';
 import type { LogLevel, LogEntrySource } from '@/data/logStore';
 import { startLogSubscription } from '@/data/logSubscription';
 import { createPersistedState } from '@/data/persisted-state';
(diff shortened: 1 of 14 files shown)


srobroek:
@coderabbitai review

srobroek:
@coderabbitai review

srobroek:
@codex review

srobroek:
@coderabbitai review

PR #1661, opened 2026-08-19
Fail CI when a test file is never compiled

srobroek:
## Summary

- `tests/contract/` keeps its test targets at the package root. Cargo
  auto-discovery reaches only `<pkg>/tests/*.rs`, so every file there needs an
  explicit `[[test]]` entry. `spec062_portable_contracts.rs`,
  `catalog_entry_file_test.rs`, and `catalog_manifest_parity_test.rs` lacked one
  and had never compiled.
- `spec062_portable_contracts.rs` asserted
  `residualSkyRotationCapDeg: 90.0` against a validator that fixed the value at
  90.0, while the contract fixes it at 10. Both the validator bug and the test
  that encoded it survived because the file never ran. That file's registration
  and constant fix are on `fix/pssh-matching-settings-validator`.

## The guard

`scripts/check-test-targets-declared.sh` compares root-level `*.rs` files against
the declared `[[test]]` paths in both directions: an undeclared file fails, and so
does a declared path with no file. Wired into `just check` and the Linux CI
ratchet lane.

Both arms verified by hand: it names the still-undeclared file, and it names a
`[[test]]` path pointing at nothing.

## The catalog tests cover a subsystem #250 removed

`catalog_entry_file_test.rs` and `catalog_manifest_parity_test.rs` both import
`targeting_catalogs` and `contracts_core::catalogs`. Spec 035 T034/T035 removed
that surface in #250, and spec 014 carries a wholesale supersession banner. #257
added these tests five hours after the removal, so they have never compiled
against anything. The signed-manifest design they cover was replaced by the
bundled seed index in `crates/targeting/resolver/src/seed.rs`.

Declaring them, as the bead's step 1 asks, would mean restoring a deleted
subsystem to satisfy tests written for it. Deleting them is what the bead's step 1
resolves to once the imports are checked.

Tracks-Bead: astro-plan-cfbe
Merge-Bead: astro-plan-mj4c
Closes-Bead: astro-plan-cfbe


srobroek:
Blocked on #1656, and the failure is the guard doing its job.

`Unit + integration (L1+L2) — ubuntu-latest` fails with:

```
ERROR: tests/contract has test file(s) with no [[test]] entry in Cargo.toml.
       Cargo never builds these, so their assertions never run:
         spec062_portable_contracts.rs
```

That file's `[[test]]` registration, and the `residualSkyRotationCapDeg` constant fix it exposes, are on #1656. Declaring it here instead would either conflict with #1656 or land a registered test asserting the wrong constant.

Merge order is recorded as a beads edge: `astro-plan-mj4c` depends on `astro-plan-prqn`. Rebasing this branch after #1656 lands clears the lane.

srobroek:
@coderabbitai review

srobroek:
@codex review

srobroek:
@coderabbitai review

PR #1662, opened 2026-08-19
fix: perf regression gate no longer fails on a busy CI runner

srobroek:
## Summary

- `calibration_suggest_1k_masters` read `sqlx_stmts=11` against a baseline of 10
  on CI, then 10 on a rerun of the same commit. `scripts/check-perf-baseline.sh`
  hard-fails on any increase, so the gate blocked a PR whose diff did not touch
  Rust, and the obvious reading was that the PR caused it.
- The counter matched `target().starts_with("sqlx")`. sqlx emits one event per
  statement under `sqlx::query`, but `sqlx-core/src/pool/inner.rs:305` also emits
  under `sqlx::pool::acquire`, and only when a connection acquire exceeds
  `acquire_slow_threshold`. It never fires on an idle machine and does fire on a
  contended runner, which is why the same commit produced both counts.
- The match is now exact on `sqlx::query`. Both the normal and slow-statement
  branches of `sqlx-core/src/logger.rs` use that target, so a slow statement is
  still counted once.
- `crates/app/core/tests/artifact_reconcile_batching.rs` uses the same predicate
  and is fixed with it.

## Verification

Statement counts are unchanged on an idle machine: 0, 660, 9, 40113, 10 against
baselines 0, 660, 9, 40116, 10. `bash scripts/check-perf-baseline.sh` passes on
every scenario, so the baseline does not need regenerating.

The new unit test in `support.rs` emits `sqlx::query` at debug and at warn, plus
`sqlx::pool::acquire`, an `sqlx_sqlite::*` target, and an unrelated target, and
expects 2. Restoring the prefix predicate makes it fail with 4, so it constrains
the predicate rather than merely exercising it.

`plan_apply_progress` keeps its `stmts_timing_dependent` opt-out. Its variance
comes from a time-bounded flush that genuinely issues more statements on a slower
runner, which is a different problem from this one.

Tracks-Bead: astro-plan-hgh6
Merge-Bead: astro-plan-h2ns
Closes-Bead: astro-plan-hgh6

### DIFF
--- a/crates/app/core/tests/artifact_reconcile_batching.rs
+++ b/crates/app/core/tests/artifact_reconcile_batching.rs
@@ -30,9 +30,14 @@ use tracing_subscriber::layer::SubscriberExt as _;
 const N: usize = 60;
 const PROJECT_ID: &str = "proj-batch";
 
-/// Counts tracing events emitted under the `sqlx` target (one per statement
-/// execution). `max_level_hint` must claim DEBUG or the registry drops sqlx
-/// events before this layer sees them.
+/// Counts tracing events emitted under the `sqlx::query` target (one per
+/// statement execution). `max_level_hint` must claim DEBUG or the registry drops
+/// sqlx events before this layer sees them.
+///
+/// The target match is exact. A `starts_with("sqlx")` prefix also catches
+/// `sqlx::pool::acquire`, which fires only when a connection acquire exceeds
+/// `acquire_slow_threshold` and so adds a phantom statement under load
+/// (astro-plan-hgh6).
 struct SqlxCounterLayer(Arc<AtomicU64>);
 
 impl<S: tracing::Subscriber> tracing_subscriber::Layer<S> for SqlxCounterLayer {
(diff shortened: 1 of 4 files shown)


srobroek:
@coderabbitai review

srobroek:
Addressed in `7ba90ba3e`. Correct, and the gap was wider than the constructors named: the stem alternation also missed `query_scalar_with`, `query_file_as!`, and `query_file_scalar!`, and `.fetch(` was absent from the consumer list. The stems now admit `[[:alnum:]_]*` before the call/bang/turbofish suffix, `.fetch(` is covered, and `"sqlx::query"` stays excluded because a quote follows it. Six self-test cases added. The scan still reports 0 sites outside `crates/persistence/`, so the broader pattern moved no baseline.

srobroek:
Both findings addressed in 874933d72 with one change, because they are the same gap seen from either side.

The awk scanner now removes block-comment spans before matching, rather than skipping lines that contain one. `/* sqlx::query("SELECT 1") */` counts zero, and `sqlx::query /* runtime */ ("SELECT 1")` counts one: `PATTERN` allows only whitespace between the constructor and its `(`, and deleting the comment leaves exactly that.

Nesting is not handled and the comment says so. Rust nests block comments; awk has no depth counter here, so a nested comment ends the span early and leaves the tail to be matched. That over-counts, which fails a CI run rather than admitting a leak.

Four self-test cases added (inline comment, multi-line span, comment between constructor and call, comment before real code on the same line). `--self-test` passes all 17; the full scan still reports 0 production query sites outside `crates/persistence/`.

PR #1663, opened 2026-08-19
fix(ci): sync the merge scanner's required-check list to branch protection

srobroek:
## Summary

- `scripts/pv-watch-queue.sh` decided PR mergeability against a `$required` array
  that named `Real-UI journeys (L3)` on two runners. Those jobs never run on
  `pull_request` under the `e2e.yml` two-tier strategy, so every PR read ABSENT on
  two contexts. The array also omitted `Real-UI smoke (L3) - ubuntu-latest` and
  `Supply chain (cargo-deny)`, which branch protection does require.
- The list now matches `gh api repos/platevault/platevault/branches/main/protection`
  exactly: `Detect changes (CI)`, `Unit + integration (L1+L2)` on ubuntu and
  windows, `UI mock-mode (Playwright)`, `Real-UI smoke (L3) - ubuntu-latest`,
  `Supply chain (cargo-deny)`.
- `scripts/check-required-contexts.sh` reads live branch protection and diffs it
  against the embedded array, reporting required-but-missing and
  named-but-not-required separately.

## The fixture duplicated the value under test

`scripts/tests/pv-watch-queue.bats:93` embedded a copy of the same wrong list, so
all 14 tests passed while the scanner misclassified every PR. The suite asserted
that the script agreed with a duplicate of its own mistake. The fixture is
corrected here, and the drift guard is what makes the agreement meaningful, since
it compares against the API rather than against another copy.

## Verification

`bash scripts/check-required-contexts.sh` reports 6 matching contexts.
`bats scripts/tests/pv-watch-queue.bats` passes 14 tests.

The guard is mutation-verified: replacing `Supply chain (cargo-deny)` with a
context that protection does not require exits 1 and names both directions, so it
catches a stale extra entry and not only a missing one. The extra entry is the half
that caused this bug, and a guard checking only for omissions would have stayed
green.

Tracks-Bead: astro-plan-lgyr
Merge-Bead: astro-plan-024h
Closes-Bead: astro-plan-lgyr


srobroek:
@coderabbitai review

srobroek:
@coderabbitai review

srobroek:
Four findings addressed in `560aedf9a`.

- **Not wired into anything** — correct. `just lint` now runs it. The policy comparison is offline; the live comparison after it skips with a NOTE without network or token.
- **Compare against the checked-in policy** — correct, and this was the important one. The gating comparison is now the embedded array against `scripts/branch-protection-main.json`, so a synchronised PR passes. Live protection is compared against that same JSON separately, with `branch-protection-main.sh` named as the remediation for an out-of-band change.
- **Empty context list vs API failure** — correct. The API's exit status is kept separately and a successful empty response is compared rather than skipped.
- **Hard-coded readiness count** — correct, and stronger than reported: `$checks` is built by mapping over `$required`, so `($checks | length) == 6` could only ever be false when the array size changed, which is exactly when it would hold every green PR non-READY. The clause is dropped and the progress column reads `$required | length`.

The 14-case bats suite passes.

srobroek:
All three findings addressed in cd90fce27.

**Missing jq** — now a hard failure. The gating comparison is offline and claims to always run, so exiting 0 before it reported a pass that measured nothing. `pv-watch-queue.sh` needs jq to run at all, so any environment that can use the watcher can check it.

**Live comparison** — now `WARN` + `exit 0`. You are right that it would make `just lint` red for exactly the synchronised PR the gating comparison was rewritten to admit, since protection is applied at merge and not at lint.

**Remediation command** — now `branch-protection-main.sh apply`. Confirmed at `scripts/branch-protection-main.sh:104`: `case "${1:-show}" in`, so the argument-less form only read protection.

Verified: the script prints both OK lines and exits 0 against live protection, the `gh`-missing path prints its NOTE after the gating comparison passes, and the 14-case bats suite still passes.

srobroek:
@coderabbitai review

Issue #831, opened 2026-07-14
Settings → Audit Log: entity rows have zero cross-link affordance (journey spec requires row → entity navigation)


srobroek:
## Journey / step
Journey 13 (Audit & activity investigation), step 3 / Touch & validate: "Filtering by entity/date narrows the trail; an audit row links back to the entity and, for plan events, to the plan's item list."

## Repro
1. Navigate to Settings → Audit Log (`#/settings/audit-log`).
2. Any row, e.g. `project · bf6f5e26-7edd-480b-b7eb-e7e7c02e2a71` / event "Archive" / outcome "refused".
3. Click the entity cell (`.alm-audit-log__entity`) or the row.

## Expected
Per the journey spec, an audit row should link back to its entity (and for plan events, to the plan's item list) — landing on the entity's page with it selected.

## Observed
- `getComputedStyle(row).cursor` = `"auto"` (no pointer affordance), `row.onclick` = `false` for all rows checked.
- Clicking the entity cell or the row does nothing — `window.location.hash` stays `#/settings/audit-log`.
- This is distinct from #626 (LogPanel/Activity-panel cross-links routing to nonexistent routes for 3/5 entity types) — here the durable Audit Log table has **no** cross-link behavior at all, for any entity type.

## Classification
UI — missing interaction affordance / routing, not a data problem (the entity_type/entity_id are present and correct in the underlying row).

## Evidence
Screenshot: shots/j13-13-search-entity.png (entity cell shows raw UUID, no visual affordance). DOM query above run via webview_execute_js against the live Settings → Audit Log page.

Issue #626, opened 2026-07-11
LogPanel entity cross-links are broken for 3 of 5 entity types


srobroek:
## Problem
Clicking an entity link in the Log panel navigates to a route that doesn't exist for 3 of the 5 linkable entity types, silently redirecting to Sessions instead of the intended destination.

## Evidence
`LogPanel.tsx:62-77` — `/plans/:id`, `/settings?tab=catalogs`, and the `/audit` fallback all point at nonexistent routes.

## Recommendation
Point the catalog link at the actual settings/catalogs pane path, point the audit link at `/settings/audit`, and drop the plan link until plans are deep-linkable (or add real deep-linking support for plans).

**Severity:** P2 · **Area:** Log panel

_Source: automated design review 2026-07-11 (docs/development/design-review-2026-07-11.md, PR #598)._

srobroek:
Per-kind fresh evidence (2026-07-12 sweep):
- **project** rows navigate to bare `#/projects` — nothing selected.
- **plan** rows build `/plans/<id>` (LogPanel.tsx:65), a route that doesn't exist → not-found fallback → user lands on Sessions with no context.
- **target** rows render with no link affordance at all (`--link` class/role absent; entityId missing on the emitted events).
Journey 13 requires each entity cross-link to land on an existing route with the entity selected.

PR #1665, opened 2026-08-20
refactor: share the note byte counter, entity-path builder, and dismiss hook across UI call sites

srobroek:
## Summary

- Each autosave note editor derived its byte count, over-limit flag, and near-limit warning threshold from its own copy. `noteByteStatus()` in `apps/desktop/src/lib/notes.ts` now derives all three from one 90% threshold, and `NoteByteCounter` renders it.
- The log panel and the audit log each built an entity link path with their own resolver. `entityPath(entityType, entityId, {fallback})` in `apps/desktop/src/lib/entity-path.ts` replaces both; the fallback that was their only divergence is now an explicit option, and `plan` resolves to null ahead of the fallback so no caller links a plan by accident (refs #831, #626).
- The checklist popover and the root card each carried a pointerdown-outside plus Escape listener. `useDismiss(refs, onDismiss, enabled)` in `apps/desktop/src/hooks/useDismiss.ts` replaces both. `refs` is a list because a portalled panel is not a DOM descendant of its anchor.
- The note render bodies stay separate. The project section toggles between view and edit, has an explicit Save, fetches its own note, and maps three errors; the session section is always editable and flushes on unmount. They do not share an interaction model.

## Test plan

`pnpm test` (270 files, 2408 tests) and `pnpm typecheck` in `apps/desktop`; `node scripts/check-eslint-baseline.mjs` at the repo root.

`apps/desktop/src/lib/entity-path.test.ts` and `apps/desktop/src/app/log-panel-model.test.ts` cover the extracted builder. Their cases derive from the two switches this PR replaced, read at base commit 5f15a992b, so they assert the pre-extraction behaviour rather than restating the new helper: `plan` resolves ahead of the fallback, the fallback is not consulted for a kind that has its own detail page, and ids are interpolated without percent-encoding.

`origin/main` is merged in. The merge conflicted only on two line numbers in `scripts/eslint-alm-baseline.txt`, which is regenerated against the merged tree and holds the same 87-entry suppression set.

## Spec Context

Tracks-Bead: astro-plan-kyo7
Merge-Bead: astro-plan-q3en


srobroek:
Reviewed head c8582761c against merge-base 5f15a992b. Verified myself: `pnpm test` in `apps/desktop` gave 267 files / 2376 tests passed, exit 0; `pnpm lint` exit 0. Every call-site conversion is behaviour-preserving:

- `buildEntityPath` (apps/desktop/src/app/log-panel-model.ts:103) reproduces the old switch exactly, `catalog` and default fallback included; `AuditLog` (apps/desktop/src/features/settings/AuditLog.tsx:425) had `default: null` before and gets no fallback now.
- `NoteByteCounter` (apps/desktop/src/components/NoteByteCounter.tsx) keeps both test ids, both class ternaries, and the same `projects_notes_bytes_unit` key that both editors already used.
- `useDismiss` (apps/desktop/src/hooks/useDismiss.ts) keeps `pointerdown`+`keydown` on `document`, the `enabled`-only registration, and the popover Escape `stopPropagation` / focus-restore, now in the callback.
- `scripts/eslint-alm-baseline.txt`: exactly four line-number shifts, identical suppression set, as the body claims.

1. `apps/desktop/src/lib/entity-path.ts:39` - the module holding the only new branching in this PR (the `fallback` option, and `plan` ordered ahead of it) has no test anywhere; `rg -l 'entityPath|buildEntityPath' --glob '*.test.*'` returns nothing. A later edit moving `case 'plan'` after `default` would silently make the LogPanel link plans to `/settings/audit?...`, the exact regression the docstring warns about. Add a unit test for `entityPath`: the three mapped types, `plan` to `null` with a non-null fallback supplied, and an unmapped type to `null` with no fallback.

PR #1666, opened 2026-08-20
test: pin observing-site latitude, longitude, and horizon bounds

srobroek:
## Summary

- The observing-site ranges were already enforced at runtime by `check_observer_sites`, but only one out-of-range case was tested. Tests now accept a value exactly on each inclusive bound and reject one 0.1 past it, for latitude, longitude, and minimum horizon altitude.
- A second test pins what happens on read: `get_settings` repairs an invalid value by deleting the key, so one out-of-range site discards every sibling site. The behaviour is now asserted rather than incidental (see astro-plan-mq5m for the fix).
- The bounds for latitude, longitude, and horizon altitude were hardcoded in four TypeScript locations. `observer-site.ts` now exports them once and the settings form, the setup step, and the location picker consume them, giving the pending Rust-to-TypeScript constant export one target.
- Validation stays in the settings-apply path rather than moving to a domain constructor: `ObserverSite` is built by `serde_json::from_value` in the descriptor apply closure, so a constructor invariant would be bypassed on every real load. Out-of-range values are rejected, not clamped, so the stored site equals what the user entered.

## Test plan

`cargo test -p app_core_settings` and `pnpm test` in `apps/desktop`.

## Spec Context

Tracks-Bead: astro-plan-zvxw
Merge-Bead: astro-plan-otv5

The moon-phase half of astro-plan-zvxw is not covered here, so the bead stays open.


srobroek:
@codex review

PR #1667, opened 2026-08-20
feat: notify when a plan apply, workflow run, or target resolution finishes

srobroek:
## Summary

- An OS notification is now raised at three completions:
  - a plan apply reaching a terminal state, on the bus topic `plan.applying.completed`,
  - a tool workflow run attributed as complete, on the bus topic `workflow.run_completed`,
  - an ingest-resolution drain pass that resolved at least one queued target, called from the 30-second drain scheduler rather than a bus topic.
- `tauri-plugin-notification` was declared in the workspace manifest but never added to the desktop crate, so no notification path existed.
- Notifications are suppressed while the main window is focused, because a toast plus an OS popup for the same event is worse than either alone. Unknown focus notifies, since a missed notification is the worse failure.
- Bus payloads are deserialised into the producer's own event struct (`PlanApplyingCompleted`, `WorkflowRunCompleted`), so a payload that does not match raises nothing. `plan.applying.completed` carries one degraded publisher that emits only `planId` and `cancelledCount` when `list_pending_items` fails twice (`crates/app/core/src/plan_apply/terminal.rs:388`). That row would otherwise be announced as a successful apply with zeroed counts, and announced again when the real completion follows.
- Rejected: every drain tick, because the 30-second backstop is usually a no-op; scan and classify passes, because the user is already watching those surfaces; a replay of completions dropped to broadcast lag, because that outcome is long since visible in the UI; and `manifest.write.success`, because the manifest subscriber replays historical `workflow.run_completed` rows from the durable events table after a lag and republishes a success event per replay (`crates/app/projects/src/project_manifests.rs:416`).
- No runtime permission request and no capability grant. The plugin desktop backend reports `Granted` unconditionally, the notification is raised from Rust rather than the webview so no ACL applies, and an OS-level denial or a missing Linux notification daemon returns an `Err` from `show()` that each call site logs at debug and ignores.
- `drain_and_backfill_once` now returns its `DrainSummary` so the caller can gate on `resolved > 0`. A failed pass returns a zeroed summary, making failure and an empty queue equivalent for that gate.

## Test plan

`apps/desktop/src-tauri/src/bootstrap/notify.rs` unit-tests the focus gate, both payload summaries, and topic dispatch through `notification_for`: the subscribed topics, an unsubscribed topic, and an undeserialisable payload on a subscribed topic.

On-device notification delivery is untested (T051). Verification counts are recorded on `astro-plan-i1ks`.

## Spec Context

Tracks-Bead: astro-plan-i1ks
Merge-Bead: astro-plan-x4rm

Spec 051 US8 T045-T050. T051 is manual on-device verification and stays open, so the bead does too.


srobroek:
@codex review

srobroek:
Both findings confirmed against source and fixed. The commit is blocked on a hung 1Password `op-ssh-sign` approval prompt (two processes stuck 58m+), so the fix is written and tested locally but not yet pushed.

**Finding 1 — replayed workflow runs (notify.rs:89). Confirmed.** `project_manifests.rs:458` handles `Lagged` by calling `replay_workflow_events_since`, which pages `workflow.run_completed` rows out of the durable events table from a cursor that starts at zero, writes a manifest for each, and each write publishes a fresh `manifest.write.success` (`project_manifests.rs:261-272`). One lag therefore produced a burst of notifications for runs the user finished long ago, which is exactly what this notifier's own `Lagged`-drops comment claims not to do.

Fix: notify off `TOPIC_WORKFLOW_RUN_COMPLETED` instead. The attribution pass publishes it once, live (`crates/app/lifecycle/src/artifact/launches.rs:102-114`), and the replay path reads those rows without republishing them, so replays are invisible to the notifier by construction rather than by a filter that could drift. `TOPIC_MANIFEST_WRITE_SUCCESS`, `MANIFEST_REASON_WORKFLOW_RUN` and `manifest_reason` are gone. The body now names the tool and the artifact count.

**Finding 2 — degraded completion payload (notify.rs:111). Confirmed.** `terminal.rs:388-394` writes `json!({"planId": plan_id, "cancelledCount": cancelled_count})` to `TOPIC_PLAN_APPLYING_COMPLETED` when `list_pending_items` fails twice, and `handle_cancelled` publishes the real typed completion afterwards (`terminal.rs:234-249`). The defaulting closure rendered the first as `done: 0 applied, 0 failed, 0 skipped`.

Fix: `plan_apply_body` and `workflow_run_body` both deserialize the typed payload (`PlanApplyingCompleted` / `WorkflowRunCompleted`) and return `None` on mismatch, so the caller notifies nothing. The two tests that asserted the defaulting behaviour are replaced by four: full-payload summaries for each topic, the degraded bulk-cancel shape notifying nothing, and an untyped workflow payload notifying nothing. `cargo test -p desktop_shell --lib notify`: 8 passed.

srobroek (commented):
Review dimension: verification (round 1). Head 8401bca.

Verified: the notification payload keys and reason values match their producers (`crates/audit-types/src/event_bus.rs:386-397` camelCase; `crates/app/projects/src/project_manifests.rs:263-270`; `crates/project/structure/src/manifest.rs:42`), the workflow-run write path reaches `write` with `DtoManifestReason::WorkflowRun` (`crates/app/projects/src/project_manifests.rs:497`), and `cargo test -p desktop_shell --lib -- bootstrap::notify` is 7 passed / 0 failed.

Two verification gaps:

1. `apps/desktop/src-tauri/src/bootstrap/notify.rs:22,26` - the topic string and the `workflow_run` reason are re-declared as local literals that no test or compiler check ties to their producers (`crates/app/projects/src/project_manifests.rs:263`, `crates/project/structure/src/manifest.rs:42`), and the new tests assert only against JSON authored in the same file, so a producer-side rename silently disables both notifications with a green suite. Export the topic as a `pub const` from `project_manifests` (reachable as `app_core::projects::project_manifests`, no new dependency) and derive the reason from `ManifestReason::WorkflowRun.as_db_str()`, or add a test that publishes through the real producer.

2. `apps/desktop/src-tauri/src/bootstrap/notify.rs:78-106` - `spawn_completion_notifier` is the only untested unit in the change: the topic dispatch, the workflow-run filter, and the lag/closed arms have no coverage, and the concrete `AppHandle` (unlike the generic helpers at :38, :48, :61) blocks the `tauri::test::MockRuntime` pattern already used in `apps/desktop/src-tauri/tests/onboarding_subscriber_integration.rs:32`. Extract the decision as a pure `fn notification_for(topic: &str, payload: &Value) -> Option<(&'static str, String)>` and unit test the three branches (plan-apply, workflow-run manifest, non-workflow-run manifest); that also gives item 1 a place to assert.

Not blocking, recorded: `crates/app/targets/src/ingest_resolution.rs:379,387` add a zeroed-summary failure contract with no test, but `resolve_pending`'s counts are covered at :573-616 and the wrapper needs a live resolver, so this stays untested by design.

srobroek (commented):
REVIEW round=2 dimension=rework-verification+body-accuracy verdict=changes head=8249f203 (request-changes refused: own PR)

Round-1 FIX 1 and FIX 2 are resolved. Topic dispatch goes through `audit::event_bus::TOPIC_PLAN_APPLYING_COMPLETED`/`TOPIC_WORKFLOW_RUN_COMPLETED` and deserialises into the producers' own structs; both publishers construct those structs (crates/app/core/src/plan_apply/terminal.rs:137,236; crates/app/lifecycle/src/artifact/launches.rs:104), so a producer rename breaks the build or the tests. `notification_for` is pure and covered. `cargo test -p desktop_shell --lib -- bootstrap::notify` = 12 passed. Body pointers terminal.rs:388 and project_manifests.rs:416 check out; the degraded-payload rejection holds because `PlanApplyingCompleted` has no optional fields (crates/audit-types/src/event_bus.rs:388-397).

1. apps/desktop/src-tauri/src/lib.rs:684-685 - the registration comment still reads 'plan-apply and workflow-run **manifest** completions', the exact claim removed from the PR body as false at this head; `spawn_completion_notifier` deliberately ignores `manifest.write.success` (apps/desktop/src-tauri/src/bootstrap/notify.rs:73-79) - drop 'manifest' and name the two topics it subscribes to.

PR #1668, opened 2026-08-20
refactor: move the targets and projects tables onto their generated stylesheets

srobroek:
## Summary

- `targets-table.css.ts` and `projects-table.css.ts` were authored earlier but never imported, so the hand-written stylesheets stayed the only owners of the table rules. `TargetsTable`, `TargetsTableColumns`, `TargetsPage`, `FilterBadges`, and `ProjectsTable` now consume the modules, and the superseded stylesheet blocks are deleted so each rule has one owner.
- Defects fixed in the unadopted modules before adoption:
  - The containing-block rule keyed `.pv-listpage__main`, a class `ListPageLayout` stopped emitting when it moved to a hashed class. Unmatched, the `position: absolute` table wrapper would escape to the viewport and overlay the page top bar. It is now keyed on `[data-testid='listpage-main']`, matching the rule it replaces.
  - `filterBadgeVariants` modelled five modifiers as one `styleVariants` set, but `FilterBadges` composes tier and viability together. Split into `filterBadge` plus `filterBadgeTier` plus `viable`/`notViable`/`unknown`.
  - `tableBody` and `spacer` were missing, so the windowed-body positioning and the zero-padding spacer rows had no replacement.
- `.pv-targets-my-empty` and its `__icon` are dropped rather than migrated: no component or test references them. Built CSS goes from 1945 to 1943 rules, which is exactly those two; each other migrated selector is present under its hashed name.
- Unit tests that asserted migrated class names by value are repointed. The imaging-time glyph queries use the `data-testid` the end-to-end spec already relies on, and the tier and selected-row assertions import the vanilla-extract export.

## Test plan

`pnpm test` and `pnpm lint` in `apps/desktop`; built-CSS rule count compared before and after. `scripts/eslint-alm-baseline.txt` regenerated.

## Spec Context

Tracks-Bead: astro-plan-j3by
Merge-Bead: astro-plan-5a8m

Two of the five feature tables in astro-plan-j3by are migrated here. SessionsTable, MastersTable, and ArchiveTable need the shared dense-table module tracked by astro-plan-etzl, so the bead stays open.


srobroek:
@codex review

PR #1669, opened 2026-08-20
feat: reopen an approved plan for further edits

srobroek:
## Summary

- An approved plan can now be returned to editing. There was no command and no affordance in the plan review overlay, so a reviewed plan that needed a change was stuck.
- `plans.reopen` transitions a plan back from the two states that have an edge to editing in the domain transition table: `approved` and `ready_for_review`. A request from any state without that edge is refused with `plan.invalid_state`. An `applying` or `applied` plan has filesystem actions in flight or done, so its reviewed intent can no longer be withdrawn (constitution II).
- The same `UPDATE` that changes the state nulls `approval_token` and `approved_at`, and its `WHERE` clause carries the expected state rather than relying on a preceding read. `verify_approval_token` treats a null stored token as `plan.approval.stale`, so a token issued before the reopen cannot authorise an apply of the since-edited plan.
- The writes a reopen races are guarded in the reverse direction too. `set_approved` predicates on `ready_for_review`, so an approval that read the state before a concurrent reopen loses instead of restoring `approved` with a live token. `cas_approved_to_applying` binds `approval_token` into its predicate, so an apply authorised before a reopen-edit-reapprove cycle is refused with `plan.approval.stale` instead of applying the pre-edit item set. A lost CAS on either path reports `plan.invalid_state`, not `plan.not_found`.
- Reopening reverses a user decision that is not re-derivable from the filesystem, so it emits `plan.reopened` recording the prior state and the actor (constitution V Tier 1) rather than only flipping state.
- The target state is already in the `plans.state` check constraint and the transition is already in the allowed set, so the change ships without a migration.

## Test plan

`cargo test -p app_core -p persistence_plans -p contracts_core`: 697 passed, 0 failed. `pnpm test` in `apps/desktop`: 2388 passed across 268 files. `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`, `pnpm lint`, and `pnpm typecheck` all exit 0.

Removing `approved_at = NULL, approval_token = NULL` from the reopen `UPDATE` fails `reopen_plan_invalidates_the_approval_token_for_apply` and `reopen_plan_returns_approved_plan_to_draft`. The token test reaches `approved` through a bare state write, because re-approval mints a fresh token that hides the cleared column.

## Spec Context

Spec 017 T023.

Tracks-Bead: astro-plan-yq33
Closes-Bead: astro-plan-yq33
Merge-Bead: astro-plan-nx7d


srobroek:
@codex review

srobroek:
Findings 3, 4 and 5 are implemented locally and verified (619/619 tests in `persistence_plans` + `app_core`, clippy `-D warnings` clean). The commit is queued behind a 1Password SSH-signing prompt on this machine and will push once that clears.

**Finding 5 — approval could undo a concurrent reopen.** `set_approved` now carries `AND state = 'ready_for_review'` in its `WHERE` clause, so an approval that read the old state loses instead of overwriting the reopen. No fixture needed changing: `setup_with_approved_plan` already promotes to `ready_for_review` first (`plan_apply.rs:1056`).

**Finding 4 — apply could run under a revoked token.** `approval_token` joined the `cas_approved_to_applying` predicate, closing the window between the caller's pre-flight token check and the CAS in which a plan could be reopened, edited and re-approved. A revoked token now returns a new `DbError::ApprovalStale`, which maps to `ErrorCode::PlanApprovalStale` — the same contract code `verify_approval_token` already returns for the pre-flight mismatch, so one cause has one failure code.

**Finding 3 — a lost CAS reported as a database fault.** New `state_cas_error` helper reads the row back and returns `NotFound` only when the plan is genuinely absent, `CasFailed` naming the state it found otherwise. `app_core::plans::db_err` routes `CasFailed` to `plan.invalid_state` (blocking, non-retryable) instead of the generic retryable `internal.database`, so the UI stops offering a retry for a request that can never succeed.

Nine tests cover the approval CAS, the reopen error discrimination, and the token-revoked apply path.

**Findings 1 and 2: filed, not fixed here.**

Finding 1 is a spec-017 self-contradiction rather than a bug in this diff, and fixing it means picking which spec statement holds. T023 mandates `draft` as the reopen target; T025 makes `approve_plan` require `ready_for_review`; T027 asserts "PlanDetailPage.tsx already handles draft → Approve & Apply". No standalone `draft → ready_for_review` command exists — every such transition lives inside a plan generator (`confirm.rs:329`, `cleanup_plan.rs:267`, `source_view_generate/generate.rs:627`), and a generator writes its own items so it cannot promote an existing plan. One correction to the finding: `plans.discard` accepts any state except `Applying`/`Paused` (`discard.rs:43`), so a reopened plan is discardable and regenerable — "strands the plan" overstates it. Filed as `astro-plan-4puw` (P1).

Finding 2 is codebase-wide publish-after-commit, not specific to reopen: `approve.rs` commits at :81 and publishes at :116, `discard.rs` at :61 then :64, and `crates/audit/src/bus.rs` offers no transaction-aware variant alongside `publish` (:79) / `publish_batch` (:121). Constitution v1.1.0 §II states audit records are user-facing history and not transactional guarantees, and §V places audit rows in Tier 2 where batched or asynchronous writes are permitted. So this is the documented design position. The question worth settling once is whether any topic carries user knowledge that is not re-derivable — `plan.reopened` is recoverable from state plus a null `approval_token` — and that is filed as `astro-plan-i3nw` (P3).

srobroek:
Review dimension: correctness + constitution + body-accuracy. Round 1. Reviewed PR head 4dd35388 against `origin/main` (merge-base d3d4c10d7).

The reopen path itself is sound: the transition set is read from `domain_core::lifecycle::plan::is_allowed` rather than a hand-written list, `set_reopened` nulls the approval in the same statement as the state change, and the tests cover both refusal states, both reopenable states, idempotence, and the audit event. The mutation proof in the description checks out.

Requesting changes on four points. All four are already fixed in the local commit 83a5ef854 (`fix(plans): guard approval and apply against a concurrent reopen`) on the `feat/yq33-reopen-plan` branch — that commit is not on this PR head, so pushing it resolves 1-4.

**FIX 1 — the branch tip is not pushed.** `origin/feat/yq33-reopen-plan` (head 4dd35388) contains only `e99b85e7c` plus a merge of main. The local branch carries a second commit, 83a5ef854, that this PR does not include. Push it; the description's concurrency claims read as true only with it.

**FIX 2 — `crates/persistence/plans/src/repositories/plans.rs:396-409`.** `set_approved` runs `UPDATE plans SET state = 'approved', ... WHERE id = ?` with no state predicate, and `approve_plan` (`crates/app/core/src/plans/approve.rs:48-55`) reads the state before calling it. A reopen landing between that read and the write is clobbered: a plan the user just returned to `draft` becomes `approved` again with a live token. Add the state predicate to the `UPDATE`.

**FIX 3 — `crates/persistence/plans/src/repositories/plan_apply.rs:125-126`.** `cas_approved_to_applying` predicates on `state = 'approved'` only. `apply_plan` verifies the supplied token and loads the item rows before that CAS (`crates/app/core/src/plan_apply/apply.rs:54,72,134`). A reopen, edit, and re-approval between those steps leaves the CAS satisfied, so the apply proceeds against the pre-edit item snapshot with a token the reopen revoked — a filesystem mutation from an intent the user withdrew (constitution II). Bind `approval_token` into the CAS predicate.

**FIX 4 — `crates/app/core/src/plans/reopen.rs:94`.** `set_reopened` returns `DbError::NotFound` both when the plan is absent and when it is present in a non-reopenable state, and `db_err` maps that to `plan.not_found`. A reopen that loses the race with a transition into `applying` therefore tells the caller a plan it has open no longer exists. Distinguish the two and report the lost race as `plan.invalid_state`.

Description accuracy: bullet 3 states that the `WHERE`-clause predicate means "a concurrent transition into `applying` cannot be clobbered". That is true for the reopen write itself, but the paragraph reads as a closed concurrency story while the approve and apply writes are unguarded in the other direction. Please narrow or extend it to match whatever lands.

PR #1670, opened 2026-08-20
ci: fail the build when a generated stylesheet module has no importer

srobroek:
## Summary

- An unimported `*.css.ts` emits no CSS, so whatever it was written to style renders unstyled while the build succeeds and the tests pass. The component falls back to a plain-CSS rule that may itself be scheduled for deletion. `scripts/check-orphan-ve-modules.mjs` fails the build when a vanilla-extract module is imported by nothing.
- The gate walks `apps/desktop/src` from the two Vite entry points (`main.tsx`, `splash/main.ts`) and reports every `*.css.ts` the production graph does not reach. A module imported only by a test, or only by another orphan, counts as unreached. Import specifiers come from TypeScript's own preprocessor, so a commented-out import is not an importer; CSS `@import` is matched after block comments are stripped.
- `scripts/check-orphan-ve-modules.test.mjs` runs in the same lint chain as the gate. It exercises the detector against fixture trees where the orphan set is known and asserts that the entry-point guard fires, because a gate that reports zero orphans is otherwise indistinguishable from a gate that cannot find any.
- Five modules with no importer are deleted: `components/DetailPanel.css.ts`, `components/GroupHeader.css.ts`, `components/ListGroup.css.ts`, `features/projects/lifecycle-stepper.css.ts`, and `features/settings/naming.css.ts`. Each was written as a replacement for a plain-CSS BEM block that no component imported, so vanilla-extract compiled none of their rules while the components kept emitting BEM class names from the stylesheets that stayed live. Dead plan-panel BEM rules in `styles/components/inbox.css` go in the same pass.
- `features/projects/projects-table.css.ts` and `features/targets/targets-table.css.ts` are kept. #1668 gave both production importers; this branch merges `origin/main` and drops its earlier deletion of the two.
- knip does not report unimported vanilla-extract modules. Verified by planting an orphan `.css.ts` and an orphan plain `.ts` under `apps/desktop/src`: neither `npx knip` nor `npx knip --include files` named either file. This corrects the claim recorded on astro-plan-mv02 and kyo7.109.3 that enabling knip's unused-files check would cover this.

## Test plan

- `pnpm -C apps/desktop lint`: green. The gate reports all 51 vanilla-extract modules reachable; the self-test reports PASS.
- `pnpm -C apps/desktop typecheck`: green.
- `pnpm vitest run` in `apps/desktop`: 2378 of 2380 passed. `src/data/theme.persistence.test.ts` fails on `origin/main` content this branch does not touch (astro-plan-0z1rj); `src/dev/devSurface.release.test.ts` timed out in the full run and passes in isolation.
- Mutation-checked: planting `apps/desktop/src/planted.css.ts` fails the gate and names the file, both with no importer and with a commented-out importer in `main.tsx`.

## Spec Context

Tracks-Bead: astro-plan-ibk5
Tracks-Bead: astro-plan-mv02
Merge-Bead: astro-plan-ale7


srobroek:
@codex review

srobroek:
All three findings addressed in 370ed828f. You were right that a substring search over a text corpus cannot answer this question; the gate now walks the production import graph.

**Arbitrary text matching** — specifiers are parsed, not searched: static `import` / `export ... from`, bare side-effect `import`, dynamic `import()`, and CSS `@import`. A comment has no specifier, so it matches nothing. Your `PropertyTable.css` masking `Table.css.ts` case is now a self-test case, as is the comment-mention fixture you described.

**Test-only importers** — test files never enter the BFS queue, so `ve-class-application.test.tsx` no longer keeps a stylesheet alive. Reachability also catches a third case neither of us named: a module imported only by another orphan.

**Missing self-test** — `scripts/check-orphan-ve-modules.test.mjs`, wired into the `test` recipe beside `check-eslint-baseline.test.mjs` and `check-mock-baseline.test.mjs`.

Two things the fix needed beyond the finding: resolution through the `@/` alias (declared in both `tsconfig.json` and `vite.config.ts`; without it reachability found almost nothing and reported 30 false orphans), and both HTML entry points from `rollupOptions.input`, since `splash.html` loads `src/splash/main.ts` separately. Added a vacuous-pass guard for the case where the entries reach nothing.

Verified: all 49 modules reachable, exit 0; self-test passes.

srobroek:
Verification review (round 1).

The gate is wired into a job that runs: the `apps/desktop` `lint` script invokes it, and ci.yml line 410 runs that script. `scripts/**` sits in `force_full_shared` (ci.yml line 141), so a change to the gate script activates the frontend lane. The seven deletions check out: `git grep` at merge-base `d3d4c10` finds zero importers for any of the seven `*.css.ts` modules, and the ten removed `.pv-plan-panel__*` / `.pv-listpage__detail-body` rules have no remaining reference outside `specs`. `.pv-listpage__detail-body` keeps its canonical definition in `styles/components/tables-lists.css` at line 312. The gate reports OK over 49 modules; the self-test passes.

Two verification defects.

**1. The self-test runs in no CI job.** Its only invocation is the `test` recipe in the justfile, and CI does not run `just test`. The root `test` script is `cargo test --workspace && pnpm -r --if-present test`, and vitest's include is `src/**/*.{test,spec}.{ts,tsx}` (vitest.config.ts line 51), which does not reach `scripts`. So the fixture suite that is this PR's evidence for the detector never executes on a PR. The script's own comment at lines 239-241 names exactly this failure mode for `check-mock-baseline.mjs`. The two sibling self-tests share the gap, so the precedent is the bug, not the excuse.

Required: append `node ../../scripts/check-orphan-ve-modules.test.mjs` to the `apps/desktop` `lint` script, next to the gate invocation this PR already edits, or add a CI step that runs it.

**2. A commented-out import keeps an orphan reachable.** The specifier patterns at lines 99-104 scan raw source. Running the second pattern against the single line `// import './orphan.css';` captures `./orphan.css`. The docstring at line 98 states the opposite: "a comment naming a module matches nothing: a comment has no specifier". The fixture at test lines 61-69 only covers the prose form (`// ./commented.css is only mentioned here`), not the commented-statement form. This is a fail-open: comment out a component's stylesheet import during a refactor and the gate stays green while the bundle loses the CSS. A commented import already exists in the tree, in the `lib/i18n.ts` docstring at line 14, so the shape is not hypothetical.

Required: strip comments before matching, or otherwise exclude commented specifiers, and add a fixture asserting that a module whose only importer is a commented-out `import` statement is reported as an orphan.

Non-blocking: the entry-point guard at line 242 throws `ERR_INVALID_ARG_TYPE` from `pathToFileURL(process.argv[1])` whenever `process.argv[1]` is undefined, which is any `node -e` importer. Guarding the argv read would make the module importable from an eval context.

srobroek (commented):
Independent round-2 review (agent reviewer, wisp astro-plan-wisp-evl, opus, its own prepared checkout at a3b790e18).

VERDICT: APPROVE. Both round-1 FIX items verified as genuinely fixed at the reworked head. Every claim in this PR body holds against the diff and HEAD.

Recorded by the orchestrating session on the reviewer's behalf: GitHub refuses `--approve` on a PR from the same account, so this is a comment review rather than a formal approval. The review itself was performed by a separate agent that did not write this branch.

PR #1671, opened 2026-08-20
fix: settings panes no longer show the value a toggle just replaced

srobroek:
## Summary

- `settingsQueryOptions` caches each scope with `staleTime: Infinity`, which is only safe because `updateSettings` invalidates the key afterwards. Several modules call `commands.settingsUpdate` directly and skip that. A new CI check fails when a settings write bypasses the invalidation chokepoint.
- Two of those bypasses were live bugs, not latent ones, and this branch fixes them. `LogPanelContext.tsx` wrote the `advanced` scope and `StepCatalogs.tsx` wrote the `cleanup` scope; `Advanced.tsx` and `Cleanup.tsx` read those same scopes through `useQuery` with `staleTime: Infinity`. Toggling "remember follow logs" from the log panel, or the default protection level in the setup wizard, left the corresponding settings pane displaying the value the user had just replaced until the pane remounted. Both writes now go through `updateSettings`.
- The remaining bypasses are latent: no `useQuery` reads the scopes they write. A bypass becomes a real bug the moment someone adds a reader, and the symptom is a pane that ignores a change the user just made, with nothing in the diff to suggest why. The constraint used to live in a docstring and a PR body, neither of which can hold a line.
- The remaining writers are allowlisted rather than prohibited, each annotated with the scope it writes and why it cannot use the chokepoint. Theme and locale write during boot, before a `QueryClient` necessarily exists, so rewriting them is a separate change from stopping the set from growing. A new bypass fails until it is either routed through `updateSettings` or added deliberately.
- The gate refuses to pass vacuously: if it cannot find the chokepoint itself calling `settingsUpdate`, it fails rather than reporting zero bypasses. A regex over a directory tree can match nothing or match everything, and matching nothing is indistinguishable from a clean repo.
- `scripts/eslint-alm-baseline.txt` re-points its `StepCatalogs.tsx` entry from line 168 to 173. That baseline is keyed by file + line + rule, so the lines added above the component shifted a grandfathered `alm/require-root-testid` violation and the gate read it as new. The violation is unchanged.

## Test plan

`node scripts/check-settings-write-invalidation.test.mjs` covers six cases against a synthetic tree, and the gate was mutation-checked:

- a new direct writer fails, naming the file
- breaking the grep pattern fails with "not measuring anything" rather than a false OK
- a doc-comment mention is not treated as a call, which caught a wrong allowlist entry while the gate was being written (`settingsQueries.ts` only names the command)
- tests may write freely; paths recurse and sort

## Spec Context

Merge-Bead: astro-plan-uvgo


srobroek:
@codex review

srobroek:
All six findings fixed in 46028338d — plus two live bugs the first finding uncovered.

**The chained-call miss was not latent.** `app/LogPanelContext.tsx:158` wrote `advanced` and `features/setup/steps/StepCatalogs.tsx:88` wrote `cleanup`, and both scopes have a cached reader (`Advanced.tsx:83`, `Cleanup.tsx:142`). So toggling follow-logs or the default protection level left the matching Settings pane on the replaced value for the session. Both now call `updateSettings`. The matcher allows whitespace around the property access.

**`settingsRestoreDefaults`** is matched as a write. Its wrapper invalidates `queryKeys.settings.all()`, so a direct call strands every cached scope rather than one; `shared/planner/guidance-settings.ts:165` is the existing caller and is baselined.

**Cached readers for bypassed scopes.** Each allowlist entry declares the scopes its writes affect, and a scope gaining a `settingsQueryOptions` reader now fails as `[cached-reader]`. Scope arguments are module-local constants at every call site, so `resolveScope` reads the local declaration — literal-only matching resolved nothing. Scopes derived from the call site are compared against the declaration (`[scope-drift]`) so a bypass that changes scope stays covered; `data/persisted-state.ts` is exempt because its scope is a parameter.

**Stale entries** are `[stale]` errors, not warnings.

**Windows paths** normalize through `toPosix` before any comparison.

**Self-test** runs from the `test` recipe beside its two siblings. 22 cases: chained call, restore-defaults, scope resolution and the unresolvable case, reader collection, all four violation kinds, separator normalization, and the entry-point guard. `violations()` takes the allowlist as a parameter so each kind is driven from a fixture rather than the repo's own passing set.

Verified: gate reports 6 baselined bypasses over 4 cached scopes, exit 0; 22 self-test cases pass; `tsc` clean; the four LogPanel suites pass (16 tests).

Two failures on this branch are pre-existing — both reproduce with my edits stashed: `alm/require-root-testid` on StepCatalogs.tsx, and `theme.density-fontsize.test.ts` failing only when run in a single vitest invocation alongside the three setup suites (cross-file isolation, not a regression here).

srobroek:
Closing per owner decision, on the evidence in bead astro-plan-45he0.

The stale-value defect this targets was fixed by #1660. This branch adds a CI guard against regression, and main already performs that check: `apps/desktop/src/data/settingsWrite.chokepoint.test.ts` is present on main and tolerates zero bypasses.

The guard as written cannot fail. `scripts/check-settings-write-invalidation.mjs:43` names the pre-#1660 chokepoint, so it fails its own anti-vacuity check against main and would land as 574 lines of always-passing CI, which reads as coverage without providing any. The script exists only on this branch, so closing leaves nothing behind on main.

The one residual gap is tracked separately as astro-plan-manpc: `settingsRestoreDefaults` at `apps/desktop/src/shared/planner/guidance-settings.ts:154` bypasses the chokepoint.

PR #1672, opened 2026-08-20
ci: require a tracking reference for newly baselined dead functions

srobroek:
## Summary

- The dead-caller baseline is a debt ledger, and 131 of its 148 entries carried no bead or issue reference, so nothing connected a function that never runs back to work that would make it run. An untracked entry reads as settled.
- That is how a missing `acquisition_fingerprint` writer left calibration matching disabled for about two and a half months: the gap was recorded, the record went nowhere, and the specs kept reporting the feature as shipped. See astro-plan-siyk, fixed in #1655, and the recurring pattern on astro-plan-68c7, which now has four instances.
- A newly added name must carry an `astro-plan` id or a `#NNNN` reference in the comment block directly above it. One comment block covers the whole contiguous group of names beneath it.
- Deleting the reference from an entry that already had one is reported as a regression, so the rule holds after the entry lands.
- Existing entries are grandfathered on purpose. Demanding 131 back-annotations at once would either block unrelated work or be satisfied with a copy-pasted placeholder, which is worse than silence because it looks like tracking. The file shrinks toward fully tracked as it turns over.
- The check diffs the baseline against the merge base of `HEAD` and `origin/main`, so it judges only what the branch itself adds. Comparing against main's tip instead reports every name main deleted from the baseline after the branch point as one the branch added — the last six baseline-touching commits remove 39 entries, enough to fail a branch that never opens the file. A PR job checks out at depth 1 and has no merge base to find, so `resolve_base_commit` deepens the history until one exists: `--deepen` first, then `--unshallow`, which is the rung that lifts the checked-out branch's own shallow graft. When no merge base is reachable the check fails closed under `CI` and prints a note locally, because a gate that silently skips in CI reads as enforced while checking nothing.
- On a main push `origin/main` is the commit just pushed, so the check falls back to `HEAD~1`. The `dead-caller-main` canary job now checks out with `fetch-depth: 2`; at the default depth 1 that commit is absent and the job fails closed.
- `tracking_self_test` and `base_resolution_self_test` run from the enforcement path, not only under `--self-test`, so a classifier that stops classifying, or a base that stops being the branch point, fails the build instead of passing everything.

## Test plan

`scripts/check-dead-callers.sh --self-test` asserts, on fixture baselines: a group block carries its reference to every name under it, while a blank line and a comment after a name reset the block; a `#NNNN` reference is accepted; a pre-existing untracked name stays grandfathered; a new untracked name is reported while a group-mate under a tracked comment is not; an entry that loses its reference is reported as `regressed`.

`base_resolution_self_test` builds a depth-1 checkout of a fixture repository whose main drops an untracked baseline entry on a commit the branch does not have, then asserts that the resolved base is the branch point and that the branch — which never opens the baseline — passes.

Mutation-verified against the real baseline, since a gate that never fires and one that always fires are equally useless:

- appending an untracked name fails with `added: zz_untracked_probe`
- appending a name under an `astro-plan` comment passes
- forcing `classify_baseline` to return `T` for everything fails the pre-flight guard with three failed self-test cases, exit 1
- restoring the base-ref-tip fallback fails both `base_resolution_self_test` assertions and prints `added: deleted_on_main`
- the 148 existing entries, untouched, pass
- a depth-1 clone of the branch with `DEAD_CALLERS_BASE_REF` pointing at its own ref and `CI=1` cannot resolve a base, exit 1; the same clone at depth 2 passes
- a depth-1 clone of main plus a depth-1 fetch of this branch from `origin` — the PR job's shape — has no merge base before the run and resolves `d3d4c10d7` during it, exit 0

## Spec Context

Merge-Bead: astro-plan-wmry


srobroek:
@codex review

srobroek:
All three findings fixed in 3b0f642.

**Fetch the base ref** — `resolve_base_commit` now fetches `refs/heads/main` at depth 1 when the ref is absent (the shallow-checkout case at `ci.yml:280`), then takes `merge-base HEAD origin/main`. When that resolves to HEAD itself — the post-merge canary at `ci.yml:866` — it falls back to `HEAD~1` so the canary checks the pushed commit's own additions instead of comparing the baseline with itself. If no base is reachable and `CI` is set the check now fails rather than skipping; the silent skip is what made it unreachable. Override with `DEAD_CALLERS_BASE_REF`.

**Preserve tracking references** — both revisions are classified tracked/untracked, so there are two violation classes: `added` (a name this revision introduces without a reference) and `lost ref` (a name that held a valid reference in the base and no longer does). Pre-existing untracked names stay grandfathered.

**Group blocks** — `classify_baseline` carries the active comment block across the whole contiguous group. A blank line ends the group; a comment line following a name opens a new block rather than extending the previous one. This matches the format at `dead-callers-baseline.txt:32-48` and `:71-96`.

`tracking_self_test` covers each: group carrying, both resets, `#NNNN` acceptance, grandfathering, a group-mate accepted, a bare addition rejected, and a merged entry losing its reference. Wired into `--self-test`.

Verified: both self-test suites pass; the full run reports 148 baselined, exit 0; appending `probe_untracked_name` fails against base `5f15a992` (real merge-base, not a skip); shellcheck clean.

srobroek:
Review dimension: verification (round 1). Head `ce8cb11` vs base `03d61d2`.

Local evidence: `scripts/check-dead-callers.sh --self-test` passes both blocks; the enforcing run reports `OK: no new dead pub fns (148 baselined)`.

1. `scripts/check-dead-callers.sh:635` — the enforcement path runs only `self_test`; `tracking_self_test` runs only under `--self-test`, and no workflow passes that flag (`rg 'self-test' .github/workflows` matches only `ci-affected-crates.sh` and `e2e-app-cache-key.sh`). The new tracking logic ships with no automated verification — the vacuous-green shape this script exists to prevent. Required: invoke `tracking_self_test` from the same pre-flight guard in `main()`, or add a CI step that runs `--self-test`.

2. `scripts/check-dead-callers.sh:377-381` — the `HEAD~1` fallback cannot resolve in a depth-1 clone, so the `dead-caller-main` push job (`.github/workflows/ci.yml:915-922`, plain `actions/checkout@v4`, no `fetch-depth`) gets an empty base and, with `CI` set, hard-fails at `:391`. Reproduced: `git clone --depth=1 -b main` (so `origin/main == HEAD`, `HEAD~1` absent) then `CI=1 bash scripts/check-dead-callers.sh` → `FAIL: cannot resolve a base revision for the tracking check`, exit 1. That turns the main canary red on every push. Required: deepen the history (`git fetch --deepen=1` / unshallow) before the `HEAD~1` fallback, or set `fetch-depth: 2` on that job, and re-verify against a shallow clone.

3. `scripts/check-dead-callers.sh:270-281` — the fixture-list comment that documented `self_test` (adjacent to it at base `03d61d2:281`) is now separated from it by the inserted tracking block, so it reads as documentation for `classify_baseline`. Required: move it back directly above `self_test()` at `:497`.

4. PR body — "it skips with a note when `origin/main` is unavailable rather than failing closed" contradicts `:390-393`, which fails closed whenever `CI` is set; the test plan also lists four mutation cases while `tracking_self_test` asserts five and omits the lost-reference case. Required: restate the body against head behaviour.

srobroek:
Round-2 review (dimension: rework verification + body accuracy). Verdict: **changes**.

Round-1 items 1-3 verified fixed at head `855efc27`:
- `tracking_self_test` runs from the enforcement path (`scripts/check-dead-callers.sh:641-644`); mutating `classify_baseline` to return `T` for every line fails the pre-flight with 3 failed cases, exit 1 — matches the body.
- `dead-caller-main` checks out with `fetch-depth: 2` (`.github/workflows/ci.yml:917-923`). Reproduced: depth-1 clone of main + `CI=1` exits 1 ("cannot resolve a base revision"); the same clone at depth 2 exits 0 with `OK: no new dead pub fns (148 baselined)`.
- The fixture comment sits directly above `self_test` (`scripts/check-dead-callers.sh:486-498`).

Two items remain.

**FIX 1 — `scripts/check-dead-callers.sh:353-371` + `:316-340`: on the PR job the comparison base is main's *tip*, not the merge-base, so the check reports names the branch never added.**
The `integration` job uses the default depth-1 checkout, and `resolve_base_commit` fetches the base with `--depth=1`, so `git merge-base HEAD origin/main` finds no common commit and the fallback picks the ref tip. Reproduced in a depth-1 clone of this head: `origin/main` resolves to a tip with no merge-base, so the comparison runs against that tip.
Consequence, reproduced with real baselines: comparing main's tip baseline against the baseline at commit `6e4199957` (a plausible branch point) reports `new default_label`, `new delete_source_protection`, `new file_record_exists` — three entries main deleted since, on a branch that never touched the file. The last six commits touching the baseline remove 39 entries, so any branch older than a removal fails this gate until it merges main. Back-annotating an existing entry on main produces the same failure as `regressed`.
Required: resolve a real merge-base (deepen/unshallow the base fetch), or restrict `new`/`regressed` to names the branch's own baseline diff changes. The body sentence "The check diffs the baseline against `origin/main` so it only judges what a branch adds" does not hold and must be restated.

**FIX 2 — PR body bullet 1 and `scripts/check-dead-callers.sh:275,284`: the count 134 is wrong.**
Running the shipped `classify_baseline` over `scripts/dead-callers-baseline.txt` gives 17 tracked / 131 untracked of 148, at this head and at `5f15a992b`, `7314ad67b`, and `eca1f1e19`. The 17 tracked names come from 4 comment blocks. Required: restate as 131 (and 131 back-annotations) in both the body and the script comment.

A formal `--request-changes` review is not available to me on this PR (self-authored); recorded on `astro-plan-wisp-1du`.

PR #1673, opened 2026-08-20
perf: reconcile a library root in linear time instead of quadratic

srobroek:
## Summary

- `reconcile_root` scanned the whole disk-file `Vec` once per known frame, making a reconcile pass O(frames x files). 10,000 frames took 7,509 ms while 2,500 took 591 ms: four times the frames for 12.7 times the wall time, with only 9 SQL statements issued for the whole pass.
- The listing is now collected into a `HashSet<PathBuf>` once and each candidate looked up. Timings after: 2,500 in 163 ms, 10,000 in 692 ms, 40,000 in 3,753 ms. Four times the frames now costs 4.2 times the time, so the complexity is linear rather than merely faster.
- The replaced `paths_match` had an `a == b || a.components().eq(b.components())` fallback documented as covering Windows casing and separator differences. The second arm is unreachable: std defines `impl PartialEq for Path` as `self.components() == other.components()`, and `impl Hash for Path` hashes component-wise over the same normalisation. Set membership therefore matches exactly what the linear scan matched, on every platform, because the normalisation is in `components()` rather than in the removed helper.

## Test plan

`cargo test -p fs_inventory`, plus a test pinning the redundant-separator and `.`-segment cases the set relies on. Only the `reconcile_root_frames` `wall_ms` baseline is re-recorded.

## Spec Context

Tracks-Bead: astro-plan-6lsz
Merge-Bead: astro-plan-a3id

### DIFF
--- a/crates/fs/inventory/src/reconcile.rs
+++ b/crates/fs/inventory/src/reconcile.rs
@@ -23,6 +23,7 @@
 //! aborts with [`RootUnavailable`] otherwise. Without that guard an unmounted
 //! drive reads as "every frame deleted".
 
+use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 
 pub use fs_pathsafe::RootUnavailable;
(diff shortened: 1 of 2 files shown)


srobroek:
@codex review

PR #1674, opened 2026-08-20
fix: onboarding spotlight no longer gives up on a control that is still loading

srobroek:
## Summary

- The onboarding spotlight gave up on controls that were about to render. `FindSpotlight` split its resolve budget as 3000 ms before arrival and 1500 ms after, on the rationale that navigation is the slow phase.
- No route in `router.tsx` declares a loader or `ensureQueryData`, so a route commit waits only on a lazy chunk and page data is fetched after the commit inside the mounted component. The waiting therefore happens after arrival, and the data-gated anchors (inbox rows, sessions rows, master rows) are exactly the ones that mount late. A premature `missing` is terminal, because the resolve effect re-runs only on `[selector]`, so the callout persisted even once the anchor appeared.
- Post-arrival goes to 4000 ms. The pre-arrival grace stays at 3000 ms rather than shrinking, because it also covers the deep-link `fetchFirstSessionId` lookup that precedes navigation. Worst-case total rises from 4.5 s to 7 s: a slightly longer successful wait beats a spotlight that abandons a control that was about to render.

## Test plan

`pnpm test` in `apps/desktop`. The arrival test probes were pinned at 2 s, which no longer distinguishes the phases; they now bracket each deadline, and a new case renders the anchor 2.5 s after arrival. Three of the six cases fail under the old split. The line number in the alm baseline shifts with the longer comment.

## Spec Context

Tracks-Bead: astro-plan-l3tl
Closes-Bead: astro-plan-l3tl
Merge-Bead: astro-plan-bafl

### DIFF
--- a/apps/desktop/src/features/onboarding/FindSpotlight.arrival.test.tsx
+++ b/apps/desktop/src/features/onboarding/FindSpotlight.arrival.test.tsx
@@ -2,18 +2,16 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 
 /**
- * Which signal starts the spotlight's short anchor deadline (spec 056 FR-022).
+ * Which signal starts the spotlight's anchor deadline (spec 056 FR-022).
  *
  * Two budgets share one resolve loop: a 3s pre-arrival navigation grace, and a
- * 1.5s post-arrival anchor wait restarted from the moment the target route is on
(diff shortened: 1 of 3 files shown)


srobroek:
@codex review

PR #1676, opened 2026-08-20
refactor: split the inbox page into data, selection, summary, status, and top-bar layers

srobroek:
## Summary

- `InboxPage` was 1012 lines with a cognitive complexity of 30 against a ceiling of 20, around 78 hook calls, and a render block of roughly 450 lines. This is movement only: no behaviour, prop value, or `data-testid` string changes.
- Extracted layers:
  - `useInboxListData`: list and open-plan queries, root derivation, rescan targets, destination roots, and the page-level search filter.
  - `useInboxSelection`: URL-backed selection plus the `reclassify_v2` re-split and classify-split handoffs and their shared stale-cleanup gate.
  - `useInboxPlanSummaries`: the per-ingestion frame-type hint and the merged three-source breakdown tallies `PlanPanel` renders.
  - `useInboxPageStatus`: derived stats, folder and master summary, and the memoised status-bar node.
  - `InboxTopBar`: `FilterToolbar` wiring and the review-plans, bulk-confirm, and rescan actions with their render conditionals.
- The `InboxDetail` remount key stays inline in `InboxPage`. `InboxDetailProps` is already exported, so a wrapper component would have forwarded twelve identical props for one key expression.
- `PageTopBar` renders `data-testid="topbar"` on its own root, so the alm require-root-testid rule cannot see it on the caller's JSX node. The escape hatch and its reason match `MasterDetail.tsx:113`, and a comment now records that.
- `FindSpotlight.anchors.test.ts` follows the `inbox.confirm-row` anchor to its new file. The anchor string itself is unchanged.

## Test plan

`pnpm test` and `pnpm lint` in `apps/desktop`.

## Spec Context

Tracks-Bead: astro-plan-kyo7
Merge-Bead: astro-plan-3znx

Refs astro-plan-kyo7.112.

### DIFF
--- a/apps/desktop/src/features/inbox/InboxPage.tsx
+++ b/apps/desktop/src/features/inbox/InboxPage.tsx
@@ -29,55 +29,32 @@
  *
  * spec 039: the list is a cross-root aggregate of all unacknowledged items
  * (inbox.list), grouped/labelled by their registered root.
+ *
+ * The page is composition only. Data and derivation live in
+ * `useInboxListData`, selection and its two mid-flight handoffs in
+ * `useInboxSelection`, plan tallies in `useInboxPlanSummaries`, the status-bar
+ * slot in `useInboxPageStatus`, and the top bar in `InboxTopBar`.
  */
 
-import { useNavigate, useSearch } from '@tanstack/react-router';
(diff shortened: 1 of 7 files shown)


srobroek:
@codex review

PR #1677, opened 2026-08-20
docs: retire three superseded spec task checklists

srobroek:
## Summary

- This repo does not author `tasks.md`: task state is recorded in beads, and a pre-tool hook says so on every read. The `tasks.md` files under specs 005, 014, and 052 outlived that rule and were still presenting checkbox counts as if they meant something. They are deleted.
- Checkbox state at deletion, with the successor that overtook each file:
  - 005 inbox-mixed-folder-split, 0 of 51 checked. Superseded by 041, whose single-type model shipped.
  - 014 catalog-index-licensing, 15 of 20. Superseded by 035; the download-bundled-catalog mechanism was abandoned.
  - 052 simbad-caching-dual-lookup-cone-search, 12 of 35. Implemented, and its own banner already disowned the file, noting that T002 asks to bump simbad-resolver to 0.2.0 while the workspace is on 0.5.0.
- Verified before deleting that each `spec.md` carries an authoritative status banner, so no status record is lost. 005 and 014 both open with a superseded notice naming their successor, and 052 with a post-hoc implemented record.
- Live pointers into the deleted files are dropped: the 005 and 052 `plan.md` documentation trees lose their `tasks.md` entry, and 005 `research.md` loses its citations of task ids `T0-IMAGETYP-Research`, `T-NormTable`, and a repair task that existed only in the deleted checklist.
- The 052 `spec.md` banner said its outstanding process gates were tracked separately without naming a tracker. It cites `astro-plan-n6zv` for the speckit-verify run and `astro-plan-ltoy` for Windows verification.

## Test plan

No code paths touched.

`rg -n 'tasks\.md'` over `specs/005-inbox-mixed-folder-split/`, `specs/014-catalog-index-licensing/`, and `specs/052-simbad-caching-dual-lookup-cone-search/` returns one hit: `specs/052-simbad-caching-dual-lookup-cone-search/spec.md:10`, the sentence that records the retirement itself.

Two repo-wide references to the deleted paths remain, at `specs/GRILL_DECISIONS_2026-05-21.md:360` and `:654`. They stay: that file is a dated decision record, and a stale path in it is preferable to editing the record.

## Spec Context

Tracks-Bead: astro-plan-44tb
Closes-Bead: astro-plan-44tb
Merge-Bead: astro-plan-6cg2


srobroek:
@codex review

srobroek:
Worked both findings. The dangling-reference half was real and is fixed; the preserve-the-checklist half rests on a premise this repo does not hold, so it is answered with a bead rather than by keeping the file. Commit is blocked on a hung 1Password `op-ssh-sign` prompt, so the edits are local.

**Finding 2 — surviving references. Partly confirmed, fixed.** Two of the three cited references exist:

- `specs/005-inbox-mixed-folder-split/plan.md:88` listed the checklist in its documentation-layout tree. Removed from the tree, replaced with a pointer to `bd list --spec 005-inbox-mixed-folder-split`.
- `GRILL_DECISIONS_2026-05-21.md:654` (and `:360` for spec 014, which the finding did not mention) list the checklist under "Affected artifacts (2026-05-22)". That is a historical record of what one review pass touched, so deleting the bullet would falsify it; both are annotated "(since retired; beads holds the task state)" instead.

The rest of the finding does not hold at this revision. `T027a`, `T-PlanRepair` and `T030` are not cited anywhere outside the deleted file, and `GRILL_DECISIONS` does not use them as supporting artifacts for repository invariants — `rg 'T027a|T-PlanRepair'` across the repo returns nothing but the file being deleted. No archival copy is warranted for IDs nothing references.

**Finding 1 — keep the checklist until verification completes. Premise rejected, substance kept.** Retaining a reconciled checklist would not unblock the gate. `gates.yaml` declares `verify` and `verify-tasks` as `requires: [tasks]` and `check-prerequisites.sh` exits when the file is absent, but this repo's speckit guard makes `specs/*/tasks.md` read-only legacy that must not be written or created, so the input those gates want cannot be maintained here at all. Keeping the file would preserve a stale artifact and still not let the gate run.

What is real is the record: `T034`/`T035` were the only trace that speckit-verify and Windows verification had never run for spec 052, and the spec's status line said only that they were "tracked separately" without naming where. Filed as `astro-plan-n6zv`, which owns the FR/SC cross-check, the Windows verification (gated behind the journey-validation run), and the underlying question of whether the `requires: [tasks]` prerequisite should read beads instead. `spec.md` now cites the bead by id, so "tracked separately" is checkable.

PR #1678, opened 2026-08-20
fix(ci): stop hiding the primary CI workflow from every scanner

srobroek:
## Summary

- `.github/workflows/ci.yml` was tracked and also listed in `.gitignore` as `# parked: needs workflow-scoped push`. git keeps honouring a tracked file, so the entry changed nothing about commits or CI and did not warn. What it changed is that every ignore-respecting traversal skipped the primary workflow.
- Measured on `main`: `rg -n 'run: just ' .github/workflows/` returned 0; `rg --no-ignore -n ...` returned 5. ripgrep, opengrep, gitleaks, trivy, zizmor, actionlint, osv-scanner, and ast-grep were all reporting a clean surface they never opened.

## Changes

- The `.gitignore` entry is gone.
- `scripts/check-github-not-ignored.sh` seals it at zero and runs from `just lint`. It fails when `git ls-files -i -c --exclude-standard -- .github` returns anything. Verified in both directions: clean after the fix, and it names `ci.yml` when run against the pre-fix `.gitignore`.
- ci.yml:518 carried the one actionlint shellcheck issue the invisibility hid. `$args` holds a `-p <crate>` word list that has to split, so the SC2086 quote would collapse it into one bad argument. A disable directive and a reason sit above it. `actionlint .github/workflows/ci.yml` is now clean.

## Out of scope

- zizmor reports 10 `unpinned-uses` and 6 `artipacked` against ci.yml. That is the same unpinned-action posture as the workflows that were never hidden (`e2e.yml` reports 23 and 9), so it is a repo-wide policy question rather than something this entry concealed.
- Other tracked-but-ignored paths exist outside `.github/` (`.claude/settings.local.json`, seven review screenshots). The guard is scoped to `.github/` because that is where an invisible file changes what CI enforces.

## Spec Context

Tracks-Bead: astro-plan-o1om
Merge-Bead: astro-plan-6t6y

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -523,6 +523,10 @@ jobs:
             # zero runnable tests (e.g. an e2e_tests-only change — all its
             # tests are #[ignore]d for the Layer-2 e2e.yml job), and nextest
             # otherwise exits 4 ("no tests to run", run 28751553778).
+            #
+            # $args is a word list of `-p <crate>` flags and has to split, so the
+            # SC2086 quote would collapse it into one bad argument.
+            # shellcheck disable=SC2086
             cargo nextest run --profile ci --no-tests=pass $args
           fi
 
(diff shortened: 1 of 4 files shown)


srobroek:
@codex review

srobroek:
@coderabbitai review

PR #1679, opened 2026-08-20
feat(db): store the relation evidence behind every proposal and revision

srobroek:
## Summary

- Spec 062 requires `RelationEvidence` non-`Option` on `RelationProposal` (`crates/contracts/core/src/relations.rs:372`) and `ManualRelationCreateRequest` (`:939`), and requires `representativeEvidenceId` / `capturedUnionEvidenceId` on the panel-group and mosaic revisions. No table stored any of it, so none of the ten US2 commands could register.
- Migration `0002_relation_evidence.sql` adds the envelope and its three typed lists, mapped from the `relation_evidence` section of `specs/062-session-heterogeneity/data-model.md`. That section is not on `main` yet: it lands with #1657, so merge #1657 first or this PR cites a design nobody can read.
  - `relation_evidence`: four verdict columns, the optional geometry, the input digest, the matching-settings revision the measurement ran under, and one density target per list.
  - `relation_evidence_missing_code`, `_allowed_rotation`, `_measurement`: each bounded at its contract size (100, 16, 100) with a `NOT NULL` ordinal. SQLite evaluates a `CHECK` over a null to null and `UNIQUE` admits repeated nulls, so a null ordinal would otherwise clear both the bound and the per-envelope key while leaving the list unordered.
  - Measured and threshold values are signed integer microunits. `REAL` would let the same measurement produce two different `input_digest` values.
- `relation_evidence_missing_code` also carries `UNIQUE (evidence_row_id, code)`, which the data model does not specify. The same reason listed twice is a corrupt list. Worth a reviewer's eye: this constraint is mine, not the design's.
- The three `NOT NULL` foreign keys the design puts on `relation_proposal`, `panel_group_revision`, and `mosaic_revision` are absent here. SQLite cannot `ADD COLUMN` a `NOT NULL` column carrying a `REFERENCES` clause, so they need a rebuild of three tables holding 14 foreign-key dependents between them, two of them compound. The migration header records the exclusion. The rebuild is an owner decision.
- Merged `main` at `da89da269`. `crates/persistence/core/tests/baseline_invariants.rs` moves its schema-object oracle 309 to 314 for the five new objects. `0001` stays byte-frozen, and `migration_set_has_frozen_0001_and_unique_append_only_versions` confirms `0002` appends legally.
- Track C (topology repository) and Track D (the `relation_proposal.manual.create` use case) remain open, so `astro-plan-ps48` is tracked rather than closed. Without D these tables have no production writer, which is the `astro-plan-68c7` pattern.

## Test plan

`cargo test -p persistence_core`: 8 new tests in `relation_evidence_migration.rs`, whole crate green. `cargo clippy -p persistence_core --tests` clean.

Mutation-checked rather than asserted. Dropping `UNIQUE (evidence_row_id, code)` fails the missing-code test, dropping the `lower_udeg <= upper_udeg` check fails the rotation test, and widening the `footprint_coverage_ppm` range fails the geometry test. The subject-lookup test reads `EXPLAIN QUERY PLAN` and fails if SQLite plans a table scan instead of the index.

## Spec Context

Tracks-Bead: astro-plan-ps48
Merge-Bead: astro-plan-3gni


srobroek:
@coderabbitai review

srobroek:
Closing per owner decision on bead astro-plan-0sgb.6, following two independent review verdicts.

This branch adds a `relation_evidence` table that no code writes. It duplicates `proposal_measurement` with a contradictory CHECK vocabulary, is unreachable from its own subjects, and carries density columns that its own test contradicts. All of that lands inside a frozen append-only migration chain, so a later correction cannot edit the migration -- the cost is permanent and the capability is zero.

The relation-evidence direction continues on #1657, which is being reworked to separate `spec062_config_revision` from `matching_settings_revision` and to stop making an envelope column NOT NULL on rows that `relation_proposal.accept` never writes.

PR #1680, opened 2026-08-20
fix: calibration no longer matches a master when a tolerance or exposure is not a real number

srobroek:
## Summary

A wrong calibration match is silent corruption of the user's science data, and four separate float comparisons in the matcher could produce one.

- `SoftDimConfig::penalty` (`crates/calibration/core/src/ranking.rs`) compared `delta > self.tolerance` and nothing else. Every comparison against a NaN is false, so that test took the match branch for a NaN delta, a NaN tolerance, a negative tolerance, and an infinite tolerance: four ways for an unrelated master to score as a full-confidence match. It now rejects a delta or tolerance that is NaN, infinite, or negative, keeps the boundary inclusive, and clamps the result into `0.0..=1.0`.
- `rank_matches` sorted with `partial_cmp().unwrap_or(Equal)`, which left a NaN pair in input order and let a NaN confidence take the top slot. It then suppressed the ambiguity warning, because `(NaN - second).abs() < 0.05` is also false. Non-finite confidences now map to `NEG_INFINITY` before `total_cmp`, and `suggest_status` returns `"ambiguous"` when either of the top two is not a real number.
- `rules/dark.rs` divided by the master exposure after guarding only `me == 0.0`. A negative master exposure produced a negative percentage difference, which is never greater than the tolerance, so a 300s light matched a -300s dark at full confidence. A zero, negative, or non-finite exposure on either side is now reported as missing metadata.
- Nothing validated the persisted tolerances on the way in. With the matcher now rejecting a garbage tolerance, keeping one would have turned its dimension into a reject-everything rule without telling the user, so `load_config_from_db` falls back to the engine default for a tolerance that is not finite and non-negative, and for an override penalty outside `0.0..=1.0`.

`CalibrationMatch::new` already clamps confidence into `0.0..=1.0`, and `f64::clamp` returns NaN for a NaN input, so NaN is the only non-finite confidence that reaches the ranker. A test pins that clamp so the ranking tests cannot be read as covering the infinities too. A NaN also cannot arrive through the `calibration_tolerances` row: the column is `REAL NOT NULL`, SQLite stores a NaN bind as NULL, and the insert fails.

## Bead claims that do not hold

`astro-plan-vj6x` asserts a boundary split of "9 inclusive vs 4 exclusive". Three of the four exclusive sites (`rules/mod.rs:21`, `dark.rs:48`, `bias.rs:46`) are `HARD_RULE_EPSILON` equality-within-1e-9 tests, which is a different construct from a tolerance window, and the fourth, `rotation.rs:109`, does not exist: rotation lives in `flat.rs:79`. The `families.rs` comparisons the bead groups with `ranking.rs` are integer-only and carry no float hazard. Changing those four sites would be churn for no behaviour change, so this PR leaves them.

## What is split out

The bead also covers `exposure_tolerance_s` and `aging_limit_days`, which the UI writes and matching ignores. That half is now `astro-plan-rcvr`, because it is not a wiring fix: `exposure_tolerance_s` is absolute seconds against a matcher tolerance expressed as a percentage of the reference exposure, and `aging_limit_days` has no `MatchingRuleConfig` field and no rule that consumes master age. Both need a design decision.

## Test plan

`cargo test -p calibration_core -p app_core_calibration -p persistence_calibration`: 178 pass. `cargo clippy` clean on all three. `scripts/check-test-targets-declared.sh` exit 0. Merged `main` and re-ran.

The new tests are mutation-checked rather than asserted. Removing either config sanitizer fails a loader test (`tolerance -3 reached the engine; got -3`). Every ranking and penalty test was confirmed to fail against the old code first. Two of them exposed a wrong premise on my side, that the infinities reach the ranker, and were rewritten against what the constructor actually does.

## Spec Context

Closes-Bead: astro-plan-vj6x
Tracks-Bead: astro-plan-vj6x
Tracks-Bead: astro-plan-rcvr
Merge-Bead: astro-plan-0vta


srobroek:
@coderabbitai review

PR #1683, opened 2026-08-20
fix: catch drift in the generated contract type declarations

srobroek:
## Summary

- `just check-generated` covered `specs/*/contracts/*.generated.json` and
  `apps/desktop/src/bindings/`, and never regenerated
  `packages/contracts/src/generated/`. The 23 committed declaration files had been
  free to rot since they were added, and four of them had.
- `scripts/check-generated-drift.sh` replaces the inline `git diff --exit-code` and
  covers all three paths. It reads `git status --porcelain --untracked-files=all`,
  because `git diff --exit-code` exits 0 for a regeneration that only **adds** a
  file, so a new binding could have landed uncommitted with the gate still green.
- New `check-contracts-ts` recipe runs the declaration half without the cargo
  steps. The existing CI drift step needs the Rust toolchain and system webkit,
  installed only when the `rust` filter fires, so it could not widen to cover a
  schema-only change. A new `spec_contracts` filter triggers the cargo-free leg for
  `specs/*/contracts/**`, which otherwise sets only `docs`.
- `build-schemas.mjs` now fails instead of going quiet on two inputs. A missing
  allowlist entry was skipped, so a renamed spec folder produced fewer declaration
  files and the gate accepted the smaller tree. Two schemas reducing to the same
  stem overwrote each other, last write wins.
- That collision was live. `packages/contracts/schemas/inventory.list.schema.json`
  carries `limit`, `offset`, and `hasMore` and is the copy
  `tests/contract/contract_jsonschema_roundtrip.rs` validates against; the spec 006
  copy never gained those fields, was generated second, and won. Both spec 006
  entries are dropped from the allowlist. Regenerating gives `inventory.list.d.ts`
  the three pagination fields and reflows three files that a
  `json-schema-to-typescript` bump had rewrapped.

Verified: the gate fails on the drift that was already committed and on an
untracked file added to a generated directory, and passes once the regenerated
output is committed. `pnpm --filter @astro-plan/contracts typecheck` and its
conformance suite pass, and shellcheck reports nothing on the new script.

Out of scope, filed separately: the stale spec 006 contract (`astro-plan-9aq6`) and
the absence of any parity check between the two generated TypeScript surfaces
(`astro-plan-nhq9`).

## Spec Context

Repository tooling, outside any spec.

Tracks-Bead: astro-plan-2add
Closes-Bead: astro-plan-2add

Merge-Bead: astro-plan-pzdr

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,7 @@ jobs:
       frontend_full: ${{ steps.decide.outputs.frontend_full }}
       docs_only: ${{ steps.decide.outputs.docs_only }}
       scripts: ${{ steps.filter.outputs.scripts }}
+      spec_contracts: ${{ steps.filter.outputs.spec_contracts }}
       all_files: ${{ steps.filter.outputs.all_files }}
     steps:
       - uses: actions/checkout@v4
@@ -172,6 +173,11 @@ jobs:
               - '**/*.css'
             scripts:
               - 'scripts/**'
+            # Feeds the TypeScript declaration generator, so it has to trigger
+            # the drift gate. Narrower than `docs`, which would put a
+            # Rust-compiling gate behind every markdown edit.
+            spec_contracts:
+              - 'specs/*/contracts/**'
             docs:
               - 'docs/**'
               - 'e2e-agentic-test/**'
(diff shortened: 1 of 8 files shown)


srobroek:
@coderabbitai review

PR #1684, opened 2026-08-20
test: stop the plan-resume retry test from racing the executor it targets

srobroek:
## Summary

- `resume_then_retry_of_pre_pause_failed_item_reaches_terminal_state` fails
  intermittently. It needs the resumed run to still be draining when
  `retry_plan_item` arrives, because that call is refused with a retryable
  `RunNotFound` once the run closes its retry queue, and the run closes it as soon
  as the forward pass drains. The refusal is correct production behaviour; the test
  was betting on timing.
- The only lever the test had was `OTHER_PENDING_ITEMS = 30`, and 30 small
  same-directory moves against a warm pool can finish inside its three awaits. It
  lost that bet on an unrelated calibration PR (#1680, job 96452003924) while the
  other 1798 tests in the job passed.
- `PV_TEST_ITEM_DELAY_MS` adds a per-item sleep to `drive_plan`, read once per run
  so a run cannot change pace midway, and ignored when unset, unparseable, or zero.
  30 items at 20ms bounds the resumed pass at ~600ms, which exceeds the retry call
  by orders of magnitude on any runner.
- The test sets it through an `EnvVarGuard` that removes it on drop, including on
  panic unwind. `cargo nextest` runs one process per test, but plain `cargo test`
  shares one across a binary.
- `tokio` gains `features = ["time"]` in `fs_executor`; the workspace set is
  `["sync", "rt-multi-thread", "macros", "fs"]`.

The seam follows `PV_E2E_OS_TRASH_FAKE` in
`crates/fs/executor/src/ops/trash_op.rs`.

Verified: 25 consecutive runs of the target test pass, the binary's other 8 tests
pass, and clippy with `-D warnings` is clean on 1.95.0 and 1.98.0.

## Spec Context

Test reliability, outside any spec.

Tracks-Bead: astro-plan-ytx7
Closes-Bead: astro-plan-ytx7

Merge-Bead: astro-plan-z471

### DIFF
--- a/crates/app/core/Cargo.toml
+++ b/crates/app/core/Cargo.toml
@@ -58,6 +58,11 @@ uuid = { version = "1", features = ["serde", "v4", "v5"] }
 
 [dev-dependencies]
 app_core_cache = { path = "../cache" }
+# plan_resume_integration.rs paces the executor's forward pass so its retry lands
+# while the resumed run is still draining. The feature is off on the
+# `[dependencies]` edge above, so the seam is absent from every binary built from
+# this crate.
+fs_executor = { path = "../../fs/executor", features = ["test-pacing"] }
 tempfile = { workspace = true }
 # spec 048 T010: forces two fixture frames to a single mtime so the
 # moved-vs-catalogued `file_record` comparison tests the writer, not the clock.
--- a/crates/app/core/tests/plan_resume_integration.rs
+++ b/crates/app/core/tests/plan_resume_integration.rs
@@ -610,14 +610,32 @@ async fn resume_refused_when_plan_not_paused() {
 
 // ── resume + retry item-set agreement (review fix) ────────────────────────────
 
-/// Removes the named var on drop, including on panic unwind, so a failed
-/// assertion cannot leak executor pacing into the rest of this binary. `cargo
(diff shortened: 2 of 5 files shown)


srobroek:
@coderabbitai review

srobroek:
Both findings reduce to one point, addressed in 04ed1b69.

**Applied:** `EnvVarGuard` now captures the prior value and restores it on drop instead of unconditionally removing, so an outer runner that set `PV_TEST_ITEM_DELAY_MS` for its own reasons gets it back. Panic-unwind isolation is unchanged.

**Declined:** replacing the env seam with explicit executor configuration or test-only two-way synchronisation. The executor is spawned inside `resume_plan`, so reaching `drive_plan` from a test means threading a pacing parameter through the production resume path. That widens a production API to avoid a variable production never sets — the review labels it a heavy lift, and the cost lands in the wrong place.

On the `unsafe` question the analysis raised: the workspace is edition 2021, where `set_var`/`remove_var` are safe. CI runs `cargo nextest --profile ci`, one process per test, so the var cannot reach a sibling test there; the guard exists for plain `cargo test`, which shares a process per binary.

9/9 tests in the binary pass; clippy `-D warnings` clean.

PR #1685, opened 2026-08-20
fix: the inventory list contract documents the pagination it already supports

srobroek:
## Summary

- Spec 006 T001 copied its two contracts into `packages/contracts/schemas/` as
  build-time mirrors, which left one contract with two records and nothing holding
  them together. They came apart: `limit`, `offset`, and `hasMore` were added to
  the mirror and implemented in `crates/contracts/core`, `crates/app/core`, and
  `crates/persistence/targets`, and the spec copy never got them.
- Neither copy is canonical, and the generator resolves the tie the other way from
  what the mirror's test coverage suggests: `build-schemas.mjs:87` lists the
  mirrors first and the spec contracts second, then `:106` writes each in that
  order, so the spec copy is written last and silently overwrites the mirror's
  output for the same contract. The stale `inventory.list.d.ts` in this branch is
  what that looks like — the mirror had the pagination fields and the declarations
  did not. Porting the three fields into the spec copy makes both parse to equal
  JSON, which removes the question rather than answering it.
- `spec_contracts_match_their_packages_contracts_mirrors` holds both pairs
  together from here. It reports the JSON pointers that disagree instead of
  printing both 6 KB documents, so a failure can be acted on without diffing by
  hand.

Verified: 25/25 tests in `contract_schema_parity` pass. Removing `limit` from the
spec copy and flipping `hasMore.default` fails the new test with both pointers
named. Clippy `-D warnings` is clean on 1.95.0 and 1.98.0.

`inventory.session.review.json` was already byte-identical to its mirror and is
unchanged; it is now covered by the same test.

## Spec Context

Spec 006 contract artifact, brought back in line with what the backend
implements.

Tracks-Bead: astro-plan-9aq6
Closes-Bead: astro-plan-9aq6

Merge-Bead: astro-plan-sf3b

## Why no check caught the stale declarations

`scripts/check-generated-drift.sh:25` covers `packages/contracts/src/generated/`
in its default path set, and `just check-generated` runs it — but only on the Rust
lane. This change touches `packages/contracts/schemas/` and `specs/*/contracts/`,
neither of which sets the `rust` filter, so no job ran a drift check at all. #1683
adds the no-Rust leg that closes it. (An earlier commit message on this branch
blamed the justfile recipe's path list; that was wrong, and the squash body
supersedes it.)

### DIFF
--- a/packages/contracts/src/generated/envelope.d.ts
+++ b/packages/contracts/src/generated/envelope.d.ts
@@ -9,12 +9,7 @@
  * Transport-neutral operation envelopes, operation events, handles, and error DTOs.
  */
 export type AstroLibraryManagerContractEnvelopes =
-  | RequestEnvelope
-  | OkResponseEnvelope
-  | ErrorResponseEnvelope
-  | OperationHandle
-  | OperationEvent
-  | ContractError;
+  RequestEnvelope | OkResponseEnvelope | ErrorResponseEnvelope | OperationHandle | OperationEvent | ContractError;
 export type ContractVersion = "1.0.0";
 export type OperationName = string;
 export type RequestId = string;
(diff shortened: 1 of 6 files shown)


srobroek:
@coderabbitai review

PR #1686, opened 2026-08-20
fix: sexagesimal coordinates no longer show an impossible 60 seconds

srobroek:
## Summary

- A target's RA/Dec could display an impossible `60` seconds field — RA 15.25°
  rendered as `01h00m60s`, meaning `01h01m00s`. An exhaustive millidegree sweep
  finds 3115 RA values and 1420 Dec values affected. These are ordinary
  coordinates, not edge cases, and the reader sees a wrong sky position rather
  than an error.
- Cause is upstream: `skymath` 0.7.1 documents `format_ra`/`format_dec` as
  carry-safe, but rounds in value units, truncates each field from the result,
  and then rounds the leftover seconds remainder a second time. The defect is
  unfixed in 0.7.1, the latest published version.
- `astro_format` now decomposes the coordinate itself, rounding once to whole
  seconds in integer units and splitting the integer. A value that rounds up to
  60 seconds becomes the next minute before any field exists. `skymath` still
  owns the coordinate; the public API is unchanged.

Verified: 8/8 tests in `targeting::astro_format` pass on debug and release. The
new millidegree sweep of the display domain fails on the previous implementation
in 4535 places and runs in 3s debug. Clippy `-D warnings` clean on 1.95.0 and
1.98.0.

Follow-up outside this repo: report the defect to `skymath`.

## Spec Context

Fuzz finding from the code:tauri-commands surface sweep.

Tracks-Bead: astro-plan-3v3r.8.40
Closes-Bead: astro-plan-3v3r.8.40

Merge-Bead: astro-plan-spp1

### DIFF
--- a/crates/targeting/src/astro_format.rs
+++ b/crates/targeting/src/astro_format.rs
@@ -27,7 +27,8 @@ use crate::coords::{self, Pointing};
 const RA_SECONDS_PER_DAY: u32 = 24 * 3600;
 
 /// Arcseconds in the 90° Dec limit. [`coords::to_equatorial`] clamps Dec into
-/// `[-90, 90]`, so the rounded magnitude can reach this but not exceed it.
+/// `[-90, 90]` and `90 * 3600` is exact in binary floating point, so the rounded
+/// magnitude reaches this value but never exceeds it.
 const DEC_ARCSEC_LIMIT: u32 = 90 * 3600;
 
 /// A target's RA/Dec formatted in astronomy notation: `HHhMMmSSs` for RA,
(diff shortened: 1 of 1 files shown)


srobroek:
Closing as superseded by #1690.

Both PRs fix the same defect, `format_sexagesimal` emitting an out-of-domain `60` in the seconds field (bead `astro-plan-3v3r.8.40`), in different places. This branch reimplements sexagesimal carry locally in `crates/targeting` (+146/-36). #1690 bumps the `skymath` dependency to 0.7.2, where the carry is fixed upstream.

Sweeping the target catalogue twice measured 3115 out-of-domain RA values and 1420 out-of-domain Dec values at skymath 0.7.1, against zero at 0.7.2. Those are the cases this branch was written for, so the upstream bump covers them, and `skymath` is this project's own crate, meaning a parallel implementation here would be a second thing to keep correct with no coverage gain.

The regression test that guards the behaviour is a blocking item on #1690's review, so it is merged there rather than lost with this branch.

PR #1687, opened 2026-08-21
fix: an unreadable folder no longer stops artifact detection or reports present artifacts as missing

srobroek:
## Summary

- One unreadable folder under a project root no longer aborts the on-attach
  artifact reconciliation pass for the whole root.
- Artifacts inside a folder the app cannot read stay `present`. They are no
  longer reported as missing on the strength of a directory nobody could open.
- A skipped folder is recorded in the audit trail (`artifact.scan_incomplete`)
  so a partial scan is visible.

## What changed

`real_read_dir_into` in `apps/desktop/src-tauri/src/watcher.rs` propagated the
`read_dir` error for a subdirectory it could not read, so one permission-denied
directory anywhere under a project root aborted reconciliation for that root and
every artifact elsewhere under it went unobserved.

Skipping the unreadable directory alone is not sufficient, and on its own is
worse than the abort. `workflow_artifacts::reconcile` classifies a recorded row
whose file name is absent from the walk's file list as `ReconcileOutcome::Gone`,
and `run_attach_reconciliation` persists that through `mark_missing_batch`. A
walk that silently omits a subtree therefore reports every present artifact
under it as `missing` — a custody claim about files that are on disk and intact.

Absent and unreadable are now distinct states:

- `real_read_dir` returns a `workflow_artifacts::DirListing`: the real files
  found plus the directories whose contents could not be read
  (`crates/workflow/artifacts/src/reconciler.rs`).
- `reconcile` classifies a row that resolves inside one of those directories as
  the new `ReconcileOutcome::Unknown`. `run_attach_reconciliation` neither
  touches nor transitions an `Unknown` row.
- The skipped paths reach an audit row via the new `artifact.scan_incomplete`
  topic (`crates/audit-types/src/event_bus.rs`, published by
  `app_core::artifact::report_scan_incomplete`), alongside the existing `warn`
  log. Constitution II: a partial scan of a custody-relevant walk is recorded,
  not swallowed.

A `read_dir` failure on the project root itself still returns `Err`, which
signals root-unavailable rather than a partial subtree.

`fs_pathsafe::real_files_under` (`crates/fs/pathsafe/src/lib.rs:185`) already
skips rather than propagates, so the two walkers now agree.

## Tests

`apps/desktop/src-tauri/tests/artifact_watcher_unreadable_subdir.rs` records an
artifact under `output/`, detaches the watcher, sets `output/` to mode 000, and
re-attaches. It asserts the row is still `present`, that no `artifact.missing`
fires, and that `artifact.scan_incomplete` names the skipped directory.

`real_read_dir_skips_an_unreadable_subdirectory_and_reports_it` in `watcher.rs`
covers the walk: a mode-000 subdirectory alongside a top-level file and a nested
file, both siblings still returned, the locked path reported, `Ok` not `Err`.

`row_under_an_unreadable_directory_is_unknown_not_gone` in `reconciler.rs`
covers the classification split: a row under the unreadable directory is
`Unknown`, a row absent from a readable directory is still `Gone`.

Both filesystem tests are `unix`-only and skip themselves with a printed message
when the environment can read a mode-000 directory (running as root), where the
assertion would be vacuous.

## Verification

Run in the branch worktree with `CARGO_TARGET_DIR` set per-branch:

| Command | Result |
| --- | --- |
| `cargo test -p workflow_artifacts -p app_core_lifecycle -p audit_types -p desktop_shell` | 190 passed, 20 suites, exit 0 |
| `cargo clippy -p workflow_artifacts -p app_core_lifecycle -p audit_types -p desktop_shell --all-targets` | no issues, exit 0 |
| `cargo fmt --all` | clean |

Merge-Bead: astro-plan-gn3ix
Tracks-Bead: astro-plan-0sgb.4
Tracks-Bead: astro-plan-3v3r.8.41
Closes-Bead: astro-plan-0sgb.4
Closes-Bead: astro-plan-3v3r.8.41

### DIFF
--- a/apps/desktop/src-tauri/src/watcher.rs
+++ b/apps/desktop/src-tauri/src/watcher.rs
@@ -181,28 +181,59 @@ fn tool_id_from_project_tool(project_tool: &str) -> String {
 /// for project output folders in v1) — neither into a symlinked directory
 /// nor as a symlinked file.
 fn real_read_dir(dir: &Path) -> Result<Vec<PathBuf>, String> {
-    let mut out = Vec::new();
-    real_read_dir_into(dir, &mut out)?;
-    Ok(out)
+    let (files, skipped) = real_read_dir_reporting(dir)?;
+    if !skipped.is_empty() {
+        tracing::warn!(
+            count = skipped.len(),
+            "artifact watcher: unreadable directories skipped during reconcile: {}",
+            skipped.iter().map(|p| p.display().to_string()).collect::<Vec<_>>().join(", ")
+        );
+    }
+    Ok(files)
 }
 
-/// Recursion helper for [`real_read_dir`]; `out` accumulates real file paths
-/// across the whole subtree.
-fn real_read_dir_into(dir: &Path, out: &mut Vec<PathBuf>) -> Result<(), String> {
+/// [`real_read_dir`] plus the subdirectories that could not be read.
+///
+/// `workflow_artifacts::reconcile` takes a `&Path -> Result<Vec<PathBuf>, _>`
+/// lister, so the skipped list cannot travel with the files; [`real_read_dir`]
+/// logs it instead.
+fn real_read_dir_reporting(dir: &Path) -> Result<(Vec<PathBuf>, Vec<PathBuf>), String> {
+    let mut files = Vec::new();
+    let mut skipped = Vec::new();
     let entries =
         std::fs::read_dir(dir).map_err(|e| format!("read_dir({}) failed: {e}", dir.display()))?;
+    real_read_dir_into(entries, &mut files, &mut skipped);
+    Ok((files, skipped))
+}
+
+/// Recursion helper for [`real_read_dir_reporting`]; `files` accumulates real
+/// file paths across the whole subtree and `skipped` the directories whose
+/// `read_dir` failed.
+///
+/// An unreadable subdirectory is skipped, not propagated: one
+/// permission-denied directory must not abort reconciliation for every
+/// artifact elsewhere under the project root (matches
+/// `fs_pathsafe::real_files_under`).
+fn real_read_dir_into(
+    entries: std::fs::ReadDir,
+    files: &mut Vec<PathBuf>,
+    skipped: &mut Vec<PathBuf>,
+) {
     for entry in entries.flatten() {
         let Ok(file_type) = entry.file_type() else { continue };
         if file_type.is_symlink() {
             continue;
         }
+        let path = entry.path();
         if file_type.is_dir() {
-            real_read_dir_into(&entry.path(), out)?;
+            match std::fs::read_dir(&path) {
+                Ok(child) => real_read_dir_into(child, files, skipped),
+                Err(_) => skipped.push(path),
+            }
         } else if file_type.is_file() {
-            out.push(entry.path());
+            files.push(path);
         }
     }
-    Ok(())
 }
 
 /// Real (size, mtime) probe. Uses `symlink_metadata` and rejects symlinks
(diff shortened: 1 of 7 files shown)


srobroek (commented):
Requesting changes. One blocking finding.

## 1. BLOCKING — skipping a subtree now marks its artifacts missing

`apps/desktop/src-tauri/src/watcher.rs:377` passes `real_read_dir` to
`workflow_artifacts::reconcile`, which classifies every known row whose file
name is absent from the returned list as `ReconcileOutcome::Gone`
(`crates/workflow/artifacts/src/reconciler.rs:93-97`). The caller turns those
into `gone` and calls `app_core::artifact::mark_missing_batch`
(`watcher.rs:420`), which writes `mark_artifacts_missing` and publishes an event
(`crates/app/lifecycle/src/artifact/reconcile_batch.rs:78-82`).

Trigger: a project with recorded artifacts under a subdirectory that becomes
unreadable (permission change, or a nested mount that dropped). Before this
change the pass aborted and no state moved. After it, those artifacts are
marked missing and an event is published, even though the files are present —
the walk simply could not look. `real_read_dir` discards the one piece of
information needed to avoid that: the `skipped` list, deliberately dropped to a
`warn` (`watcher.rs:184-193`).

Required action: do not report missing under a skipped subtree. Capture
`skipped` from the lister at `watcher.rs:377` (a closure over a
`RefCell<Vec<PathBuf>>` around `real_read_dir_reporting` keeps the
`Fn(&Path) -> Result<Vec<PathBuf>, String>` signature), then either filter
`gone` entries whose path is under a skipped directory, or skip the
`mark_missing_batch` phase entirely while `skipped` is non-empty. Add a
regression test asserting a known artifact under the locked directory is not
reported gone.

## 2. Non-blocking — skipped paths reach only a tracing warn

`watcher.rs:185-192` is the sole consumer. Nothing user-facing or auditable
records that a subtree went unobserved, so a reconcile that saw half a project
is indistinguishable from a complete one at the UI and in the audit log. The
existing residual note on the node acknowledges the lister contract as the
cause; a follow-up bead for the report path would close it.

## 3. Non-blocking — pre-existing gaps the new taxonomy leaves open

`entry.file_type()` failure still `continue`s without landing in `skipped`
(`watcher.rs:220`), and `entries.flatten()` drops per-entry io errors, so those
subtrees vanish silently rather than being reported. Recursion is still
unbounded depth where `fs_pathsafe` uses an explicit stack
(`crates/fs/pathsafe/src/lib.rs:192`). Both predate this change.

## Verified

A standalone rustc replica of `real_read_dir` / `real_read_dir_reporting` /
`real_read_dir_into` at `508a854ba` run against the PR's own fixture: siblings
returned (2 files), `skipped == [locked]`, whole walk `Ok`, missing root `Err`,
unreadable root itself `Err`. The `#[cfg(unix)]` gate, the root-bypass early
return with a printed message, and the permission restore before the assertions
are all present at `watcher.rs:852-899`; the test is not vacuous in a non-root
environment. `real_files_under` is at `crates/fs/pathsafe/src/lib.rs:185` and
does use skip-and-continue, so the body's convention claim holds. The parent
walker propagates via `?` at `watcher.rs:200`, matching the body's claim that it
fails the same fixture.

## Untested

`cargo test -p desktop_shell --lib` (claimed 57 passed) and
`cargo clippy -p desktop_shell --all-targets -- -D warnings` were not
reproduced: 17Gi free on the shared volume with concurrent reviewer builds, and
the brief forbids building past a space failure.

## Prose

The title and body describe the change accurately and the numbers check out.
They omit the consequence in finding 1, which the changelog entry should not
claim away — "no longer stops artifact detection" is true, but the same commit
starts reporting present artifacts as missing.

srobroek (commented):
APPROVE round=2 head=65e3e0fedfa7fb8bf01a6b185354d282e0f56350

Round-1 FIX 1 (blocking) is resolved. `real_read_dir` now returns a
`DirListing` carrying the skipped directories (`watcher.rs:183`), `reconcile`
classifies a row resolving under one of them as `ReconcileOutcome::Unknown`
(`crates/workflow/artifacts/src/reconciler.rs:117-121`, `under_unreadable` at
`:119`), and `run_attach_reconciliation` neither adds it to `seen_ids` nor to
`gone` (`watcher.rs:387`), so `mark_missing_batch` never sees it. Round-1
FIX 2 is resolved by the `artifact.scan_incomplete` audit row
(`crates/audit-types/src/event_bus.rs:501`,
`crates/app/lifecycle/src/artifact/reconcile_batch.rs:127`).

VERIFIED by run in the review worktree at this head, `CARGO_TARGET_DIR`
isolated:

- `cargo test -p workflow_artifacts -p app_core_lifecycle -p audit_types` —
  76 passed, 6 suites, exit 0; includes
  `reconciler::tests::row_under_an_unreadable_directory_is_unknown_not_gone`.
- `cargo test -p desktop_shell --lib --test artifact_watcher_unreadable_subdir
  --test artifact_watcher_missing_reconciliation` — 59 passed, 3 suites, exit 0;
  `artifact_under_an_unreadable_subdirectory_is_not_marked_missing` ran and no
  `skipping:` line was printed, so the mode-000 root bypass at
  `tests/artifact_watcher_unreadable_subdir.rs:97` did not fire and the
  assertions are not vacuous. The file is `#![cfg(unix)]` at line 12.
- `cargo clippy -p workflow_artifacts -p app_core_lifecycle -p audit_types
  -p desktop_shell --all-targets -- -D warnings` — no issues, exit 0.

Scope: the round-2 commit touches 7 files, all inside the reviewed scope. The
third commit on the branch is the `origin/main` merge of #1689, not PR content.
`ReconcileOutcome` has exactly one consumer (`watcher.rs:380-387`), so the new
variant cannot silently no-op elsewhere. No generated contract covers audit
payloads, so the new topic introduces no schema drift.

Non-blocking:

1. `apps/desktop/src-tauri/src/watcher.rs:191` — the doc comment still points at
   `[real_read_dir_reporting]`, deleted in this commit. Broken intra-doc link
   and a dangling reference for the next reader.
2. `apps/desktop/src-tauri/src/watcher.rs:204-205` — round-1 FIX 3 is unchanged:
   `entries.flatten()` drops per-entry io errors and a failed `file_type()`
   `continue`s without landing in `skipped`, so such an entry is absent from
   both lists and a recorded row for it still reaches `Gone` -> `missing`. Same
   custody hole as FIX 1 through a rarer trigger; pre-existing, not a
   regression of this PR. Nearest open bead: astro-plan-3v3r.20.28.
3. `apps/desktop/src-tauri/src/watcher.rs:411` — `report_scan_incomplete` fires
   on every attach while the directory stays unreadable, so repeated drawer
   opens append one audit row each. Recording is correct per constitution II;
   the repetition is unbounded.

UNTESTED: that the integration test fails without the fix was not proven by
mutation run — a reviewer may not edit the tree. By construction it cannot pass
pre-fix: `ReconcileOutcome::Unknown` and `TOPIC_ARTIFACT_SCAN_INCOMPLETE` do not
exist at 508a854ba, and the sibling
`artifact_watcher_missing_reconciliation` suite (passing) demonstrates that a
row absent from the walk's file list is persisted `missing`.

PROSE: the PR body matches the diff, including the widened scope, and does not
claim FIX 3 away.

PR #1688, opened 2026-08-21
fix: correct the documented SQLite durability level for the local database

srobroek:
`crates/persistence/core/tests/wal_journal_mode.rs:16` said `synchronous` was
left at SQLite's default `FULL`. `crates/persistence/core/src/lib.rs:113` sets
`SqliteSynchronous::Normal`. This corrects the test module doc and moves the
pragma assertion onto a file-backed database.

Merge-Bead: astro-plan-r0tgj
Work-Bead: astro-plan-q18u

## What changed

| File | Change |
| --- | --- |
| `crates/persistence/core/tests/wal_journal_mode.rs` | Module doc now states the level the pool sets. New test `database_connect_reports_synchronous_normal` asserts `PRAGMA synchronous == 1` on a file-backed database. |
| `crates/persistence/core/src/lib.rs` | The `Database::connect` docstring deferred Tier-1 escalation to bead kyo7.49 as pending. kyo7.49 is closed and merged (#1586). The docstring now names the landed mechanism, scopes it to filesystem-mutation intents, and states that a failed checkpoint is discarded without logging. Deletes `synchronous_normal_on_fresh_connection`, superseded by the file-backed test. |
| `crates/persistence/plans/src/repositories/plan_apply.rs` | `checkpoint_intent`'s docstring said "Callers log and proceed". Both callers discard the error without logging; the docstring now says so. |

No pragma, pool option, or call site changed. The pool-wide default remains
`synchronous = NORMAL` (`src/lib.rs:113`).

## Statements audited

| Location | Claim | Verdict |
| --- | --- | --- |
| `crates/persistence/core/src/lib.rs:86` | NORMAL | True: matches `:113` |
| `crates/persistence/core/tests/wal_journal_mode.rs:16` | "left at SQLite's default `FULL`" | False; corrected here |
| `specs/062-session-heterogeneity/data-model.md:1141` | `PRAGMA synchronous = FULL` | False; outside this branch's scope, filed as astro-plan-rar3 |

`CLAUDE.md` Principle V and its 1.1.0 amendment log adopt relaxed global
durability "to enable batched writes and synchronous=NORMAL SQLite mode", with
custody guarantees concentrated on Tier-1 records. NORMAL is therefore the
correct level, and the two `FULL` claims are the wrong ones.

## Test coverage

`synchronous_normal_on_fresh_connection` asserted the pragma on an in-memory
database, where the setting is inert. It detected removal of
`.synchronous(SqliteSynchronous::Normal)` just as well as a file-backed
assertion does: sqlx emits no `synchronous` pragma unless one is set, so SQLite
reports its compiled default `FULL` (2) either way. The file-backed test is
therefore not stronger evidence, only better scoped: it observes the level
where it has durability consequences. Being a near-duplicate of the in-memory
one, it replaces it rather than joining it.

Mutation check, by deleting `.synchronous(SqliteSynchronous::Normal)` and
re-running each test:

| Test | Result with the option removed |
| --- | --- |
| `synchronous_normal_on_fresh_connection` (in-memory, now deleted) | FAILED, `left: 2` |
| `database_connect_reports_synchronous_normal` (file-backed) | FAILED, `left: 2` |

Neither test observes durability; both read the pragma back.

## Tier-1 durability

One Tier-1 escalation already exists, on one path. `checkpoint_intent`
(`crates/persistence/plans/src/repositories/plan_apply.rs:38`) issues
`PRAGMA wal_checkpoint(TRUNCATE)` after the filesystem-mutation intent commit,
called at `:169` and `:319`. It landed with kyo7.49 in #1586. It is best-effort:
both call sites discard the result with `let _ =` and do not log it, and a
failed checkpoint is not fatal because boot reconciliation is the backstop
(`plan_apply.rs:34-37`).

Making that failure visible would require a logging facade in the persistence
layer, which has none: `rg 'tracing::' crates/persistence` returns nothing, and
neither `persistence_core` nor `persistence_plans` depends on `tracing` or
`log`. This PR corrects the two docstrings that claimed otherwise instead of
adding that dependency.

Tier-1 record classes with no paired filesystem action (attribution, lifecycle
choices, user overrides) may or may not need the same escalation. That reading of
Principle V is the repo owner's call, not a code question, so it is parked on
astro-plan-53b8 with four options costed. Their write sites sit outside this
branch's scope and this PR does not audit them.

## Verification

| Command | Exit |
| --- | --- |
| `cargo fmt --all --check` | 0 |
| `cargo test -p persistence_core` | 0 (51 tests) |
| `cargo clippy -p persistence_core --all-targets -- -D warnings` | 0 |
| `cargo clippy -p persistence_plans -- -D warnings` | 0 |
| `cargo test -p persistence_plans --features persistence_core/test-fixture` | 0 (67 tests) |

`cargo test -p persistence_plans` without that feature flag fails to compile at
this branch's base as well: the crate's dev-dependency on `persistence_core`
declares `features = []` while its tests import `persistence_core::test_support`,
which is gated behind `test-fixture`. CI does not hit it because it builds
`--workspace`. Filed as astro-plan-8b0d5; not addressed here.

### DIFF
--- a/crates/persistence/core/src/lib.rs
+++ b/crates/persistence/core/src/lib.rs
@@ -88,9 +88,11 @@ impl Database {
     ///   at most losing the last un-checkpointed transaction — the database is
     ///   never corrupted.  This is materially faster than `FULL` on spinning or
     ///   network-backed storage without sacrificing integrity guarantees the app
-    ///   depends on.  Tier-1 (`FULL`) escalation per-connection for high-value
-    ///   writes (audit events, filesystem plan commits) is tracked as a follow-up
-    ///   in kyo7.49.
+    ///   depends on.  Tier-1 records are not covered by the pool default: the
+    ///   filesystem-mutation intent path escalates per commit with
+    ///   `PRAGMA wal_checkpoint(TRUNCATE)` (see `checkpoint_intent` in
+    ///   `persistence_plans::repositories::plan_apply`) rather than raising
+    ///   `synchronous` for every writer.
     /// - **`foreign_keys = ON`**: enforced explicitly so the constraint is not
     ///   silently absent if a future sqlx version changes its default.
     /// - **`busy_timeout`** (#1231): pins the wait interval so it is ours, not
--- a/crates/persistence/core/src/lib.rs
+++ b/crates/persistence/core/src/lib.rs
@@ -88,11 +88,15 @@ impl Database {
     ///   at most losing the last un-checkpointed transaction — the database is
     ///   never corrupted.  This is materially faster than `FULL` on spinning or
     ///   network-backed storage without sacrificing integrity guarantees the app
-    ///   depends on.  Tier-1 records are not covered by the pool default: the
-    ///   filesystem-mutation intent path escalates per commit with
+    ///   depends on.  No writer raises `synchronous`; this pool default is the
+    ///   only setting in the repo.  One tier-1 path syncs separately: the
+    ///   filesystem-mutation intent commit is followed by a best-effort
     ///   `PRAGMA wal_checkpoint(TRUNCATE)` (see `checkpoint_intent` in
-    ///   `persistence_plans::repositories::plan_apply`) rather than raising
-    ///   `synchronous` for every writer.
+    ///   `persistence_plans::repositories::plan_apply`), whose failure both
+    ///   call sites log and ignore.  The other tier-1 records §V names —
+    ///   frame-to-session attribution, lifecycle choices, user overrides — have
+    ///   no such escalation; whether they require one is open
+    ///   (astro-plan-53b8).
     /// - **`foreign_keys = ON`**: enforced explicitly so the constraint is not
     ///   silently absent if a future sqlx version changes its default.
     /// - **`busy_timeout`** (#1231): pins the wait interval so it is ours, not
@@ -308,15 +312,6 @@ mod tests {
         let fk_on: i64 =
             sqlx::query_scalar("PRAGMA foreign_keys").fetch_one(db.pool()).await.unwrap();
         assert_eq!(fk_on, 1, "PRAGMA foreign_keys must be 1 (ON) on every connection");
-    }
-
-    #[tokio::test]
-    async fn synchronous_normal_on_fresh_connection() {
-        let db = super::Database::in_memory().await.expect("in-memory connect");
-        // SQLite returns the numeric value: 0=OFF, 1=NORMAL, 2=FULL, 3=EXTRA.
-        let sync_level: i64 =
-            sqlx::query_scalar("PRAGMA synchronous").fetch_one(db.pool()).await.unwrap();
-        assert_eq!(sync_level, 1, "synchronous must be NORMAL (1) per tier-2 policy");
     }
 
     #[tokio::test]
--- a/crates/persistence/core/src/lib.rs
+++ b/crates/persistence/core/src/lib.rs
@@ -93,7 +93,9 @@ impl Database {
     ///   filesystem-mutation intent commit is followed by a best-effort
     ///   `PRAGMA wal_checkpoint(TRUNCATE)` (see `checkpoint_intent` in
     ///   `persistence_plans::repositories::plan_apply`), whose failure both
-    ///   call sites log and ignore.  The other tier-1 records §V names —
+    ///   call sites discard without logging — the persistence layer has no
+    ///   logging facade, and boot reconciliation is the backstop.  The other
+    ///   tier-1 records §V names —
     ///   frame-to-session attribution, lifecycle choices, user overrides — have
     ///   no such escalation; whether they require one is open
     ///   (astro-plan-53b8).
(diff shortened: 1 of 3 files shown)


srobroek (commented):
REQUEST_CHANGES — dimension doc-correctness, round 1, head be8da82.

Verified by running: `cargo test -p persistence_core` → 52 passed, exit 0; both `tests::synchronous_normal_on_fresh_connection` and `database_connect_reports_synchronous_normal` pass. The PR body's verification table matches.

Verified by reading: `checkpoint_intent` exists at `crates/persistence/plans/src/repositories/plan_apply.rs:37`, called at `:168` and `:318`; kyo7.49 is closed and #1586 is merged; the only remaining `synchronous = FULL` claim in the repo is `specs/062-session-heterogeneity/data-model.md:1141`, deferred to astro-plan-rar3, which names that file.

## FIX

1. `crates/persistence/core/src/lib.rs:91-95` — blocking. "Tier-1 records are not covered by the pool default: the filesystem-mutation intent path escalates per commit with `PRAGMA wal_checkpoint(TRUNCATE)`" reads as the whole Tier-1 mechanism. Constitution V Tier 1 also names frame-to-session attribution, lifecycle choices, and user overrides; no write path for those issues a checkpoint — `checkpoint_intent` has exactly two callers, both in the plan-apply intent path. The PR body concedes this and parks it on astro-plan-53b8, so the docstring, read alone, asserts coverage the code does not provide. Required: scope the sentence to filesystem-mutation intents and name the open question.

2. PR body, section "Existing coverage was weaker than it looked" — blocking. The claim is not supported. `Database::in_memory` delegates to `Database::connect` (`src/lib.rs:134`), and sqlx emits no `synchronous` pragma unless one is set (`sqlx-sqlite/src/options/mod.rs:189-191`), so SQLite's compiled `FULL` (2) would be reported if `.synchronous(Normal)` (`src/lib.rs:107`) were dropped — and the pre-existing in-memory assertion at `src/lib.rs:314` would fail exactly as the new one does. The old test's weakness is semantic (the setting is inert in memory), not evidential; the new test adds no detection power, and it does not observe durability either — it reads the same pragma. Required: state the true, narrower claim.

3. `crates/persistence/core/tests/wal_journal_mode.rs:174-189` — non-blocking, follows from FIX 2. This is a near-duplicate of `src/lib.rs:313-320`: same pragma, same expected value, same construction path. Required: re-point the existing assertion at a file-backed database and drop one of the two, rather than keeping both.

4. `crates/persistence/core/src/lib.rs:92` — non-blocking. "escalates per commit" omits that the escalation is best-effort: both call sites discard the result (`plan_apply.rs:168`, `:318`) and that function's own docstring (`plan_apply.rs:34-36`) states a failed checkpoint is not fatal. Required: say best-effort.

5. PR body — non-blocking nit. It cites `src/lib.rs:311` for `synchronous_normal_on_fresh_connection`; the function is at `:314` at this head (`:312` at base).

The corrected statements themselves are accurate: `synchronous = NORMAL` matches `src/lib.rs:107`, and the removal of "without weakening durability" from the test module doc is right, since NORMAL does trade the last committed transactions on power loss.

srobroek:
REQUEST_CHANGES — dimension doc-correctness, round 2, head 154dfb3.

Verified by running (`CARGO_TARGET_DIR=/Users/sjors/tmp/cargo-target/verify-1688`):

- `cargo test -p persistence_core` → 51 passed, exit 0; `database_connect_reports_synchronous_normal ... ok`. Matches the body's stated count.
- Mutation: deleting `.synchronous(SqliteSynchronous::Normal)` from `crates/persistence/core/src/lib.rs:111` makes the surviving test FAIL at `tests/wal_journal_mode.rs:188` — `left: 2, right: 1`. Reverted; 51 passed again, tree clean at 154dfb3. The kept assertion has real detection power, and the body's round-1 claim about the deleted in-memory test is now stated correctly.
- `cargo fmt --check` → exit 0; `cargo clippy -p persistence_core --all-targets -- -D warnings` → exit 0.

Verified by reading: `.synchronous(` appears once in the workspace (`src/lib.rs:111`), so the docstring's "this pool default is the only setting in the repo" holds. No pragma, pool option, or call site changed. astro-plan-53b8 and astro-plan-rar3 both exist and name the deferred loci; the only remaining `FULL` claim is `specs/062-session-heterogeneity/data-model.md:1141`. Round-1 FIX 1, 2, 3, 4, and 5 are all addressed.

## FIX

1. `crates/persistence/core/src/lib.rs:95-96` — blocking. "whose failure both call sites log and ignore" is false. Neither call site logs: `plan_apply.rs:168` and `:318` are bare `let _ = checkpoint_intent(...).await;`, and the comment at `plan_apply.rs:165-166` states the error "is intentionally swallowed (the persistence layer carries no logging facade)". This PR's own body says the same thing correctly ("both call sites discard the result with `let _ =`"), so only the docstring is wrong. A doc-correction commit that lands a new false claim in the same paragraph repeats the defect it fixes. Required: say the failure is discarded, not logged.

2. `crates/persistence/plans/src/repositories/plan_apply.rs:36` — non-blocking, out of this diff. "Callers log and proceed" is the pre-existing source of FIX 1 and is contradicted by `:165-166` in the same file. Required: file it as a follow-up so the corrected docstring does not diverge from the text it points at.

PR #1690, opened 2026-08-21
fix: target coordinates no longer display 60 in the seconds field

srobroek:
## Summary

- Raises `skymath` from 0.7.1 to 0.7.2 across all seven workspace manifests and
  `Cargo.lock`.
- Target coordinate displays keep the seconds field in domain.
  `sexagesimal(15.248, -89.85)` returned RA `01h00m60s` and Dec `−89°50′60″` at
  0.7.1; it returns `01h01m00s` and `−89°51′00″` at 0.7.2.
- The defect affected 3115 RA and 1420 Dec millidegree values. skymath rounded in
  value units and then re-rounded the seconds remainder in its format string;
  `crates/targeting/src/astro_format.rs:51-52` delegates carry handling to skymath,
  so platevault printed the bad field directly. Fixed upstream by
  `nightwatch-astro/skymath#35`, released as `v0.7.2`.
- `seconds_carry_never_shows_60` in `crates/targeting/src/astro_format.rs` pins
  the three inputs skymath 0.7.1 printed with a `60` field and asserts the
  carried strings. A downgrade or lock refresh that reinstates value-unit
  rounding fails there.
- The module doc in that file names 0.7.2 as the version its carry-safety claim
  depends on.

## Test plan

`cargo test -p targeting -p metadata_core -p sessions -p calibration_core -p
app_core_inbox -p targeting_resolver`: 598 passed and 1 ignored across 17 suites.
`cargo fmt -p targeting -- --check` and `cargo clippy -p targeting --all-targets`
are clean.

## Supersedes #1686

PR #1686 addresses the same defect by decomposing locally in
`crates/targeting/src/astro_format.rs`. skymath 0.7.2 covers every case that
branch covers: both millidegree sweeps report 3115 RA and 1420 Dec out-of-domain
values at 0.7.1 and zero at 0.7.2, and #1686's zero-padding, U+2212 sign, and
pole clamp all hold through the existing wrapper. The review ruling on the
collision is SUPERSEDED, so #1686 closes and `astro-plan-3v3r.8.40` re-points to
this branch.

## Scope

All seven manifests move together so the workspace states one minimum for the
crate. The caret requirement already admitted 0.7.2, so the pins record the
version the fix needs rather than changing resolution by themselves.

`Cargo.lock` holds skymath 0.6.0 under four crates: `fits-header` 0.4.2,
`xisf-header` 0.4.3, `simbad-resolver` 0.5.0, and `target-match` 0.5.1. The
`simbad-resolver` edge is a dev-dependency, so it is absent from the build
graph. None of the four calls `ra_sexagesimal`,
`dec_sexagesimal`, `format_ra`, or `format_dec` in its published source, so no
platevault display path reaches sexagesimal formatting through that pin.

Tracks-Bead: astro-plan-h6w2
Merge-Bead: astro-plan-hcs8o

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -211,7 +211,7 @@ dependencies = [
  "sessions",
  "sha2 0.10.9",
  "simbad-resolver",
- "skymath 0.7.1",
+ "skymath 0.7.2",
  "sqlx",
  "target-match",
  "targeting",
@@ -856,7 +856,7 @@ version = "0.1.0"
 dependencies = [
  "serde",
  "serde_json",
- "skymath 0.7.1",
+ "skymath 0.7.2",
  "time",
 ]
 
@@ -1431,7 +1431,7 @@ dependencies = [
  "serde",
  "serde_json",
  "simbad-resolver",
- "skymath 0.7.1",
+ "skymath 0.7.2",
  "specta",
  "specta-typescript",
  "sqlx",
@@ -3443,7 +3443,7 @@ name = "metadata_core"
 version = "0.1.0"
 dependencies = [
  "serde",
- "skymath 0.7.1",
+ "skymath 0.7.2",
  "thiserror 2.0.19",
  "toml 0.9.12+spec-1.1.0",
 ]
@@ -5680,7 +5680,7 @@ name = "sessions"
 version = "0.1.0"
 dependencies = [
  "domain_core",
- "skymath 0.7.1",
+ "skymath 0.7.2",
  "thiserror 2.0.19",
  "time",
  "uuid",
@@ -5822,9 +5822,9 @@ dependencies = [
 
 [[package]]
 name = "skymath"
-version = "0.7.1"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fe8c5c80aca9d4212b3b0a5f743d8e9f5c945e221136f346c5b141149e68660"
+checksum = "f8778103b2c0df97997a13aaf7ca04ba98e406cdd1bbab48bbe364a0afeedd78"
 dependencies = [
  "thiserror 2.0.19",
  "time",
@@ -6443,7 +6443,7 @@ name = "targeting"
 version = "0.1.0"
 dependencies = [
  "simbad-resolver",
- "skymath 0.7.1",
+ "skymath 0.7.2",
  "target-match",
  "uuid",
 ]
@@ -6459,7 +6459,7 @@ dependencies = [
  "serde",
  "serde_json",
  "simbad-resolver",
- "skymath 0.7.1",
+ "skymath 0.7.2",
  "sqlx",
  "targeting",
  "tempfile",
--- a/apps/desktop/src-tauri/Cargo.toml
+++ b/apps/desktop/src-tauri/Cargo.toml
@@ -29,7 +29,7 @@ targeting = { path = "../../../crates/targeting" }
 # target.moon_opposition.batch (#634): geocentric Sun/Moon position
 # (sun_position/moon_position — not re-exported by `targeting`, which only
 # re-exports the coordinate primitives).
-skymath = "0.7.1"
+skymath = "0.7.2"
 contracts_core = { path = "../../../crates/contracts/core" }
 domain_core = { path = "../../../crates/domain/core" }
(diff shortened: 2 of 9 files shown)


srobroek:
REQUEST_CHANGES — dimension-2 ruling: `SUPERSEDED`.

skymath 0.7.2 covers every case PR #1686 covers. Verified by running both versions
against #1686's own assertions (`CARGO_TARGET_DIR=/Users/sjors/tmp/cargo-target/review-1690`,
throwaway crate outside the repo, wrapper logic replicated from
`crates/targeting/src/astro_format.rs:44-48`):

| case | 0.7.1 | 0.7.2 |
| --- | --- | --- |
| millidegree RA sweep, 0..360 000 | 3115 out-of-domain | 0 |
| millidegree Dec sweep, ±90 000 | 1420 out-of-domain | 0 |
| `sexagesimal(15.248, -89.85)` | `01:00:60` / `-89:50:60` | `01:01:00` / `-89:51:00` |
| RA 15.25, 16.0 | `01:00:60`, `01:03:60` | `01:01:00`, `01:04:00` |
| RA carry off the top (359.99999) | `00:00:00` | `00:00:00` |
| `-1e308` both axes | `04:15:60` (per #1686) | `04:16:00` / `-90:00:00` |
| Dec `-0.0` | `+00:00:00` | `+00:00:00` |
| Dec clamp at ±90.0001 | — | `±90:00:00` |

#1686 fixes nothing 0.7.2 misses. Its zero-padding, U+2212 sign, and pole clamp all
already hold through the existing wrapper on 0.7.2. `cargo test -p targeting
-p metadata_core -p sessions -p calibration_core -p app_core_inbox -p targeting_resolver`:
597 passed, 0 failed, 1 ignored, 17 suites — matches the body exactly.

## Blocking

1. `crates/targeting/src/astro_format.rs:71-104` — no committed regression test. The
   body defers the guard to #1686, but #1686 is superseded and will close, so no guard
   lands anywhere. Trigger: any future `skymath` bump or lock refresh reintroducing
   value-unit rounding prints `60` again with a green suite. Required: add a test to
   this module asserting the seconds field is in domain for the known-bad inputs —
   `sexagesimal(15.248, -89.85)` is `01h01m00s` / `−89°51′00″`, `sexagesimal(16.0, 0.0).ra`
   is `01h04m00s` — reusing #1686's field-domain helper if a sweep is wanted.
2. PR body, Scope paragraph — the transitive enumeration is incomplete and the defect
   claim is unsupported. `Cargo.lock` pins skymath 0.6.0 through four crates, not three:
   `fits-header` 0.4.2, `xisf-header` 0.4.3, `simbad-resolver` 0.5.0, `target-match` 0.5.1.
   And none of the four calls `ra_sexagesimal`/`dec_sexagesimal`/`format_ra`/`format_dec`
   anywhere in its `src` (ripgrep, all four vendored sources), so no platevault path
   reaches sexagesimal formatting through them and the 0.6.0 types cannot cross into the
   0.7.2 formatter. Required: name the fourth crate and say the 0.6.0 pin is unreachable
   from platevault display rather than implying live exposure.

## Non-blocking

3. PR body, Overlap section — "`astro-plan-3v3r.8.40` stays assigned to #1686" and
   "whoever merges second owns the reconciliation" become false once #1686 closes under
   this ruling. This text is the squash changelog entry; restate it as #1686 superseded
   and the bead re-pointed here.
4. `crates/targeting/src/astro_format.rs:4-17` — the module doc asserts skymath's rounding
   is carry-safe. True at 0.7.2, false at 0.7.1, and the doc names no minimum version, so
   the claim reads as version-independent. State that it holds from 0.7.2.
5. Bead `astro-plan-3v3r.10.20` is stale at this head: only one formatter exists.
   `apps/desktop/src-tauri/src/commands/target_lookup.rs:278` already delegates to
   `targeting::astro_format::sexagesimal`, and
   `apps/desktop/src/features/targets/TargetDetail.tsx:102` records the hand-rolled TS
   `fmtRa`/`fmtDec` as already deleted. Nothing to converge; close 3v3r.10.20 as resolved.

Verified by running: the version table, the test counts, the seven bumped manifests, the
lock's four 0.6.0 dependents, the absence of formatting calls in all four vendored crates,
and the single-formatter finding. Concluded by reading: the type-identity argument for
0.6.0 unreachability (`untested` — no compile-failure experiment was constructed).

PR #1691, opened 2026-08-21
test(fs-executor): gate the pacing seam on an ordering instead of a sleep

srobroek:
## What changed

The `test-pacing` seam in `fs_executor` parks the forward pass at each
non-terminal item boundary and publishes the arrival, so a test drives the run
step by step. It replaces a per-item sleep.

`resume_then_retry_of_pre_pause_failed_item_reaches_terminal_state` now asserts
the ordering it depends on:

1. Wait for the resumed pass to park at its first item boundary, then assert the
   plan is still `applying`.
2. File the retry; assert item 0 left `failed` for `applying` and that its source
   file has not moved.
3. Release exactly one item and wait for the next arrival, which is strictly
   after the post-item retry drain; assert item 0's move happened.
4. Drop the gate, releasing the rest of the run.

The seam is a `tokio::sync::Semaphore` pair behind the default-off `test-pacing`
feature: arrivals accumulate as permits the test consumes, releases are permits
the loop consumes, and `close()` on the guard's `Drop` releases every remaining
boundary, including on a panic unwind. Arrival waits are bounded at 30s and panic
on expiry, which turns a broken ordering into a reported failure rather than a
hang; the bound does not establish the ordering.

The gate registry is keyed by plan id. Integration tests share one process, so a
single process-global gate let a concurrent test's executor park on this gate's
`release` semaphore and publish arrivals into its `arrived` semaphore. `gate_item`
takes `&item.plan_id` and returns without touching a semaphore when that plan has
no installed gate. `ItemGate::install` panics on a duplicate plan id rather than
replacing a live gate.

Plan id rather than the alternatives: `RetryQueue` and the executor instance are
both constructed inside `app_core`, so a test cannot name either one; a tokio
task-local does not reach the executor, which runs in a task the test did not
spawn; threading a handle through `execute_plan` would put a test-only parameter
in the production signature.

The production path is unchanged: the module, its call site, and the
`tokio/time` dependency are all gated on `test-pacing`, which is reachable only
through a `[dev-dependencies]` edge.

## EnvVarGuard

The bead also asked for an `EnvVarGuard` dedupe. At HEAD the workspace holds
exactly one definition (`crates/app/core/src/plan_apply/tests.rs:1089`), and no
other env-restoring `Drop` impl exists, so no copies remain to delete.

## Verification

Every run below used `CARGO_TARGET_DIR=/Users/sjors/tmp/cargo-target/fix-0sgb1-r2`
at `4e7a38148`.

| Command | Result |
| --- | --- |
| `cargo test -p app_core --test plan_resume_integration`, 6 consecutive runs, no name filter | 6 pass, 0 fail; 9 tests per run |
| `cargo test -p app_core` | exit 0, 536 passed, 0 failed |
| `cargo test -p fs_executor` | exit 0, 123 passed across 4 suites |
| `cargo clippy -p fs_executor --all-targets -- -D warnings` | exit 0 |
| `cargo clippy -p fs_executor --all-targets --features test-pacing -- -D warnings` | exit 0 |
| `cargo clippy -p app_core --all-targets -- -D warnings` | exit 0 |
| `cargo fmt --check` | exit 0 |

The full-binary run with no name filter is the measurement that matters here: an
`--exact` run executes the gating test alone and cannot observe cross-test permit
interference.

Inverting the ordering proves the gate is load-bearing. Releasing the whole run
and waiting for the plan to reach a terminal state before filing the retry fails
the test at `plan_resume_integration.rs:705` with `PlanNotInApply`. That edit was
reverted before the commit.

Merge-Bead: astro-plan-vp8w6
Tracks-Bead: astro-plan-0sgb.1
Closes-Bead: astro-plan-0sgb.1

### DIFF
--- a/crates/app/core/tests/plan_resume_integration.rs
+++ b/crates/app/core/tests/plan_resume_integration.rs
@@ -625,17 +625,17 @@ async fn resume_refused_when_plan_not_paused() {
 /// files + the real spawned executor, no seeding shortcuts): pause on a
 /// stale item, resolve the staleness, resume, then retry the pre-pause-failed
 /// item while the resumed run is still draining its other pending items.
-/// `fs_executor::run::pacing` paces the resumed executor's forward pass so the
-/// retry lands while that run is still draining. Item count alone does not:
-/// `retry_plan_item` is refused once the run closes its retry queue, and 30
-/// small same-directory moves against a warm pool can finish inside this
-/// test's three awaits (`astro-plan-ytx7`).
+///
+/// `fs_executor::run::pacing` parks the resumed forward pass at its first item
+/// boundary, and this test files the retry only after observing that arrival, so
+/// "the run is still draining" is an ordering it asserts rather than a duration
+/// it waits out. Item count alone does not establish it: `retry_plan_item` is
+/// refused once the run closes its retry queue, and 30 small same-directory
+/// moves against a warm pool can finish inside this test's three awaits
+/// (`astro-plan-ytx7`).
 #[tokio::test]
 async fn resume_then_retry_of_pre_pause_failed_item_reaches_terminal_state() {
     const OTHER_PENDING_ITEMS: usize = 30;
-    // 30 items at 20ms bounds the resumed pass at ~600ms, which is longer than
-    // the retry call by orders of magnitude on any runner.
-    const ITEM_DELAY_MS: u64 = 20;
 
     let (db, _repo, bus) = support::setup().await;
     let plan_id = Uuid::new_v4().to_string();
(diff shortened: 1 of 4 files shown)


srobroek:
REQUEST_CHANGES

The gate replaces a duration race with a cross-test permit race. On this
machine the full test binary fails 2 of 6 runs at head and 6 of 6 pass at the
base, so the change makes the target test less reliable than the sleep it
removes.

Evidence (`CARGO_TARGET_DIR=/Users/sjors/tmp/cargo-target/review-1691`):

| Invocation | head 86e3f89e | base 0adcdf76 |
| --- | --- | --- |
| `cargo test -p app_core --test plan_resume_integration` | 4 pass, 2 fail of 6 | 6 pass of 6 |
| same test with `-- --exact resume_then_retry_of_pre_pause_failed_item_reaches_terminal_state` | 20 pass of 20 | untested |

Both head failures: `plan_resume_integration.rs:716`, "the mid-run drain must
have re-executed item 0's move".

## FIX

1. **BLOCKING** — `crates/fs/executor/src/run/loop_.rs:232` (`GATE` static) and
   `loop_.rs:238` (`gate_item`) — the gate is process-global and matches no run,
   while `crates/app/core/tests/plan_resume_integration.rs` holds 12
   `apply_plan`/`resume_plan` call sites across 9 tests that cargo runs
   concurrently in one binary. Any concurrent executor that reaches an item
   boundary inside the gate window adds an `arrived` permit and consumes a
   `release` permit, so `wait_for_arrival` can return on a foreign run's arrival
   and `release(1)` can advance a foreign run instead of the gated one. That is
   the observed failure at `plan_resume_integration.rs:716`: the assertion runs
   before the target run re-executed item 0. Required action: scope the gate to
   the run it targets — carry the run or plan id to the boundary and park only a
   matching run — or serialize the executor-spawning tests in that binary.
   Isolating the test with `--exact` is not sufficient, because the default
   invocation is the whole binary.

2. **BLOCKING** — PR body, Verification table — the row
   `cargo test -p app_core --test plan_resume_integration` / "exit 0, 9/9 in
   2.25s" does not reproduce, and "retry test, 20 consecutive runs 20/20" holds
   only under `--exact` isolation. The body becomes the squash changelog entry.
   Required action: correct both rows and state the exact invocation each count
   came from.

3. **Non-blocking** — `loop_.rs:259` (`ItemGate::install`) — installing over a
   live gate overwrites the slot silently, and the displaced guard's `Drop`
   then nulls the new gate's slot. The docstring asserts "at most one gate per
   test binary at a time" with nothing enforcing it. Required action: make
   `install` panic or debug-assert when the slot is occupied.

4. **Non-blocking** — `loop_.rs:81` — the gate parks before the cancellation
   check at `loop_.rs:86`, so a run cancelled while another test's gate is
   installed stays parked until that gate drops. Required action: fold into the
   scoping fix in item 1.

## Verified clean

- Production reachability: `test-pacing` is enabled only from
  `crates/app/core/Cargo.toml:65` under `[dev-dependencies]` (line 59), and
  `cargo tree -e features -p fs_executor --no-dev-dependencies` reports no
  `test-pacing`. No release path reaches the semaphore.
- `EnvVarGuard`: one definition at head,
  `crates/app/core/src/plan_apply/tests.rs:1089`; `git grep` finds no other
  `set_var`/`remove_var` `Drop` guard in `crates`, `apps`, or `tests`. Nothing
  to dedupe, as the body states.
- The test's assertions are strengthened, not weakened:
  `plan_resume_integration.rs:695`-`717` add four assertions the sleep version
  did not make.
- Ordering claim holds by reading: `drain_retries` runs at the end of every
  iteration (`loop_.rs:118`) before the next boundary, so the post-release
  arrival is strictly after item 0's re-execution.
- Branch hygiene: one commit, no merge, no stray files.

## Untested

- The inverted-order probe at `plan_resume_integration.rs:697` was not
  reproduced.
- `cargo test -p app_core` 536/536 was not run.

PR #1693, opened 2026-08-21
fix: gh pr create was rejected for a correctly linked merge bead

srobroek:
## Summary

- `gh pr create` was unsatisfiable in a beads workspace: the PR guard compared
  the `tracks_beads` and `closes_beads` metadata of the merge bead against sets
  built from the raw value, and `bd update --set-metadata` stores every value as
  a string, so a list read back as the JSON text `["id"]` and never matched the
  PR body trailers.
- A string value is now decoded as JSON before comparison. A value that does not
  parse, or that parses to anything other than a list of ids, is still a
  mismatch, so the guard stays strict.
- Tests cover the JSON-encoded string form that `bd` actually produces, plus the
  byte-identity of the `.claude` and `.codex` copies:
  `python3 .claude/hooks/steering-git-workflow/scripts/test_pr_create_guard.py`.

Tracks-Bead: astro-plan-rhahd
Merge-Bead: astro-plan-kutsz


srobroek:
Verification review (round 1) — CHANGES REQUESTED (posted as a comment; GitHub blocks a formal review on one's own PR).

Verified: the new suite fails on the pre-fix script (1 failure + 7 errors) and passes at head (11 tests); `DeployedCopiesTests` pins the .codex copy byte-for-byte. Fix and regression coverage are correct.

FIX 1. `.claude/hooks/steering-git-workflow/scripts/__pycache__/pr-create-guard.cpython-314.pyc` - this PR commits a compiled bytecode artifact, the only tracked `__pycache__` entry in the repo. Because it is tracked, running the PR's own test suite rewrites it and leaves the tree dirty (reproduced: ` M ...pr-create-guard.cpython-314.pyc` plus untracked `test_pr_create_guard.cpython-314.pyc` after `python3 -m unittest test_pr_create_guard`), breaking clean-tree checks for later agents verifying this hook. Remove it from the commit and add a `__pycache__/` ignore rule so the new test file cannot re-introduce it.

Note (not blocking): no CI job runs these hook tests, matching the existing convention for the other 15 `test_*.py` files under `.agents/skills`.

srobroek:
Verification review round 2: approve.

FIX1 from round 1 resolved: no tracked bytecode (git ls-files has no __pycache__ or .pyc entries) and .claude/hooks/steering-git-workflow/.gitignore:1 keeps the tree clean after running the suite.

Regression evidence re-derived independently, not taken from the report:

- test_pr_create_guard.py: 11 tests OK at head 130f5d891.
- Loading the base-ref pr-create-guard.py (d3d4c10) in-memory against the same test case makes test_json_encoded_tracks_beads_string_is_accepted fail with the tracked/closing-metadata mismatch; the other validate cases still pass. The test pins the fix, not the harness.
- The regression cases use the JSON-text form that bd update --set-metadata actually writes, and test_bare_id_string_is_rejected covers the unparsable-string branch.
- test_claude_and_codex_copies_are_identical prevents the two deployed copies from drifting.
- gh pr checks 1693: 12 passed, 0 failed.

PR #1694, opened 2026-08-21
fix: the targets table's star and designation columns stay pinned while scrolling

srobroek:
## Summary

- The targets table pins its star and designation columns while the rest of the
  row scrolls horizontally. Since #1668 only the headers stayed put, so the body
  cells scrolled away from them. Both columns pin again.
- Added an end-to-end regression test for the pinned pair.

## Cause

The vanilla-extract port kept `left` on the two first-column header selectors but
dropped the matching `.pv-targets-table__row > td` rules that
`apps/desktop/src/styles/components/targets.css:101-111` carried before deletion.
`position: sticky` with an `auto` inset does not stick.

## Change

- `apps/desktop/src/features/targets/targets-table.css.ts:87`: the two
  left-offset rules cover the body cells as well as the headers.
- `tests/e2e/targets_table_pinned_columns.spec.ts`: 820px viewport (below the
  table's 1000px `minWidth`), scroll the container to 300px, then assert column 3
  moved by the scroll amount while columns 1 and 2 held their x and matched their
  own header x.

## Red/green evidence

| Tree state | `pnpm test:e2e targets_table_pinned_columns` |
| --- | --- |
| CSS fix reverted, spec present | 1 failed, star cell x moved 300, expected <= 1 |
| CSS fix applied | 1 passed (11.7s) |

`pnpm lint` exit 0. `pnpm typecheck` exit 0.

## Spec Context

Both files existed only in the reclaimed `css/j3by` worktree and were stranded
when #1668 merged as eca1f1e19.

Tracks-Bead: astro-plan-r0tgd
Closes-Bead: astro-plan-r0tgd
Merge-Bead: astro-plan-g7nmp

### DIFF
--- a/apps/desktop/src/features/targets/targets-table.css.ts
+++ b/apps/desktop/src/features/targets/targets-table.css.ts
@@ -87,12 +87,24 @@ globalStyle(`${scroll} ${table} thead th:nth-child(-n + 2)`, {
   position: 'sticky',
 });
 
-globalStyle(`${scroll} ${table} thead th:nth-child(1)`, { left: 0 });
+// `left` goes on the body cells as well as the headers. `position: sticky` with
+// an `auto` inset does not stick at all, so heading-only offsets left the star
+// and designation cells scrolling out from under their own pinned headers
+// whenever the table was narrower than its 1000px minimum.
+globalStyle(
+  `${scroll} ${table} thead th:nth-child(1), ${scroll} ${table} ${row} > td:nth-child(1)`,
+  { left: 0 },
+);
 
-globalStyle(`${scroll} ${table} thead th:nth-child(2)`, {
-  left: 'var(--pv-targets-star-w)',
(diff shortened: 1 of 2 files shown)


srobroek (commented):
Independent review (agent reviewer, wisp astro-plan-o0dme, opus, its own prepared checkout at 08d365a0e).

VERDICT: APPROVE.

The load-bearing claim was reproduced by the reviewer rather than taken from the author: the pinned-column regression is red without the CSS hunk, green with it, and a PARTIAL revert also reproduces red — so the e2e would catch a partial regression, not only a total one. apps/desktop/src/features/targets/targets-table.css.ts:94-107 matches the rules deleted at eca1f1e19^:apps/desktop/src/styles/components/targets.css:101-111 exactly. typecheck and biome clean.

Recorded by the orchestrating session on the reviewer's behalf: GitHub refuses `--approve` on a PR from the same account, so this is a comment review rather than a formal approval.

PR #1703, opened 2026-08-22
fix: log and audit exports refuse to overwrite an existing file

srobroek:
## Summary

- A log or audit export now refuses to overwrite an existing file, so an export
  aimed at a file you already have returns an error instead of replacing it.
- An export destination must be an absolute path inside an existing directory.
  A symlink at the destination, a symlink in the parent path, and a Windows
  device name such as `CON.json` are all refused.
- Both export commands share one destination validator, so the two sinks can no
  longer drift apart.

## Rejection rules

| Path sent to the export command | Result |
| --- | --- |
| a relative path with `..` segments | refused: not absolute |
| a parent that does not exist | refused: parent missing |
| a parent that exists but is a file | refused: parent not a directory |
| `CON.json`, `NUL.txt` and other device names | refused: reserved name |
| a file that already exists | refused: contents intact |
| a symlink at the destination | refused: link target untouched |
| a symlink as the parent | resolved to its real directory before the write |
| a symlink planted at the temporary path after validation | refused: `create_new` does not follow it |
| a destination created after validation but before the rename | refused when the re-check sees it, see Guarantee |

## Guarantee

Every rule above is checked at one instant. Two of them are decided again
immediately before the `rename`, and the rest hold for the path the validator
resolved.

The destination check and the `rename` are separate syscalls, so a file created
between them is replaced rather than refused. The window is the gap between the
re-check at `crates/fs/pathsafe/src/export_dest.rs:155` and the `rename` at
`:159`, against the whole serialisation and write in the pre-fix code. Closing it
needs an atomic no-replace rename: `RENAME_NOREPLACE` exists on Linux and
`MOVEFILE_REPLACE_EXISTING`'s absence gives it on Windows, with no macOS
equivalent that Rust's standard library exposes.

The temporary file carries no such window. `create_new` is atomic, so a symlink
or file at the temporary path fails the open instead of being followed.

## Overwrite rule

A filesystem mutation must never overwrite silently, and an archive-then-replace
flow needs a reviewable plan record that neither export sink has. An existing
destination is therefore refused and the export returns an error the user sees.

## Remaining exposure

Containment against a library root is absent. The destination comes from a
native save dialog (`apps/desktop/src/app/LogPanel.tsx:147`) and the repository
doesn't declare an export root, so a library-root allowlist would remove the
ability to save a log outside a library. A caller can still create a new file at any
absolute path whose parent exists. That residual exposure and the three ways to
close it are recorded in `astro-plan-7ggju`.

`astro-plan-3v3r.8.36` and the containment half of `astro-plan-3v3r.5.11` assert
containment against a declared root and stay open.

Finding `astro-plan-3v3r.8.20` also claims `log_export` returns
`Result<_, String>` and loses the error code. It returns
`Result<_, ContractError>` at `apps/desktop/src-tauri/src/commands/log.rs:101`;
only the `# Errors` doc comment said otherwise, and this branch corrects it.

## Tests

`crates/fs/pathsafe/src/export_dest.rs` holds one test per rejection rule above,
including a test that pins the replaced `create` plus `rename` sequence as
clobbering, so reverting either sink to it fails. `log_stream` and the
`audit_export` command tests each cover the existing-destination and
relative-path refusals at the sink.

`create_new` refuses a name held by a symlink, and a bare `create` plus `rename`
pair replaces an existing file. Each is asserted as an invariant, because a
scenario version depends on the temporary-name counter that other tests in the
same process advance. Every test in the file passes under `cargo nextest run` (one
process per test) and under `cargo test` (one process per binary), which is what
`cargo llvm-cov` wraps.

## Verification

| Command | Exit |
| --- | --- |
| `cargo clippy --workspace --all-targets -- -D warnings` | 0 |
| `cargo nextest run --workspace` | 0 (3126 passed, 31 skipped) |
| `cargo fmt --check` | 0 |
| `cd apps/desktop && pnpm lint` | 0 |

## Spec Context

Tracks-Bead: astro-plan-wo2ju
Tracks-Bead: astro-plan-3v3r.8.19
Tracks-Bead: astro-plan-3v3r.8.20
Tracks-Bead: astro-plan-3v3r.5.11
Merge-Bead: astro-plan-ka563

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -110,6 +110,7 @@ dependencies = [
  "filetime",
  "fs_executor",
  "fs_inventory",
+ "fs_pathsafe",
  "hex",
  "patterns",
  "persistence_calibration",
@@ -1421,6 +1422,7 @@ dependencies = [
  "contracts_core",
  "domain_core",
  "fs_inventory",
+ "fs_pathsafe",
  "futures-util",
  "persistence_core",
  "persistence_inbox",
@@ -2030,6 +2032,7 @@ version = "0.1.0"
 dependencies = [
  "camino",
  "proptest",
+ "safe-filename",
  "tempfile",
 ]
 
--- a/apps/desktop/src-tauri/Cargo.toml
+++ b/apps/desktop/src-tauri/Cargo.toml
@@ -33,6 +33,7 @@ skymath = "0.7.2"
 contracts_core = { path = "../../../crates/contracts/core" }
 domain_core = { path = "../../../crates/domain/core" }
 fs_inventory = { path = "../../../crates/fs/inventory" }
+fs_pathsafe = { path = "../../../crates/fs/pathsafe" }
 persistence_core = { path = "../../../crates/persistence/core" }
 persistence_inbox = { path = "../../../crates/persistence/inbox" }
 persistence_lifecycle = { path = "../../../crates/persistence/lifecycle" }
--- a/apps/desktop/src-tauri/src/commands/audit.rs
+++ b/apps/desktop/src-tauri/src/commands/audit.rs
@@ -288,8 +288,13 @@ pub async fn audit_list(
 /// a file (mirrors `log.export`). Streams backend-side; only the path and count
 /// cross IPC.
 ///
+/// The destination passes `fs_pathsafe::export_dest`, so an existing file is
+/// refused rather than replaced and the write cannot leave the canonicalized
+/// parent directory.
+///
 /// # Errors
-/// Returns `Err(ContractError)` on database or filesystem failure.
+/// Returns `Err(ContractError)` on a refused destination, database failure, or
+/// filesystem failure.
 #[tauri::command]
 #[specta::specta]
 pub async fn audit_export(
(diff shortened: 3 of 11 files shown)


srobroek (commented):
Round 3 review: approve.

Verified on head a918d2c46b4c60b4055b78e6eedae9b7d768237d.

- llvm-cov diagnosis holds. `.github/workflows/coverage.yml:14` states cargo-llvm-cov wraps `cargo test`, which runs a binary's tests as threads in one process, so the deleted test's read of the shared `TEMP_SEQ` static (`crates/fs/pathsafe/src/export_dest.rs:121`, `fetch_add` at :167) could not predict the names `create_temp` would try. Not a scheduler race. All lanes are green, including Rust coverage (llvm-cov).
- The TOCTOU claim is accurate. The re-check at `crates/fs/pathsafe/src/export_dest.rs:155` and the rename at `:159` are the two syscalls the PR body Guarantee section names, and the line references match the code.
- Verbatim-path traversal gap closed at `crates/fs/pathsafe/src/export_dest.rs:76`. `safe_filename::step3_traversal_check` rejects only `.` and `..` (`crates/safe-filename/src/lib.rs:123-128`), so a legitimate name such as `my..file.json` is not falsely refused.
- Multi-dot device names covered at `crates/fs/pathsafe/src/export_dest.rs:266` (`CON.foo.json`, `nul.a.b`).

Non-blocking nit: `create_new_refuses_a_symlink_instead_of_writing_through_it` (`crates/fs/pathsafe/src/export_dest.rs:391`) asserts std `create_new` semantics rather than `create_temp` use of them, so dropping `create_new` from `create_temp` would not fail this test. Exercising production code deterministically needs a temp-name seam; a follow-up is the right place for it.

PR #1705, opened 2026-08-22
fix: refuse project paths and plan destinations outside the chosen folder

srobroek:
## Summary

- A project path of `/` was accepted at project creation, became the source-view
  destination root, and scaffolding plus generated links resolved directly under
  the filesystem root. Project creation now refuses any path that names no
  folder, which covers every filesystem and device root.
- A frame filter or exposure value containing `../` produced a plan-item
  destination outside the view root. Pattern resolution now applies the same
  traversal check on both of its entry points.
- Plan apply now validates the destination as well as the source before it
  mutates anything, so a destination outside the root the plan declared is
  refused with `root_escape` instead of applied.

## Detail

Three guards, each sufficient on its own.

| Layer | Location | Refusal |
| --- | --- | --- |
| Project path | `crates/app/projects/src/project_setup/create.rs:47` | A path with no `Normal` component fails with `path.invalid`. |
| Pattern resolution | `crates/patterns/src/resolver.rs` | `check_assembled_path` is the one implementation both `resolve` and `resolve_pattern_str` run; a `..` segment in the assembled path fails with `path.traversal`. |
| Plan apply | `crates/fs/executor/src/run/loop_.rs` | `gate_item_paths` validates the destination against the root `dispatch::execute_item` joins it onto. |

`safe-filename` leaves `/` inside a token value on purpose and defers separators
to the resolver (`crates/safe-filename/src/lib.rs:102`). Only `resolve` acted on
that; `resolve_pattern_str` did not, and it is the entry point the source-view
generator and the Inbox confirm pipeline use.

`plans.destination_root` (migration 0003) records the destination root the
source-view generator already computes. Source-view items store an absolute
destination with no `to_root_id`, so the executor used to fall back to the item's
source root; the recorded root outranks that fallback. A NULL column leaves the
destination gate inactive, so every other plan type applies as before.

`path_gate::resolve_and_validate` walks its symlink check below the root rather
than from the path's own root component, so an absolute destination no longer
trips on ancestor symlinks above the root.

A destination outside the declared root is not a plan the user could have
reviewed, so refusing it at apply time is the constitution II promise rather than
a new policy.

## Tests

| Test | Pins |
| --- | --- |
| `create_filesystem_root_path_rejected` | astro-plan-3v3r.9.12 |
| `pattern_str_token_value_traversal_rejected`, `assembled_traversal_rejected_alike_by_both_resolvers` | astro-plan-3v3r.9.10 |
| `destination_outside_destination_root_is_refused`, `destination_under_destination_root_is_allowed` | astro-plan-3v3r.9.29 |
| `resolve_and_validate_absolute_path_under_root_ok`, `resolve_and_validate_absolute_path_outside_root_refused` | the gate's absolute-path handling |

`cargo clippy --workspace --all-targets -- -D warnings`, `cargo nextest run
--workspace` (3112 passed, 31 skipped), and `cargo fmt --check` each exit 0. No
file under `apps/desktop` changed.

Tracks-Bead: astro-plan-d8cyr
Tracks-Bead: astro-plan-3v3r.9.29
Tracks-Bead: astro-plan-3v3r.9.12
Tracks-Bead: astro-plan-3v3r.9.10
Merge-Bead: astro-plan-wgrd4

### DIFF
--- a/crates/app/core/src/plan_apply/apply.rs
+++ b/crates/app/core/src/plan_apply/apply.rs
@@ -99,9 +99,18 @@ pub async fn apply_plan(
         }
     }
 
+    let plan_destination_root: Option<Utf8PathBuf> =
+        plan_row.destination_root.as_deref().map(Utf8PathBuf::from);
     let executor_items: Vec<ExecutorItem> = item_rows
         .iter()
-        .map(|r| item_row_to_executor_item(r, &root_map, &plan_row.destructive_destination))
+        .map(|r| {
+            item_row_to_executor_item(
+                r,
+                &root_map,
+                &plan_row.destructive_destination,
+                plan_destination_root.as_deref(),
+            )
+        })
         .collect();
 
     // Overlap check + active-run registration (FR-017, R-Concur-1): the
--- a/crates/app/core/src/plan_apply/lifecycle.rs
+++ b/crates/app/core/src/plan_apply/lifecycle.rs
@@ -422,10 +422,19 @@ pub async fn resume_plan(
     // audit record (review fix, resume+retry item-set mismatch).
     // succeeded/skipped/cancelled items are still excluded: they are truly
     // terminal and never eligible for retry.
+    let plan_destination_root: Option<Utf8PathBuf> =
+        plan_row.destination_root.as_deref().map(Utf8PathBuf::from);
     let executor_items: Vec<ExecutorItem> = item_rows
         .iter()
         .filter(|r| matches!(r.item_state.as_str(), "pending" | "failed"))
-        .map(|r| item_row_to_executor_item(r, &root_map, &plan_row.destructive_destination))
+        .map(|r| {
+            item_row_to_executor_item(
+                r,
+                &root_map,
+                &plan_row.destructive_destination,
+                plan_destination_root.as_deref(),
+            )
+        })
         .collect();
 
     // Re-register the ActiveRun (R-Concur-1). A paused run has no registry
--- a/crates/app/core/src/plan_apply/paths.rs
+++ b/crates/app/core/src/plan_apply/paths.rs
@@ -178,7 +178,8 @@ pub(super) fn materialization_from_provenance(
 /// cross-root move joins `to_relative_path` against the *picked* destination
 /// root instead of silently reusing the source root (#765). Precedence:
 /// `to_root_id` → `plan_destination_root` (`plans.destination_root`) →
-/// `library_root`.
+/// `library_root`, and the last step is skipped for an absolute destination,
+/// which belongs to no root the item names.
 ///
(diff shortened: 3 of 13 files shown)


srobroek:
Requesting changes: L1, L2, and the executor gate itself are sound, but the destination-root precedence regresses source-view plans that were generated before migration 0003.

**FIX 1 -- `crates/app/core/src/plan_apply/paths.rs:258-264`**

`destination_root` falls back to `library_root` (the SOURCE root) when `plans.destination_root` is NULL. Source-view generation stores an absolute `to_relative_path` with `to_root_id = None` and `from_root_id = Some(source_root_id)` (`crates/app/projects/src/source_view_generate/generate.rs:611-620`), so for a `source_view_generation` plan row that predates migration 0003 the new destination gate (`crates/fs/executor/src/run/loop_.rs:92-96`) resolves an absolute destination against the source root, `starts_with` fails, and every item is refused with `root_escape` (`crates/fs/executor/src/ops/path_gate.rs:65-73`). `FailureCode::RootEscape` is not retryable (`crates/fs/executor/src/failure.rs:154`), so an already-reviewed pending plan becomes unappliable after upgrade.

The report on astro-plan-d8cyr states "NULL leaves the destination gate inactive, so no other plan type changes behaviour". That holds for regeneration and removal plans, which carry `from_root_id = None` (`crates/app/projects/src/prepared_views.rs:390-393`), but not for the generation plans this fix targets.

Required action: make the NULL case genuinely inactive or genuinely correct -- either do not fall back to `library_root` when the item's destination path is absolute, or backfill `plans.destination_root` for existing `source_view_generation` rows in migration 0003. Add a test at the `item_row_to_executor_item` level pinning `plan_destination_root = None` + `from_root_id = Some(..)` + an absolute `to_relative_path`; no current test covers that combination (`crates/app/core/src/plan_apply/tests.rs:1421-1507` covers only relative destinations).

While there: with the fallback in `paths.rs`, the `.or(item.library_root.as_ref())` in `gate_item_paths` (`crates/fs/executor/src/run/loop_.rs:93`) can never fire.

**Verified sound**

- L1 `crates/app/projects/src/project_setup/create.rs:68-77`: no `Component::Normal` covers `/`, a device root, `.`, and the empty string.
- L2 `crates/patterns/src/resolver.rs:284-310`: `check_assembled_path` reproduces the old `resolve` policy exactly (`..`/`.` rejection, segment byte cap, total char cap) and adds it to the pattern-string path; the `Ha/../../etc` token value now fails on the joined path.
- L3 symlink-walk fix `crates/fs/executor/src/ops/path_gate.rs:84`: `strip_prefix(root)` keeps the walk below the root for an absolute input; the `unwrap_or` default is unreachable after the prefix check.
- `set_destination_root` (`crates/persistence/plans/src/repositories/plans.rs:368-377`) checks `rows_affected`, matching #1702.

PR #1713, opened 2026-08-22
test(persistence): pin the sqlx migration-failure behaviour the backup refutation rests on

srobroek:
## Summary

- Adds `a_failed_migration_records_no_row`, which pins what sqlx writes to `_sqlx_migrations` when a migration script fails.
- Adds `every_migration_runs_inside_a_transaction`, which asserts that no migration in the compiled-in set opts out of its per-migration transaction.
- Adds `backup_to_rejects_an_existing_destination_file`, covering the destination case the suite left open.
- Adjudicates a P0 chain finding: the pre-migration backup writes nothing, and a resulting partially-applied migration cannot be diagnosed. The chain is refuted and the guards make the refutation executable. One narrow diagnosis gap survives on its own bead.

## `MigrateError::Dirty` reachability

`MigrateError::Dirty` reports a `_sqlx_migrations` row with `success = false` (`sqlx-core-0.9.0/src/migrate/migrator.rs:248-251`, reading `sqlx-sqlite-0.9.0/src/migrate.rs:109-123`). This build never writes such a row: `execute_migration` inserts `success` as a `TRUE` literal and inserts nothing at all when the script fails (`sqlx-sqlite-0.9.0/src/migrate.rs:249-272`).

That is why `migration_divergence_detail` omits a `Dirty` arm. The invariant belongs to sqlx rather than to this crate, so an upgrade that starts recording failed migrations can break it while every other test passes. `a_failed_migration_records_no_row` builds a `Migrator` whose second migration collides with the first and asserts that sqlx reports the failure as `ExecuteMigration`, that it writes no `success = false` row, that a rerun declines to escalate to `Dirty`, and that `migration_divergence_detail` returns `None`.

## The transaction guard

A migration file starting with `-- no-transaction` sets `Migration::no_tx` (`sqlx-core-0.9.0/src/migrate/source.rs:234`) and sqlx then runs that script outside a transaction. A failure partway leaves the statements it already ran in place, and since a failed migration doesn't record a bookkeeping row, nothing marks the database as partially applied. Recovery depends entirely on the pre-migration backup.

The test reads the public `no_tx` flag through `Database::migrator()`, the same static that `migrate_uncached` runs, so it binds to the production migrator. Verified failing: prepending the marker to `0003_plans_destination_root.sql` turns it red and names version 3.

## VACUUM INTO ruling

Measured against a file-backed source database. `VACUUM INTO` fails rather than silently no-opping for every destination case:

| Destination | Result |
|---|---|
| Directory absent | code 14, `unable to open database` |
| Existing non-database file | code 26, `file is not a database`; the file keeps its bytes |
| Existing database | code 1, `output file already exists`; the earlier backup survives |
| In-memory source | statement reports success without writing the file; the post-condition rejects it |

Relying on the statement's error is sufficient for a file-backed source, and the post-condition at `crates/persistence/core/src/lib.rs:212-217` covers the one case where the statement misreports. `backup_to` begins at `:204`; `:243` is `migrator()`.

The claimed endpoint, whole-database loss with no recovery path, does not hold: a failed migration script rolls back and the schema stays unchanged.

## Filed separately

- astro-plan-00yw0 (P3): `migration_divergence_detail` returns `None` for a `Dirty` database written by another sqlx tool. The compiled-in migration set cannot constrain a foreign writer, so no test here enforces it away. It costs no data, because sqlx reads the dirty version before applying anything.
- astro-plan-d6glj (P3): `run_pre_migration_backup` demotes a backup failure to a warning and boot migrates anyway (`apps/desktop/src-tauri/src/lib.rs:523-530`, `:590-594`).

## Verification

- `cargo nextest run -p persistence_core`: exit 0, 56 of 56 passed
- `cargo clippy -p persistence_core --all-targets`: clean, exit 0
- `cargo fmt --all`: exit 0

Tracks-Bead: astro-plan-f8g21
Merge-Bead: astro-plan-b4ci9

### DIFF
--- a/crates/persistence/core/src/lib.rs
+++ b/crates/persistence/core/src/lib.rs
@@ -300,7 +300,8 @@ mod tests {
     /// reachable here. A migration that opts out of the transaction would make
     /// it reachable, and would need its own user-facing repair message rather
     /// than the "written by a different revision" wording this classifier
-    /// produces.
+    /// produces. [`every_migration_runs_inside_a_transaction`] enforces that
+    /// precondition.
     #[test]
     fn divergence_detail_ignores_unrelated_failures() {
         assert!(super::migration_divergence_detail(&super::DbError::NotImplemented).is_none());
@@ -308,6 +309,34 @@ mod tests {
             sqlx::migrate::MigrateError::Dirty(71)
         ))
         .is_none());
+    }
+
+    /// Enforces the precondition that makes a partially-applied migration
+    /// impossible on SQLite, and with it the `Dirty` arm's absence from
+    /// [`super::migration_divergence_detail`].
+    ///
+    /// A migration file marked `-- no-transaction` sets `Migration::no_tx`, and
+    /// sqlx then runs that script outside a transaction.
+    #[test]
+    fn every_migration_runs_inside_a_transaction() {
+        let opted_out: Vec<i64> = super::Database::migrator()
+            .iter()
+            .filter(|migration| migration.no_tx)
+            .map(|migration| migration.version)
+            .collect();
+
+        assert!(
+            opted_out.is_empty(),
+            "migrations {opted_out:?} opt out of the per-migration transaction. \
+             Such a migration can fail partway and leave the schema partially \
+             applied, which makes MigrateError::Dirty reachable and leaves a \
+             failed migration with no rollback material. The constitution's \
+             unclean-shutdown constraint then requires a verified pre-migration \
+             backup and an explicit user resume-or-repair prompt, plus a \
+             migration_divergence_detail arm wording Dirty as a partial apply \
+             rather than a build mismatch. Satisfy those before deleting this \
+             assertion."
+        );
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


srobroek:
REVIEW round 3, dimension: whether the transaction guard can fail, and whether the refutation is complete. Verdict: changes requested. GitHub refuses a formal request-changes review on a self-authored PR, so this comment carries it.

The guard is executable and binds to the production migration set: Database::migrator() (crates/persistence/core/src/lib.rs:243-245) returns the same static schema_cache::MIGRATOR that migrate_uncached runs (:237), Migration::no_tx is a pub field (sqlx-core-0.9.0/src/migrate/migration.rs:16), and sqlx-sqlite reads it per migration (sqlx-sqlite-0.9.0/src/migrate.rs:165). CI is green on 15 checks. That part stands.

The stated mechanism does not. MigrateError::Dirty is produced only from a persisted _sqlx_migrations row with success = false (sqlx-core-0.9.0/src/migrate/migrator.rs:248-251, reading sqlx-sqlite-0.9.0/src/migrate.rs:109-123). sqlx-sqlite 0.9 never writes such a row: execute_migration inserts success TRUE after the script succeeds (sqlx-sqlite-0.9.0/src/migrate.rs:249-269), and it is called both for the no_tx branch (:166) and inside the transaction (:174). A no_tx migration that fails partway therefore inserts no row at all. It leaves a partially-applied schema that Dirty cannot name, which is worse than the documented case, and it does not make Dirty reachable.

1. crates/persistence/core/src/lib.rs:330-338 - the failure message states a no_tx migration "makes MigrateError::Dirty reachable", which is false for this sqlx build - reword to the real consequence, a partially-applied schema with no bookkeeping row and therefore no Dirty diagnosis at all, keeping the rollback-material and resume-or-repair obligations. Apply the same correction to the docstring at :296-304.

2. crates/persistence/core/src/lib.rs:303-304 and :314-316 - the claim that every_migration_runs_inside_a_transaction "enforces that precondition" for the Dirty arm is not what the test checks - Dirty stays reachable independently of this build from any pre-existing success = false row, for example a database migrated by an older sqlx or by sqlx-cli, and the guard cannot see that. State the actual ground for the Dirty to None arm, that this build never records a dirty row, and name the foreign-writer route as the residual on its own bead rather than claiming enforcement.

No change requested to backup_to_rejects_an_existing_destination_file. The VACUUM INTO ruling and the corrected .12.21 citation both check out at HEAD: backup_to is at :204 and its post-condition at :212-217.

PR #1715, opened 2026-08-23
docs: align the Sessions requirements with the shipped automatic session model

srobroek:
## Summary

- The Sessions requirements asked for a toolbar with bulk Confirm, Split, Merge,
  and Use-in-project actions over multi-selected rows. The product has none of
  them, and the requirement blocked the all-specs-implemented release gate.
- That requirement is marked superseded and now states the shipped session
  model. No product behaviour changed in this PR.

## What the product does

- `crates/app/targets/src/ingest_sessions.rs:1-30` folds each applied light
  frame into an `acquisition_session` keyed by target, filter, binning, gain,
  and observing night. Session membership is automatic.
- `apps/desktop/src/features/sessions/SessionsPage.tsx:14-30` documents a
  toolbar of search plus a group-by control. The Sessions feature has no row
  checkbox and no range select.
- `apps/desktop/src-tauri/src/commands/sessions.rs:128-170` holds
  `sessions.split` and `sessions.merge` as fixture stubs returning hardcoded
  counts. No frontend module invokes either.

## Spec Context

- `specs/027-frontend-implementation/spec.md` FR-015 keeps its id and its
  original text, marked SUPERSEDED by spec 041 FR-051 and spec 062, following
  the retirement style spec 041 uses for US5. The note records four properties
  of the session model: automatic capture-identity keying (spec 035 FR-016), one
  homogeneous capture identity per session so there is no split point, derived
  already-confirmed inventory so no Confirm applies (spec 041 FR-051), and
  project attachment through the project-creation wizard.
- The note also places panel and mosaic split, merge, and identity change on
  logical group identity rather than session membership (spec 062 FR-044). Spec
  062 US2 specifies the surface for those changes; its Tauri commands have no
  backend handlers, tracked by bead astro-plan-ic9h.20.
- The spec 027 command contract marks both commands as superseded with no caller
  and records their spec 029 T021 stub origin.
- Task T038 of spec 027 still describes the bulk toolbar. Its file is read-only
  legacy under the beads workspace guard, and no task bead asserts the behaviour.
- Follow-ups: `astro-plan-i09vh` removes the dead stubs, bindings, mocks, and
  stub tests. `astro-plan-6lt63` covers FR-017 and FR-020 to FR-024, which still
  describe the removed session review lifecycle.
- Merge bead `astro-plan-2qzxj`. Work bead `astro-plan-ci707`.

### DIFF
--- a/specs/027-frontend-implementation/contracts/tauri-commands.md
+++ b/specs/027-frontend-implementation/contracts/tauri-commands.md
@@ -34,8 +34,8 @@ This document defines the command interface that the frontend depends on. The ac
 | Command | Args | Returns | Used By |
 |---------|------|---------|---------|
 | `sessions.transition` | `{ id, action, metadata? }` | `Result<Session>` | Review queue, detail |
-| `sessions.split` | `{ id, split_at_index }` | `Result<{ original: Session, new: Session }>` | Sessions toolbar bulk action |
-| `sessions.merge` | `{ ids }` | `Result<Session>` | Sessions toolbar bulk action |
+| `sessions.split` | `{ id, split_at_index }` | `Result<{ original: Session, new: Session }>` | none -- SUPERSEDED with FR-015 |
+| `sessions.merge` | `{ ids }` | `Result<Session>` | none -- SUPERSEDED with FR-015 |
 | `projects.create_plan` | `{ wizard_state }` | `FilesystemPlan` | Wizard step 6 |
 | `plans.approve` | `{ id, delete_acknowledged? }` | `Result<Plan>` | Plan review |
 | `plans.apply` | `{ id }` | `OperationHandle` | Plan review |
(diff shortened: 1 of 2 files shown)


srobroek (commented):
Changes requested. Round 1 (the prior review wisp was blocked by an actor-guard denial before any review ran, so this is the first substantive pass).

Verified and accepted: FR-015 retirement is the right call. `sessions_split` / `sessions_merge` are fixture stubs (apps/desktop/src-tauri/src/commands/sessions.rs:134,151) with zero frontend callers, and SessionsPage.tsx carries single-row selection only (`selected` is one session UUID, SessionsPage.tsx:107), so the removed bulk toolbar genuinely does not exist. Keeping the FR-015 id with a RETIRED note is correct: contracts/tauri-commands.md:37-38 references it. Cross-references check out against their sources (spec 035 FR-016:278, spec 041 FR-051:474, spec 062 FR-044:403). The T038 residue in the legacy task file is unfixable by policy, not oversight: the speckit guard denies every Write/Edit to that path. Gates re-run here: `just check-contracts-ts` exit 0; review-docs slop-lint exit 1 with all four findings on pre-existing lines (10/26/27/29), none introduced by this diff.

## FIX

1. `specs/027-frontend-implementation/spec.md:252-254` - the replacement text asserts a surface that does not exist at HEAD: "Panel and mosaic relations are proposed and accepted individually in the Sessions detail surface (spec 062 FR-044)". `SessionsPage.tsx:336-344` records that the `RelationProposalList`, `ManualRelationDialog`, and `MatchingSettingsPanel` render sites were removed because their IPC commands have no Tauri command layer, and `SessionDetail.tsx` contains no proposal UI at all. Spec 062's own header marks US2 as not built. This is the same defect class FR-015 was amended to remove, a requirement text asserting behaviour the product does not offer, so the amendment moves the inconsistency instead of removing it. Required action: state the relation model in the same non-present terms used elsewhere in the note, or drop the sentence. If it stays, mark the surface as specified-but-unbuilt and cite the tracking bead (astro-plan-ic9h.20 is the one named in the SessionsPage comment), so no reader takes it as shipped behaviour.

PR #1720, opened 2026-08-23
fix: repository check scripts no longer report success when a scan finds no files

srobroek:
Closes the fail-open in five repository check scripts: each asserted on the
content of an enumeration and never on its size, so a scan returning zero items
satisfied every content assertion and the gate printed its OK line.

Work bead: `astro-plan-vtkpl`. Merge bead: `astro-plan-ibp9p`.
Findings addressed: `astro-plan-3v3r.16.9`, `.16.15`, `.16.16`, `.16.17`.

## Reachability, per bead

`astro-plan-3v3r.16.9` (`check-perf-baseline.sh`) is reachable by a routine
action. All five `scripts/perf-baseline.json` scenarios match the perf-bench
emitters today, so the gate does compare on `main`. Adding a sixth scenario
produced a permanent `WARN: … not found in baseline — skipping` and an
indefinitely unenforced budget.

`astro-plan-3v3r.16.15` and `.16.16` (`check-hot-read-ratchet.sh`, instance and
representative) are reachable by refactor. Both scoped directories and the
watcher file exist, so the ratchet measures 6 sites on `main`. A crate or module
move empties the scope; the old `find … 2>/dev/null || true` then guaranteed exit
0, and a current total of 0 can never exceed a shrink-only baseline. This repo
moves crates routinely — the `app/*` crates were split out of `app/core`.

`astro-plan-3v3r.16.17` (`check-db-boundary.sh`) is reachable in principle only.
Collapsing the enumeration requires top-level `crates/` and `apps/` to be absent,
which cannot happen in a CI job running at the repo root. This PR does not claim
a live false pass for it; the change removes the fail-open, it does not correct a
wrong result on `main`.

## Per item

| Bead | Site | New test | Fails at base | How confirmed |
|---|---|---|---|---|
| `.16.9` | `check-perf-baseline.sh:169`, `:204` | `self_test` empty-stream case (`:217`) | yes | reverted the `COMPARED -eq 0` floor in a copy → `--self-test` exit 1, `FAIL: … compared 0 scenarios and passed` |
| `.16.9` | `check-perf-baseline.sh:169` | `self_test` unlisted-scenario case | yes | reverted the missing-entry `fail=1` back to the WARN-skip → `--self-test` exit 1, `FAIL: a scenario absent from the baseline passed with no enforced budget`. The stream carries a baselined scenario as well, so this case cannot be satisfied by the zero-comparison floor |
| `.16.15`, `.16.16` | `check-hot-read-ratchet.sh:168` (OK line `:338`) | `self_test` empty-scope and missing-directory cases (`:205`) | yes | reverted both the per-scope floor and the existence check → `--self-test` exit 1 with both FAIL lines |
| `.16.17` | `check-db-boundary.sh:221` (OK line `:397`) | two cases appended to the existing `self_test` (`:257`) | yes | same two reverts in a copy → `--self-test` exit 1 with both FAIL lines |
| extra | `check-dead-callers.sh:174`, `:252` (OK line `:823`) | three cases in `self_test` (`:618`) | yes | three separate mutants, one per guard; each `--self-test` printed only its own FAIL line and exited 1 |
| extra | `check-eslint-baseline.mjs:109` (OK line `:158`) | two assertions in `scripts/check-eslint-baseline.test.mjs` | yes | reverted the floor in place → `node scripts/check-eslint-baseline.test.mjs` exit 1, `a zero-file lint run must produce a floor error` |

## Two extra sites

`check-dead-callers.sh` guarded the optional `src-tauri` root with `[ -d ]` but
not `crates/`, and an empty definition set yields an empty dead set — the same
output as a fully-called tree. Same severity as the hot-read ratchet: reachable
by a crate move, not wrong on `main` today.

`check-eslint-baseline.mjs` printed `eslint: OK (N files linted)` for any `N`,
including 0. Lower severity: a lint target that stops resolving is usually
visible in the same job's other steps, but the line reported a pass either way.

## pipefail

`set -o pipefail` was already enabled in all four shell scripts before this
change, and enabling it fixes nothing here. `list_files` runs inside command
substitution at every call site (`$(…)`, `< <(…)`), where its exit status is
discarded regardless of `pipefail` — `check-db-boundary.sh` demonstrated exactly
that. The load-bearing change is the positive-count assertion in the caller.

## Deliverable shape

No bats test is added. Nothing at `origin/main` runs `scripts/tests`: no runner
appears in `.github`, the `justfile`, or `package.json`, so
`scripts/tests/pv-watch-queue.bats` and `check-coderabbit-review.bats` execute
nowhere. Each negative case is instead a case in the script's own `self_test`,
which every `--check` invocation runs, matching the existing precedent in
`check-db-boundary.sh` and `check-dead-callers.sh`. No file under `.github/` is
touched. Follow-up for a real bats runner: `astro-plan-858oc`.

## Residual gaps

A census of every OK-printing script under `scripts/` accompanies this change.
Three scripts remain unguarded and are filed as `astro-plan-0x7y4` rather than
bundled here: `check-merged-pr-paths.sh:430` (highest severity — an empty
merged-PR list, a changed `base.ref` convention, or a window bug all read as a
green sweep), `check-generated-drift.sh:35`, and
`check-github-not-ignored.sh:29`. The remaining scripts either assert a positive
count already or have no enumeration whose emptiness could fake a pass.

## Verification

- `bash scripts/check-hot-read-ratchet.sh --self-test` and `--check`: pass, 6 sites across 35 scanned files.
- `bash scripts/check-db-boundary.sh --self-test` and `--check`: pass, 0 sites across 427 scanned files.
- `bash scripts/check-dead-callers.sh --self-test` and default check: pass, 148 baselined across 528 scanned files.
- `bash scripts/check-perf-baseline.sh --self-test`: pass. The `--check` path was exercised by feeding `compare_stream` a bench stream synthesized from the checked-in `scripts/perf-baseline.json`; all five scenarios compared, `COMPARED=5`. `cargo run --release -p perf-bench` was not run in this branch.
- `node scripts/check-eslint-baseline.test.mjs`: OK, 6 assertions. `node scripts/check-eslint-baseline.mjs`: OK, 732 files linted.
- `shellcheck` clean on all four shell scripts.

Tracks-Bead: astro-plan-vtkpl
Merge-Bead: astro-plan-ibp9p


srobroek (commented):
Approve. Reviewed at head 58f9759 against b1bab46.

All five briefed sites now assert a positive count at the caller and print it on the OK line: check-perf-baseline.sh:144-209 (COMPARED floor plus a missing-baseline-entry hard fail), check-hot-read-ratchet.sh:168-199, check-db-boundary.sh:221-248, check-dead-callers.sh:174-195 (plus a pub-fn definition floor at :251), check-eslint-baseline.mjs:109-123. The floor is per scope root in each case, so one root emptying cannot be masked by the others.

Verified locally: all four --self-test targets pass; --check passes for hot-read (6 sites / 35 files), db-boundary (0 / 427 files), dead-callers (0 new / 528 files). Diff is confined to scripts/ (6 files), no .github/ edit, no new runner - the eslint test is already driven by justfile:11. All five perf scenarios in scripts/perf-baseline.json have live emitters, so the new missing-entry hard fail does not turn the lane red today.

Non-blocking notes:
1. check-hot-read-ratchet.sh:218 - shellcheck reports SC2115 on the rm -rf. The path comes from mktemp -d so it is safe, and shellcheck is not a CI gate here, but the report's 'shellcheck clean' claim is not accurate.
2. check-dead-callers.sh:183 - the floor runs its own find rather than find_rs_files, so its predicates can diverge from the walk it guards.
3. check-perf-baseline.sh:37 - documents a --stdin mode the case statement rejects. Pre-existing at the base, not introduced here.
4. check-perf-baseline.sh --check end to end is untested (cargo run was out of scope per brief); compare_stream was exercised against the real baseline only.

PR #1732, opened 2026-08-23
fix: dependency refresh fails because three packages moved upstream

srobroek:
## Summary

- `apm update` aborted before installing anything: three declared dependency subpaths no longer exist upstream, so resolution failed validation.
- `speckit` moved to its own repository and absorbed `steering-speckit`; both subpaths are gone. Now points at `srobroek/speckit-conductor`, a root package rather than a subpath.
- `mcp-repomix` was dropped upstream in favour of the repomix CLI.
- Verified: `apm update --yes` reports "Updated 51 APM dependencies" where it previously exited on validation failure.

Found while applying a global and per-repo APM refresh. This breakage predates that work and would have blocked any dependency update here.

Merge bead: astro-plan-55v6f

### DIFF
--- a/.beads/formulas/journey-validation-run.formula.toml
+++ b/.beads/formulas/journey-validation-run.formula.toml
@@ -1,0 +1,53 @@
+formula = "journey-validation-run"
+description = "Drive one user journey end-to-end against the running product (tauri-mcp), fanning out fix lanes the instant an issue is found without pausing the drive. Pour one molecule per journey and attach to the journey-validation epic."
+version = 1
+type = "workflow"
+
+[vars.journey_id]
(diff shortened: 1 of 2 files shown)


srobroek:
Closing as redundant: main already carries this fix, and in better shape than this branch had it. main pins `srobroek/speckit-conductor#>=3.0.0 <4.0.0` where this PR used a floating `#main`, and both `steering-speckit` and `mcp-repomix` are already gone from the dependency list.

Verified against `origin/main:apm.yml` rather than assumed. `apm update` succeeds here now (51 dependencies), which was the symptom this PR was opened for.

No action needed; nothing from this branch is worth carrying forward.

